PR #20, opened 2022-12-01
Make type casts in spl_network crate platform agnostic

knoellle:
## Introduced Changes

I had an issue with compiling the spl_network crate on my phone. According to the [documentation](https://doc.rust-lang.org/std/os/raw/type.c_char.html), the `c_char` type can be either a `u8` or an `i8` depending on the platform.

The code generated from the SPL standard message header file by bindgen currently looks like this:
```rs
pub struct SPLStandardMessage {
    pub header: [::std::os::raw::c_char; 4usize],
   ...
}
```
However, we were casting rust `char`s to `i8` directly instead of `c_char` which led to a type error.

## ToDo / Known Issues

- [ ] Decide whether to use `c_char` from `os::raw` or `ffi`. The generated code uses `os::raw`, but `ffi` seems more fitting.

## How to Test

See if everything still compiles for x86 locally, for the nao using the toolchain, and on android.

### DIFF
--- a/crates/spl_network/src/game_controller_return_message.rs
+++ b/crates/spl_network/src/game_controller_return_message.rs
@@ -1,4 +1,4 @@
-use std::{mem::size_of, slice::from_raw_parts};
+use std::{ffi::c_char, mem::size_of, slice::from_raw_parts};
 
 use nalgebra::Isometry2;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


schmidma (approved):
LGTM, did not test on android, but nao looks fine. Can additionally be validated with a testgame on wednesday

PR #22, opened 2022-12-03
Supertwix

knoellle:
Merged version of multiple separate PRs because some of them depend on each other and having to chain rebases is a pain.
Also, no one was going to review five different PRs where some of the code was already deprecated.

## Introduced Changes

- Dockable panels using [egui_dock](https://crates.io/crates/egui_dock)
- Ball detection overlay
- Popups don't close on the first interaction anymore
- Better connection status indication

## ToDo / Known Issues

- [ ] Program crashes when closing the last tab

- Add more keybindings for navigating and manipulating the panel tree.

## How to Test

- Docking
  - Spawn new tabs by clicking the little `+` icon or pressing `Ctrl`+`t`.
  - Tabs can be dragged around to rearrange them, convert them into panels and back.
  - Changing panel type (Text, Map, Image, etc.) should (only) change the active tab.
- Ball detection overlay
  - Open an image panel and enable it in the overlays menu.
  - Should work the same as in flora
- Popups don't close on the first interaction anymore
  - pretty self explanatory
  - less annoying now when toggling more than one map layer
  - perhaps more annoying when just toggling one as you need one extra click
- Status indication
  - The label next to the "connection intent" checkbox now changes both text and color depending on the connection state.
  - White: No connection intent
  - Yellow: Connecting
  - Green: Connected
  - Red: Failed to connect

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3209,6 +3209,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "communication",
+ "convert_case",
  "eframe",
  "egui_dock",
  "egui_extras",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -947,7 +947,7 @@ dependencies = [
 [[package]]
 name = "egui_dock"
 version = "0.2.1"
-source = "git+https://github.com/Avarel/egui_dock/?branch=tab_add#d8302974bbbcfc52a061eb91120005b9ce9ab52e"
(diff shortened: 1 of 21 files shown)


Bendodroid:
During todays testgame we noticed that when removing the connect checkmark, often multiple images arrive "late". Might be delayed disconnect, might be not-yet-painted images left in a buffer.

Not merge-blocking, just something to investigate.

schmidma (approved):
We discussed to integrate the features without further cleanup. Minor bugs and room for improvement is tolerated for this kind of tooling. Feel free to open PRs to iteratively improve the user experience.

PR #25, opened 2022-12-04
Twix Remote Control

knoellle:
## Introduced Changes

Allows twix to act as a remote control for the robot. Currently only supports walking, no kicks.
Tested with my steamdeck but should™ work with any controller.

Based on: #22 

## ToDo / Known Issues

- [x] [Install libudev in CI runners](https://github.com/HULKs/hulk/actions/runs/3615021255/jobs/6091848765)

- Add kick support
- Visualization of next step in twix
- Only update parameter when axis value changes
- Add support for controlling with keyboard (WASD, arrowkeys, etc.)

## How to Test

- Connect gamepad to computer
- Start twix and open "Remote" panel
- Connect to nao
- Press the start button on the controller or check the "enabled" checkbox
- Move left/right stick to make the robot walk/turn in that direction.

### DIFF
--- a/tools/twix/src/panels/remote.rs
+++ b/tools/twix/src/panels/remote.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use eframe::egui::Widget;
-use gilrs::{Axis, Event, GamepadId, Gilrs};
+use gilrs::{Axis, Button, Gamepad, GamepadId, Gilrs};
 use serde_json::{json, Value};
 use types::Step;
 
@@ -11,6 +11,7 @@ pub struct RemotePanel {
     nao: Arc<Nao>,
     gilrs: Gilrs,
     active_gamepad: Option<GamepadId>,
+    enabled: bool,
 }
 
 impl Panel for RemotePanel {
@@ -19,11 +20,13 @@ impl Panel for RemotePanel {
     fn new(nao: Arc<Nao>, _value: Option<&Value>) -> Self {
         let gilrs = Gilrs::new().unwrap();
         let active_gamepad = None;
+        let enabled = false;
 
         Self {
             nao,
             gilrs,
             active_gamepad,
+            enabled,
         }
     }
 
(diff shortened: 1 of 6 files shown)


tools/twix/src/panels/remote.rs
```diff
@@ -0,0 +1,89 @@
...
+    gilrs: Gilrs,
+    active_gamepad: Option<GamepadId>,
+    enabled: bool,
+}
+
+impl Panel for RemotePanel {
+    const NAME: &'static str = "Remote";
+
+    fn new(nao: Arc<Nao>, _value: Option<&Value>) -> Self {
+        let gilrs = Gilrs::new().unwrap();
```
okiwi6, line 21: maybe you can present a better error message if the `Gilrs` creation failed?

tools/twix/src/panels/remote.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.nao
+            .update_parameter_value("control.step_planner.injected_step", step);
+    }
+}
+
+impl Widget for &mut RemotePanel {
+    fn ui(self, ui: &mut eframe::egui::Ui) -> eframe::egui::Response {
+        self.gilrs.inc();
+
+        if ui.checkbox(&mut self.enabled, "Enabled (Start)").changed() {
```
okiwi6, line 53: Maybe that`s just personal, but for readability I'd prefer
```rs
ui.checkbox(&mut self.enabled, "Enabled (Start)");
if self.enabled { ... }
```

okiwi6 (changes requested):
I think this feature is quite neat. If you have some time, maybe you can rebase it to main, so that I can test it

tools/twix/src/panels/remote.rs
knoellle: That's not the same logic though.
The if block is supposed executed when the enabled state changed, not every render cycle when enabled is true.

tools/twix/src/panels/remote.rs
knoellle: Changed to expect with a message explaining what happened.

knoellle:
Rebased and tested. Had to change a parameter path because we no longer prefix them with the cycler name.

okiwi6:
Doesn't work for me:
```
INFO] Try connection to ws://10.1.24.37:1337
[INFO] Connected to ws://10.1.24.37:1337
[INFO] Gamepad /dev/input/event15 (GuliKit Controller XW) connected.
[INFO] Gamepad /dev/input/event15 (GuliKit Controller XW) connected.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failed to open /home/ole/Documents/hulk/etc/configuration/hardware_ids.json

Caused by:
    No such file or directory (os error 2)

Location:
    /home/ole/Documents/hulk/crates/repository/src/lib.rs:280:14', tools/twix/src/repository_parameters.rs:26:67
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

okiwi6 (dismissed):
Now works for me, nice feature

okiwi6 (approved):
:+1:

PR #27, opened 2022-12-05
Aliveness service and pepsi subcommand

schmidma:
## Introduced Changes

This PR adds an aliveness service for querying multiple status information from NAOs in the network. The aliveness servive is built with yocto and will be integrated in the upcoming HULKs-OS version (see the aliveness branch of HULKs/meta-hulks).

Additionally, an `aliveness` subcommand is added to pepsi for querying the aliveness information with different verbosities (`summary`, `servicese`, `battery`, `ids`, `all`). Using `-j` or `--json` the aliveness information can also be returned as JSON for further processing in e.g. status bars.

The user can furthermore set a timeout via `-t` or `--timeout` (defaulting to 200ms) and specify NAO addresses (e.g. `22` or `10.1.24.22`) for querying the aliveness information only from specific NAOs.

## ToDo / Known Issues

The beacon for querying the aliveness information is send via multicast and the answers are received via unicast.
Since the answers are from a different IP address, most firewalls will block them.

Hence, the user has to allow the incoming messages in their firewall settings, e.g. for ufw by adding the following rule:

```
ufw allow proto udp from 10.1.24.0/24
```

## Ideas for Next Iterations (Not This PR)

Dynamic tab completions for pepsi subcommands such as `pepsi upload` and `pepsi poweroff` using the aliveness information are already WIP, but have been factored out to a different PR.

## How to Test

First, execute `cargo build` in `tools/aliveness` after sourcing the SDK.
Copy the target to a NAO, ssh into it and execute the target.
Now you can try any of the pepsi aliveness subcommands, e.g.

```
./pepsi aliveness summary
./pepsi aliveness services --json
./pepsi aliveness battery --timeout 500
./pepsi aliveness all 32
```

Make sure your firewall allows the incoming responses as described above if you don't see any responses.

PR #31, opened 2022-12-17
Hula next

schmidma:
## Introduced Changes

Removes the aliveness capabilities from hula.
Aliveness will be added in a future PR. (probably #27)

## ToDo / Known Issues

none

- implement aliveness

## How to Test

- flash an image with this hula (ask @schmidma)
- upload the code, see if everything still works


tools/hula/proxy/src/proxy.rs
```diff
@@ -0,0 +1,222 @@
...
+        let proxy_start = Instant::now();
+        let mut connections = HashMap::new();
+        let mut events = [Event::new(Events::empty(), 0); 16];
+        let mut writer = BufWriter::with_capacity(786, self.lola.try_clone()?);
+        let mut battery = None;
+
+        debug!("Entering epoll loop...");
+        loop {
+            let epoll_timeout = -1;
+            let num_events = epoll::wait(self.epoll_fd, epoll_timeout, &mut events)
```
h3ndrk, line 75: Please don't abbreviate 

```suggestion
            let number_of_events = epoll::wait(self.epoll_file_descriptor, epoll_timeout, &mut events)
```

Please search and replace everywhere. Thanks.

tools/hula/proxy/src/proxy.rs
```diff
@@ -0,0 +1,222 @@
...
+                }
+            }
+
+            if !connections
+                .values()
+                .filter_map(|connection| connection.last_time_received)
+                .any(|instant| instant.elapsed() < HULA_TIMEOUT)
+            {
+                send_idle(&mut writer, &battery).wrap_err(
+                    "a shadowy flight into the dangerous world of a man who does not exist",
```
h3ndrk, line 94: Nice 😂

Resolve if read.

tools/hula/proxy/src/proxy.rs
```diff
@@ -0,0 +1,222 @@
...
+                error!("Failed to read from connection: {}", error);
+                info!("Removing connection with file descriptor {}", notified_fd);
+                // remove will drop, drop will close, close will EPOLL_CTL_DEL
+                connections
+                    .remove(&notified_fd)
+                    .expect("connection file descriptor has to be registered");
+                return Ok(());
+            };
+            connection.last_time_received = Some(Instant::now());
+            // reinterpret the read buffer as a ControlFrame
```
h3ndrk, line 192: Unnecessary comment? This is what the next line states?

tools/hula/proxy/src/proxy.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn run(mut self) -> Result<()> {
+        let proxy_start = Instant::now();
+        let mut connections = HashMap::new();
+        let mut events = [Event::new(Events::empty(), 0); 16];
+        let mut writer = BufWriter::with_capacity(786, self.lola.try_clone()?);
+        let mut battery = None;
+
+        debug!("Entering epoll loop...");
+        loop {
+            let epoll_timeout = -1;
```
h3ndrk, line 74:
```suggestion
            const EPOLL_TIMEOUT: ... = -1;
```

tools/hula/proxy/src/proxy.rs
```diff
@@ -0,0 +1,222 @@
...
+                } else {
+                    handle_connection_event(&mut connections, notified_fd, &mut writer, &battery)?;
+                }
+            }
+
+            if !connections
+                .values()
+                .filter_map(|connection| connection.last_time_received)
+                .any(|instant| instant.elapsed() < HULA_TIMEOUT)
+            {
```
h3ndrk, line 92: You changed this condition from `connections.is_empty()` to this one. I'm not happy with that change. IMHO, if at least there is one connection, this connection should have exclusive access to LOLA without HULA injecting control messages. Please change to the old predicate.

h3ndrk (changes requested):

tools/hula/proxy/src/listener.rs
```diff
@@ -0,0 +1,40 @@
...
+    os::unix::net,
+};
+
+pub struct HulaListener {
+    listener: net::UnixListener,
+    path: &'static str,
+}
+
+impl HulaListener {
+    pub fn bind(path: &'static str) -> io::Result<HulaListener> {
```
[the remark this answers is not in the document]
schmidma: done

tools/hula/proxy/src/listener.rs
```diff
@@ -0,0 +1,40 @@
...
+        Ok(Self {
+            listener: net::UnixListener::bind(path)?,
+            path,
+        })
+    }
+}
+
+impl Deref for HulaListener {
+    type Target = net::UnixListener;
```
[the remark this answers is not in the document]
schmidma: done

tools/hula/proxy/src/listener.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.listener
+    }
+}
+
+impl Drop for HulaListener {
+    fn drop(&mut self) {
+        remove_file(self.path).expect("failed to remove listening unix socket");
+    }
+}
```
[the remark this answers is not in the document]
schmidma: ahhh... yes... this is currently not used. This drop is never executed, iirc. Killing the application just kills it, without further cleanup. Maybe we should discuss the details in person

tools/hula/proxy/src/proxy.rs
schmidma: boah... alter fine.. ... done

tools/hula/proxy/src/proxy.rs
schmidma: removed

tools/hula/proxy/src/proxy.rs
schmidma: moved to `NO_EPOLL_TIMEOUT`

tools/hula/proxy/src/proxy.rs
schmidma: No, I disagree. In the future, aliveness servers will attach without ever sending any data

schmidma: I agree, that this is not pretty, if you pause the framework (e.g. for debugging purposes). Do you have ideas?

h3ndrk: Okay, I understand your point but we need a compromise because both our points are important.

schmidma: What do you think of making the hula endpoints unidirectional? one for sending and one for receiving data?

tools/hula/proxy/src/proxy.rs
h3ndrk: You fixed `number_of_events` but we still have a lot of `fd` occurrences.

schmidma: I refuse to replace `fd`.  The code becomes extremely ugly and not readable anymore. Lines break, and the standard library itself uses `as_raw_fd()` this is very compact and repeatedly readable.

tools/hula/proxy/src/proxy.rs
h3ndrk: Hmmm, it would be one solution but I don't like that we would have another socket endpoint. What I could imagine is that all connected clients are read-only clients and are not counted. Once they send their first control message (e.g. our framework but not aliveness) they are opting in as being read-write and would be counted.

schmidma: That sounds nice! I'll try to implement that.

tools/hula/proxy/src/proxy.rs
h3ndrk: The standard library sadly is no benchmark. It has very bad naming in some places. For me it would be more important to have a good name without abbreviations than have pretty formatted code. Good names are also important for readability. I will create a PR on my own if you don't want to do it. Please resolve or write it out.

tools/hula/proxy/src/proxy.rs
schmidma: Done, thanks for this proposal ;)

h3ndrk: Do we want to move `connection.is_sending_control_frames = true;` earlier s.t. it does not depend on successful LoLA socket writing? I would count a connection as `is_sending_control_frames` if we receive one but not if we successfully delivered it to LoLA. :thinking:

schmidma: I think that barely matters. This error case is hopefully never triggered. We should not invest to much time on that issue

h3ndrk (approved):

PR #46, opened 2023-01-19
Add a progress indicator to pepsit

Narcha:
## Introduced Changes

This PR adds a progress indicator to pepsi's command line output.

## How to Test

Run the affected pepsi commands and observe their output.
Try running the commands on a different number of naos,
or introduce errors by attempting to execute commands on naos
which are unavailable.


tools/pepsi/src/aliveness.rs
```diff
@@ -26,15 +26,25 @@ pub async fn aliveness(arguments: Arguments) -> anyhow::Result<()> {
...
+                nao.set_aliveness(enable)
+                    .await
+                    .with_context(|| format!("Failed to set aliveness on {nao_address}")),
+            );
+        }
     });
 
-    let results = join_all(tasks).await;
-    gather_results(results, "Failed to execute some aliveness setting tasks")?;
+    join_all(tasks).await;
```
h3ndrk, line 47: Use [FuturesUnordered](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html) instead because the order does not matter probably?

For an example, look in the Communication server.

tools/pepsi/src/aliveness.rs
```diff
@@ -26,15 +26,25 @@ pub async fn aliveness(arguments: Arguments) -> anyhow::Result<()> {
...
 
-    let tasks = naos.into_iter().map(|nao_address| async move {
-        let nao = Nao::new(nao_address.ip);
-        nao.set_aliveness(enable)
-            .await
-            .with_context(|| format!("Failed to set aliveness on {nao_address}"))
+    let multi_progress = ProgressIndicator::new();
+
+    let tasks = naos.into_iter().map(|nao_address| {
+        let multi_progress = multi_progress.clone();
```
h3ndrk, line 32: @schmidma you once showed me a trick on how to clone variables to be used in threads and async tasks. Do you remember what that was? But this line looks quite good IMHO. 🤔

tools/pepsi/src/aliveness.rs
```diff
@@ -26,15 +26,25 @@ pub async fn aliveness(arguments: Arguments) -> anyhow::Result<()> {
...
+        async move {
+            let progress = multi_progress.task(nao_address.to_string());
+            progress.set_message("Setting aliveness...");
+
+            let nao = Nao::new(nao_address.ip);
+
+            progress.finish_with(
+                nao.set_aliveness(enable)
+                    .await
+                    .with_context(|| format!("Failed to set aliveness on {nao_address}")),
```
h3ndrk, line 42: We discussed in some Dev-Meeting that we want to transition to lowercase error messages by default. Do the case-style need to match the status messages? Maybe also write them lowercase?

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,67 @@
...
+            success_style: ProgressStyle::with_template("{prefix:.bold.dim} {wide_msg:.green}")
+                .unwrap(),
+        }
+    }
+
+    pub fn task(&self, prefix: String) -> Task {
+        let spinner = ProgressBar::new(10)
+            .with_style(self.default_style.clone())
+            .with_prefix(prefix);
+        spinner.enable_steady_tick(Duration::from_millis(100));
```
h3ndrk, line 31: I would be in favor of increasing that tick interval to e.g. 500 ms because we want to spend our CPU resources on executing our tasks instead of butterly smooth animations... 😉

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,67 @@
...
+    error_style: ProgressStyle,
+    success_style: ProgressStyle,
+}
+
+impl Task {
+    pub fn set_message(&self, message: impl Into<Cow<'static, str>>) {
+        self.progress.set_message(message)
+    }
+
+    pub fn finish_with_success(&self, message: impl std::fmt::Display) {
```
h3ndrk, line 51: Replace qualified path with use:

```suggestion
    pub fn finish_with_success(&self, message: impl Display) {
```

tools/pepsi/src/power_off.rs
```diff
@@ -14,15 +14,25 @@ pub struct Arguments {
...
 pub async fn power_off(arguments: Arguments) -> anyhow::Result<()> {
-    let tasks = arguments.naos.into_iter().map(|nao_address| async move {
-        let nao = Nao::new(nao_address.ip);
-        nao.power_off()
-            .await
-            .with_context(|| format!("Failed to power {nao_address} off"))
+    let multi_progress = ProgressIndicator::new();
+
+    let tasks = arguments.naos.into_iter().map(|nao_address| {
+        let multi_progress = multi_progress.clone();
```
h3ndrk, line 20: This cloning looks weird. Couldn't you just remove the cloning and directly get a task that you afterwards hand into the async task? Then there is no cloning involved in this location. You could also remove the `#[derive(Clone)]` from the multi progress because you probably want to have singleton behavior for this struct, right?

h3ndrk (changes requested):
That is a really cool PR, thanks for your contribution! I'm interested to see this by myself tomorrow. Probably this increased Pepsi's fancyness again by a lot! 💯

I left some small remarks. Feel free to discuss or just apply the comments. 😉

knoellle:
This branch has conflicts, please rebase and solve them.

tools/pepsi/src/progress_indicator.rs
knoellle: I disagree, I'm sure even your Gurke can handle updating some text ten times per second without meaningful impact on the actual tasks.

tools/pepsi/src/aliveness.rs
Narcha: This was fixed after rebasing on main.

tools/pepsi/src/aliveness.rs
schmidma: I think there is no 'trick'. You have to clone the variable. IIRC the only remark I made on your code in the past was to clone the variable in an extra scope, naming it  equally and thus shadow the previous variable. This removes these odd `variable2` or `variable_cloned` naming.
IMHO, this looks fine.

h3ndrk: Thanks for your input. Resolving.

tools/pepsi/src/power_off.rs
Narcha: I need the cloned `multi_progress` because it is `move`d into the `async move` block. If I move the line with `let progress = multi_progress.task(...)` outside of the `async{}` block, we sometimes observe "zombie" progress bar lines probably caused by some race condition.

tools/pepsi/src/aliveness.rs
julianschuler: Thanks, fixed.

tools/pepsi/src/power_off.rs
julianschuler: The "zombie" progress bar lines are now fixed and the cloning removed.

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,121 @@
...
+        Task {
+            progress: self.multi_progress.add(spinner),
+            error_style: self.error_style.clone(),
+            success_style: self.success_style.clone(),
+        }
+    }
+
+    pub async fn map_tasks<T, F, M>(
+        items: impl IntoIterator<Item = T>,
+        message: impl Into<Cow<'static, str>> + Clone,
```
h3ndrk, line 41: Why not?

```suggestion
        message: &'static str,
```

Do we want to expose the type complexity of indicatif or provide a type interface how we are using it in our code base?

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,121 @@
...
+        task: impl Fn(T) -> F + Copy,
+    ) where
+        T: ToString + Clone,
```
h3ndrk, line 44: Why do you need `Clone`?

```suggestion
        T: ToString,
```

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,121 @@
...
+            .collect::<FuturesUnordered<_>>()
+            .collect::<Vec<_>>()
+            .await;
+    }
+}
+
+pub struct Task {
+    progress: ProgressBar,
+    error_style: ProgressStyle,
+    success_style: ProgressStyle,
```
h3ndrk, line 66: Reorder these? Elsewhere you are handling the success first. Be consistent.

tools/pepsi/src/wireless.rs
```diff
@@ -44,66 +43,50 @@ pub async fn wireless(arguments: Arguments) -> Result<()> {
...
-    });
-
-    let results = join_all(tasks).await;
-    gather_results(results, "Failed to execute some network setting tasks")?;
-
-    Ok(())
+async fn set(naos: Vec<NaoAddress>, network: Network) {
+    ProgressIndicator::map_tasks(
+        naos,
+        "Retrieving available networks...",
```
h3ndrk, line 83: "Retrieving" in the `set()` function?

tools/pepsi/src/upload.rs
```diff
@@ -63,58 +121,48 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> Result<()>
...
-                .await
-                .wrap_err_with(|| format!("failed to power {nao_address} off"))
-        }
-    });
-
-    let results = join_all(tasks).await;
-    gather_results(results, "failed to execute some upload tasks")?;
+    arguments
+        .naos
+        .iter()
```
h3ndrk, line 138: Why does this not use your new infrastructure?

tools/pepsi/src/parsers.rs
```diff
@@ -175,6 +175,12 @@ impl FromStr for NaoNumberPlayerAssignment {
     }
 }
 
+impl Display for NaoNumberPlayerAssignment {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
+        self.nao_number.fmt(formatter)
+    }
+}
```
h3ndrk, line 182: You are providing an implementation detail as public functionality. It would be better that you just have this at the location where you are using it but not at this location. If I don't know your implementation, why would a number assignment only print the number?!

h3ndrk (changes requested):

tools/pepsi/src/upload.rs
julianschuler: This is due to `upload_with_progress` requiring a reference to `progress`. I was unable to rewrite `map_tasks()` in such a way to accommodate that due to the closure being executed in an async context.

tools/pepsi/src/parsers.rs
julianschuler: Alright, I now changed it to output "nao_number:player_number".

tools/pepsi/src/progress_indicator.rs
```diff
@@ -0,0 +1,121 @@
...
+            default_style: ProgressStyle::with_template("{prefix:.bold.dim} {spinner} {msg}")
+                .unwrap()
+                // The last char is ignored as it provides a final state
+                .tick_chars("⠏⠋⠙⠹⢸⣰⣠⣄⣆⡇ "),
+            error_style: ProgressStyle::with_template("{prefix:.bold.dim} {msg}").unwrap(),
+            success_style: ProgressStyle::with_template("{prefix:.bold.dim} {msg}").unwrap(),
+        }
+    }
+
+    pub fn task(&self, prefix: String) -> Task {
```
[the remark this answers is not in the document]
julianschuler: This is required for `upload()`.

tools/pepsi/src/progress_indicator.rs
julianschuler: Thanks for the suggestion, done.

tools/pepsi/src/progress_indicator.rs
julianschuler: Artifact from earlier, fixed now.

tools/pepsi/src/progress_indicator.rs
julianschuler: Thanks, done.

tools/pepsi/src/wireless.rs
julianschuler: Thanks, fixed.

h3ndrk (approved):
Thanks @Narcha for initially proposing these changes and @julianschuler for getting it ready to merge!

PR #47, opened 2023-01-19
Check OS-Version

### LINKED ISSUE #35: Pepsi check image version
Pepsi should check whether the version of the image on a target NAO is the one that was expected before uploading.
If it isn't, either issue a warning or fail the operation (with an additional option to override).

pakonro:
## Introduced Changes
Prevent upload in case of unexpected os-version.  This check can be disabled by using '--skip-check-os'.

Fixes #35 

## How to Test

pepsi upload #nao-number
and
pepsi upload #nao-number --skip-check-os 

The first command should fail if nao's os-version doesn't match
the expected version and the execution can be forced by the second command.


tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
     /// The NAOs to upload to e.g. 20w or 10.1.24.22
     #[arg(required = true)]
     pub naos: Vec<NaoAddress>,
+    #[arg(long)]
+    pub skip_check_os: bool,
+}
+
+pub async fn check_os(nao: &Nao) -> bool {
```
h3ndrk, line 45: This should live in the `nao` crate. Then you can also revert the `pub` changes. The idea of the `nao` crate is to hide all low-level SSH stuff but exposed that now. 😉

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+            println!("Unstable version on {}", nao.host);
+            println!("Use '--skip-check-os' if you still want to proceed");
+            print!("Installed os: {}, ", os_version);
+            println!("Stable os: {}", os_stable);
+
+            if os_version == os_stable {
+                return true;
+            } else {
+                return false;
+            }
```
h3ndrk, line 70: You could just:

```suggestion
            return os_version == os_stable;
```

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+        }
+    }
+
+    return true;
```
h3ndrk, line 74: Why is the fallback `true`? Shouldn't this be `false`?

tools/pepsi/src/upload.rs
```diff
@@ -93,7 +127,16 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> anyhow::Re
         let hulk_directory = hulk_directory.clone();
         async move {
             let nao = Nao::new(nao_address.ip);
+
+            // check os-version
```
h3ndrk, line 131: Please don't repeat yourself and make your code [DRY](https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself). Remove this comment, we can see what you are doing. 😉

tools/pepsi/src/upload.rs
```diff
@@ -127,3 +170,19 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> anyhow::Re
 
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+
```
h3ndrk, line 176: Don't you need `use super::*;`? 🤔

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+        .unwrap();
+
+    let request = String::from_utf8(output.stdout).unwrap();
+    let lines = request.lines();
+    for line in lines {
+        if line.contains("VERSION_ID") {
+            let line_split: Vec<&str> = line.split('=').collect();
+            let os_version = line_split[1];
+
+            println!("Unstable version on {}", nao.host);
```
h3ndrk, line 61: Could you just print `nao` instead of `nao.host`? Maybe have a look around how the other commands and the `nao` crate are printing NAOs.

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+            println!("Use '--skip-check-os' if you still want to proceed");
+            print!("Installed os: {}, ", os_version);
+            println!("Stable os: {}", os_stable);
```
h3ndrk, line 64: - Please correctly capitalize OS.
- You can put the variable into the format string.
- Write single line prints in one line. It better to see the line string in one source code line IMHO.

```suggestion
            println!("Installed OS: {os_version}, stable OS: {os_stable}");
```

h3ndrk (changes requested):
Thanks for your contribution! 🎉

I only have some small remarks to ensure our high source code quality. But in general the code looks very good and I very much like that you have a test case for the extraction function.

Well done! 💯

Please implement the requested changes but also feel free to discuss. 😃

tools/pepsi/src/upload.rs
h3ndrk, line 45: Rename this function to e.g. `has_stable_os_version()`?

h3ndrk (changes requested):

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+        .output()
+        .await
+        .unwrap();
+
+    let request = String::from_utf8(output.stdout).unwrap();
+    let lines = request.lines();
+    for line in lines {
+        if line.contains("VERSION_ID") {
+            let line_split: Vec<&str> = line.split('=').collect();
+            let os_version = line_split[1];
```
h3ndrk, line 59: Just use [`.split_once()`](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once) if you are only splitting once and are interested in one of the two parts. Something like this:

```suggestion
            let Some((_, os_version)) = line.split_once('=') else { continue; };
```

Instead of `continue` you could also return an error from the function.

tools/pepsi/src/upload.rs
```diff
@@ -127,3 +170,19 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> anyhow::Re
 
     Ok(())
 }
+
+#[cfg(test)]
+mod test {
+
+    fn test() {
```
h3ndrk, line 177: - You need a `#[test]` in the line before to mark this function as a test.
- Please name test functions after the fact you are testing. Also apply DRY and don't write the word `test` as this is a test, we can see that. Your fact is probably something like that the extracted version matches the version of the os-release file.

tools/pepsi/src/upload.rs
```diff
@@ -127,3 +170,19 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> anyhow::Re
...
+        let input = r#"ID=hulks-os
+    NAME="HULKs-OS"
+    VERSION="5.1.3 (langdale)"
+    VERSION_ID=5.1.3
+    PRETTY_NAME="HULKs-OS 5.1.3 (langdale)"
+    DISTRO_CODENAME="langdale"#;
+
+        let output = extractVersionNum(input);
```
h3ndrk, line 185: Is this function defined somewhere or did I overlook it?

h3ndrk (changes requested):

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
     /// The NAOs to upload to e.g. 20w or 10.1.24.22
     #[arg(required = true)]
     pub naos: Vec<NaoAddress>,
+    #[arg(long)]
+    pub skip_check_os: bool,
+}
+
+pub async fn check_os(nao: &Nao) -> bool {
+    let os_stable = "5.1.1";
```
knoellle, line 46: The latest OS version is `5.2.0`, [released](https://github.com/HULKs/meta-hulks/releases) two days ago.

tools/pepsi/src/upload.rs
knoellle, line 64: You are always printing these lines, regardless of whether the version matches or not, leading to output like this:
```
Unstable version on 10.1.24.28
Use '--skip-check-os' if you still want to proceed
Installed os: 5.2.0, Stable os: 5.2.0
Starting upload to 10.1.24.28
```

tools/pepsi/src/upload.rs
knoellle: Agreed. Additionally, when the check fails because no version was detected, this should be indicated to the user.

knoellle (changes requested):

tools/pepsi/src/upload.rs
pakonro: Done

tools/pepsi/src/upload.rs
pakonro: Corrected

tools/pepsi/src/upload.rs
pakonro: Ok

tools/pepsi/src/upload.rs
pakonro: Yes!

tools/pepsi/src/upload.rs
pakonro: Ok

tools/pepsi/src/upload.rs
pakonro: Ok

tools/pepsi/src/upload.rs
pakonro: Renamed

tools/pepsi/src/upload.rs
```diff
@@ -38,6 +38,40 @@ pub struct Arguments {
...
+    let output = nao
+        .ssh_to_nao()
+        .arg("cat /etc/os-release")
+        .output()
+        .await
+        .unwrap();
+
+    let request = String::from_utf8(output.stdout).unwrap();
+    let lines = request.lines();
```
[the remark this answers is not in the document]
pakonro: Renamed

tools/pepsi/src/upload.rs
pakonro: Ok

tools/pepsi/src/upload.rs
pakonro: Ok

tools/pepsi/src/upload.rs
pakonro: Function re-inserted

tools/pepsi/src/upload.rs
pakonro: Corrected

tools/pepsi/src/upload.rs
h3ndrk: Please also revert the `pub` changes.

tools/pepsi/src/upload.rs
h3ndrk: It didn't work because you didn't change anything?

tools/pepsi/src/upload.rs
h3ndrk: Please be more on-point with the test case name. I mentioned earlier that you do not need to write comments everywhere because our goal is that the code documents itself. This requires names to be good. "Extracts version num" is not the fact you are testing. Test cases check one fact of e.g. a unit. One fact may need to be checked with one or multiple assertions. In your case, you have a single assertion which also gives you the fact and this should be expressed in the test case name. Please improve the name.

crates/nao/src/lib.rs
```diff
@@ -11,14 +11,40 @@ use color_eyre::{
...
+    pub host: Ipv4Addr,
 }
 
 impl Nao {
     pub fn new(host: Ipv4Addr) -> Self {
         Self { host }
     }
 
+    pub async fn has_stable_os_version(&self) -> bool {
+        let os_stable = "5.2.0";
```
h3ndrk, line 23: @knoellle What do you think about this introduction of a new magic value? Now we need to maintain the version number twice: 1. in the `repository` crate and 2. here in the `nao` crate. We should have a single point of truth. Shared crate? Shared configuration file (reading the configuration file can also be deduplicated in a shared crate)?

crates/nao/src/lib.rs
```diff
@@ -11,14 +11,40 @@ use color_eyre::{
...
+        let output = self
+            .ssh_to_nao()
+            .arg("cat /etc/os-release")
+            .output()
+            .await
+            .unwrap();
+
+        let stdout = String::from_utf8(output.stdout).unwrap();
```
h3ndrk, line 31: You are creating a `String` but then take a `&str` of it. Iirc you can just use `str::from_utf8()`.

In addition, we decided in a recent Dev-Meeting that we aim to use all things instead of using qualified paths. Insert `use str::from_utf8;` to your imports and remove the `str::` prefix from your call.

crates/nao/src/lib.rs
```diff
@@ -333,3 +359,33 @@ impl Display for Network {
...
+    let lines = input.lines();
+        for line in lines {
+            if line.contains("VERSION_ID") {
+                let Some((_, os_version)) = line.split_once('=') else { continue; };
+		return Some(os_version.to_string());
+            }
+        }
+   None
+}    
+#[cfg(test)]
```
h3ndrk, line 373: I noticed a lot of whitespace errors in your PR. Around functions, impls, structs, enums we want to have one empty line. Within curly braces we want no empty lines in the beginning and end of the containing items.

Also, your code is not formatted. We are using automatic formatters for this. Either press Ctrl+Shift+I in your VS Code or run `cargo fmt` on the command line. Beware that the formatting does not fix empty line mistakes.

crates/nao/src/lib.rs
```diff
@@ -333,3 +359,33 @@ impl Display for Network {
         }
     }
 }
+
+fn extract_version_num(input:&str) -> Option<String>{
```
h3ndrk, line 363: Please don't abbreviate!

```suggestion
fn extract_version_number(input:&str) -> Option<String>{
```

tools/pepsi/src/upload.rs
```diff
@@ -93,7 +95,14 @@ pub async fn upload(arguments: Arguments, repository: &Repository) -> Result<()>
...
         async move {
             let nao = Nao::new(nao_address.ip);
+
+            if !arguments.skip_has_stable_os_version {
+                if !nao.has_stable_os_version().await {
+                    return Ok(());
+                }
+            }
             println!("Starting upload to {nao_address}");
+
```
h3ndrk, line 105: If you insert empty lines, you are grouping statements. Always think about how you are grouping them and if they belong together as this structures your code. Here, you changed the grouping but I would argue that the "Starting upload" message does not belong to your version check but rather to the upload process afterwards.

h3ndrk (changes requested):

h3ndrk:
One general remark: There exist a lot more strict commit message conventions than we have them at the HULKs. But a minimum standard would still be great. If you commit, please try to summarize shortly in one sentence what you did. If this is too hard, maybe your commit is too large.

crates/hulks_os_version/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "hulks_os_version"
+version = "0.1.0"
+edition = "2021"
```
h3ndrk, line 4: Please look into the other crate's `Cargo.toml`s. We are trying to...

- sort the keys alphabetically
- add a GPLv3 license

Probably copy-🍝 from another crate and change the name.

crates/hulks_os_version/Cargo.toml
```diff
@@ -0,0 +1,8 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
h3ndrk, line 8: Remove these lines, because the comment is irrelevant and we don't have dependencies. The person that introduces the first dependency can add the dependency section again.

crates/nao/Cargo.toml
```diff
@@ -7,3 +7,4 @@ version = "0.1.0"
 [dependencies]
 color-eyre = { workspace = true }
 tokio = { workspace = true }
+hulks_os_version = { workspace = true }
```
h3ndrk, line 10: Please sort dependencies alphabetically.

General remark: If you have any kind of list or enumeration of related things (not just in code), always think about a reasonable order. If you cannot find a easily understandable semantic order from the context of the list, fallback to alphabetical sorting. Alphabetically sorting also allows to easily see if an item is present or missing. The others will thank you for spending a moment thinking about this. 😉

tools/pepsi/src/pre_game.rs
```diff
@@ -47,6 +47,7 @@ pub struct Arguments {
     /// The NAOs to upload to with player number assignments e.g. 20w:2 or 10.1.24.22:5 (player numbers start from 1)
     #[arg(required = true)]
     pub assignments: Vec<NaoAddressPlayerAssignment>,
+    pub skip_has_stable_os_version: bool,
```
h3ndrk, line 50: You changed this flag's name. Could we improve that you currently have two verbs after another in the beginning? Maybe "skip stable os version check" or so...

h3ndrk (changes requested):

crates/hulks_os_version/Cargo.toml
Bendodroid: Or have a look at https://github.com/DevinR528/cargo-sort

tools/pepsi/src/pre_game.rs
knoellle: I would drop the "stable" and perhaps "version" too.
`skip-os-check` or `skip-os-version-check` should be sufficient.

"stable" only describes one of the use cases for this check, we want to avoid uploading to a NAO with an outdated OS image as well as one that has an newer one.
The latter case may be because the branch we compiled isn't based on the latest main branch or because the robot has an experimental image. "stable" only applies to this last, experimental case IMO.

crates/hulks_os_version/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub static HULKS_OS_VERSION: &str = "5.3.0";
```
julianschuler, line 1: There is a constant `SDK_VERSION` in the repository crate, which has the same value as the HULKs-OS version. I think you can use it instead of creating this specific new crate.

crates/hulks_os_version/Cargo.toml
pakonro: Alphabetically sorted. Please check if the license name I added is correct

crates/hulks_os_version/Cargo.toml
pakonro: I removed the lines

tools/pepsi/src/upload.rs
pakonro: I renamed the test to 'matches_stable_os_version'

crates/hulks_os_version/src/lib.rs
h3ndrk: Yes, please do so. Julian's PR got merged earlier and introduced a new crate for constants with a different name that you can use. Just put your constant in there.

knoellle (approved):
LGTM

PR #48, opened 2023-01-20
Fixed webots collision rotations

MaikRe:
## Introduced Changes

Fixed rotation of arm and head collision cylinders. Changed foot collision mesh to dual cylinders to improve standup front/back.

Further improvements on collision meshes to come.


jonathan-hellwig (approved):
Look good :D

PR #51, opened 2023-01-21
implement `SerializeHierarchy` for `VectorN<T>`

Vivituhh:
Co-authored-by: okiwi6 <45100017+okiwi6@users.noreply.github.com>

## Introduced Changes

Implement `SerializeHierarchy` for `Vector2<T>`, `Vector3<T>`, `Vector4<T>` in order to serialze vectors into more specific paths (xy/xyz/xyzw).

## Fixes 

This enables for example plotting vector components in twix

## Further ideas

Also manually implement `SerializeHierarchy` for `Point`

### DIFF
--- a/crates/serialize_hierarchy/src/lib.rs
+++ b/crates/serialize_hierarchy/src/lib.rs
@@ -284,7 +284,7 @@ impl<T: Serialize + DeserializeOwned, const N: usize> SerializeHierarchy
     fn deserialize_path<S>(
         &mut self,
         path: &str,
-        _data: S::Serialized,
+        data: S::Serialized,
     ) -> Result<(), Error<S::Error>>
     where
         S: Serializer,
@@ -295,7 +295,7 @@ impl<T: Serialize + DeserializeOwned, const N: usize> SerializeHierarchy
             .position(|name| name == &path);
         match index {
             Some(index) => {
-                let deserialized = S::deserialize(_data).map_err(Error::DeserializationFailed)?;
+                let deserialized = S::deserialize(data).map_err(Error::DeserializationFailed)?;
                 self[index] = deserialized;
                 Ok(())
             }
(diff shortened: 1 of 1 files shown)


crates/serialize_hierarchy/src/lib.rs
```diff
@@ -262,6 +262,144 @@ impl<T> SerializeHierarchy for Vec<T> {
...
+        S: Serializer,
+        S::Error: error::Error,
+    {
+        let idx = ["x", "y"].into_iter().position(|x| x == path);
+        match idx {
+            Some(i) => S::serialize(&self[i]).map_err(Error::SerializationFailed),
+            _ => Err(Error::UnexpectedPathSegment {
+                segment: String::from(path),
+            }),
+        }
```
h3ndrk, line 277: Please don't abbreviate.

```suggestion
        let index = ["x", "y"].into_iter().position(|name| name == path);
        match index {
            Some(index) => S::serialize(&self[index]).map_err(Error::SerializationFailed),
            _ => Err(Error::UnexpectedPathSegment {
                segment: String::from(path),
            }),
        }
```

Please also apply this in the other cases.

crates/serialize_hierarchy/src/lib.rs
```diff
@@ -262,6 +262,144 @@ impl<T> SerializeHierarchy for Vec<T> {
...
+            }
+            None => Err(Error::UnexpectedPathSegment {
+                segment: String::from(path),
+            }),
+        }
+    }
+
+    fn exists(_path: &str) -> bool {
+        matches!(_path, "x" | "y")
+    }
```
h3ndrk, line 304: Remove the underscore:

```suggestion
    fn exists(path: &str) -> bool {
        matches!(path, "x" | "y")
    }
```

h3ndrk (changes requested):
Some minor naming remarks. Feel free to comment or just implement the requested changes. 😉

h3ndrk:
What do you think about also implementing `SerializeHierarchy` manually for `PointN`?

Vivituhh:
Done 👍

crates/serialize_hierarchy/src/lib.rs
```diff
@@ -262,6 +262,240 @@ impl<T> SerializeHierarchy for Vec<T> {
...
+    fn exists(path: &str) -> bool {
+        matches!(path, "x" | "y" | "z" | "w")
+    }
+
+    fn get_fields() -> BTreeSet<String> {
+        ["x", "y", "z", "w"].into_iter().map(String::from).collect()
+    }
+}
+
+impl<T: Serialize + DeserializeOwned + Clone + Scalar> SerializeHierarchy for Point2<T> {
```
h3ndrk, line 407: I think we can reduce some code duplication for `Point` since points contain [`coords: Vector`](https://docs.rs/nalgebra/latest/nalgebra/geometry/struct.OPoint.html#structfield.coords). It should be possible to delegate the trait method calls to the `coords` field instead of manually implementing it again, right?

h3ndrk (changes requested):
Let's try to reduce a little bit of the code duplication.

If you want, you could also try to implement the trait in a more generic way instead of manually implementing it for 2,3,4 dimensional vectors and points.

In addition, please do not merge the `main` branch into your feature branch but rather rebase it. 🙏

@knoellle Please announce to our team that we prefer rebases and should not click the "Update branch" button.

h3ndrk (approved):
Absolut Oberaffengeil

PR #65, opened 2023-01-26
Add and switch to ONNX

tuxbotix:
## Introduced Changes

Introduced ONNX models. Used basic `tf2onnx` utility, there are other ways too ;).

I would still keep the hdf5 models for now, at least until we figure out a good way to version & trace our model development!

Fixes #

CompiledNN has ONNX :P we want to use that!
In addition, the pure-rust tract-onnx inference backend can directly use these ;)

## How to Test

We should run a full round of test and see.
I'll run the micro-benchmarks I have and see if there is a performance difference.


tuxbotix:
As suspected, the ONNX version/ proto they are using is rather old as far as I can see.
I tried reducing the onnx opset version to an older one, no luck.

Need to test with a smaller example and see why. These models run fine with `tract-onnx`

ahasselbring:
Unfortunately the ONNX implementation is not more advanced than this: https://github.com/bhuman/CompiledNN/blob/master/Src/CompiledNN/Formats/ONNX.cpp#L35

Creating the ONNX frontend is one of the parts of the ongoing struggle whether more time should be invested into this version of CompiledNN or whether CompiledNN should be rebuilt e.g. based on MLIR (which would allow us to use existing frontends such as the MLIR ONNX dialect).

tuxbotix:
@ahasselbring I understand the struggle of deciding what path to take. I also think a MLIR or similar approach might work out better in the long run as the ecosystem has grown a bit more.

tuxbotix:
Closing as CompiledNN does not have a complete ONNX frontend yet. See above comments for more information.

PR #70, opened 2023-01-28
Pre-Allocate segments

schmidma:
## Introduced Changes

This allocates memory for the segments vector before the loop starts.

## ToDo / Known Issues

none

none

## How to Test

This should change nothing, but runtime. You can profile it


knoellle (approved):
LGTM

PR #97, opened 2023-02-08
Improve walking parameters

knoellle:
## Introduced Changes

Improve walking parameters


schmidma (approved):
We should probably do a thorough tuning of walking parameters for the SMD

PR #98, opened 2023-02-09
Dynamic pepsi completion using aliveness

julianschuler:
## Introduced Changes

This PR adds dynamic shell completions for all pepsi subcommands that take NAO addresses as a parameter (e.g. `pepsi upload`).

The NAO addresses to complete with are retrieved using the aliveness service, so only alive NAOs will be suggested. It requires pepsi to be installed to the `PATH`.

The shells fish, zsh and bash are supported.

## ToDo / Known Issues

Nothing here.

None here.

## How to Test

Build and install the latest version of pepsi, e.g. with

```
cargo install --path tools/pepsi
```

and by adding `~/.cargo/bin` to the `PATH`.

Now, generate the completions for your favourite shell, e.g. for fish using

```
./pepsi completions fish > ~/.config/fish/completions/pepsi.fish
```

Restart the shell, make sure at least one NAO is alive and observe the behavior of hitting Tab after e.g. `pepsi upload`.

### DIFF
--- a/tools/pepsi/src/completions.rs
+++ b/tools/pepsi/src/completions.rs
@@ -9,12 +9,14 @@ use crate::aliveness::completions as complete_naos;
 pub struct Arguments {
     #[arg(long, hide = true)]
     complete_naos: bool,
+    #[arg(long, hide = true)]
+    complete_assignments: bool,
     #[clap(name = "shell")]
     pub shell: clap_complete::shells::Shell,
 }
 
 pub async fn completions(arguments: Arguments, mut command: Command) -> Result<()> {
-    if arguments.complete_naos {
+    if arguments.complete_naos || arguments.complete_assignments {
         let naos = complete_naos().await?;
 
         let separator = match arguments.shell {
@@ -23,9 +25,13 @@ pub async fn completions(arguments: Arguments, mut command: Command) -> Result<(
             Shell::Zsh => '\n',
             _ => ' ',
         };
+        let colon = match arguments.complete_assignments {
+            true => ":",
+            false => "",
+        };
 
         for nao in naos {
-            print!("{nao}{separator}");
+            print!("{nao}{colon}{separator}");
         }
         return Ok(());
     }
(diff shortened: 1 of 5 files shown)


knoellle (changes requested):
Gib gib gib!

Tested with zsh and fish, works perfectly.
A couple commands are still missing completion though, will merge once that's fixed.

tools/pepsi/src/completions.rs
```diff
@@ -0,0 +1,98 @@
...
+                ("hulk", ""),
+                ("logs", "delete"),
+                ("logs", "downloads"),
+                ("postgame", ""),
+                ("poweroff", ""),
+                ("reboot", ""),
+                ("upload", ""),
+                ("wireless", "list"),
+                ("wireless", "set"),
+                ("wireless", "status"),
```
knoellle, line 67: 👀 somehow github dropped my comment here when submitting the review...

This list is missing `shell` and `pregame`.
Not quite sure about pregame though as it requires `robot:player_number` pairs.
It might still be a good idea to at least complete the player number though, what do you think?

julianschuler: You are right, I missed `shell` since its just a single NAO address. For `pregame` it might be a good idea to suggest e.g. `27:`, i.e. with colon. Since the amount of player numbers is limited, we could also suggest all combinations, i.e. `27:1`, `27:2`, `27:3`, `27:4` and `27:5` for the above example.

Both would add a bit of complexity, if this is not desired, we could just suggest the number. Which one do you prefer?

knoellle: I think the version with the trailing colon would be ideal. That increases ease of use while still indicating that the argument is incomplete.
Suggesting every combination can be quite noisy considering up to 12 robots and 7 player numbers.

julianschuler: Alright, done.

knoellle (approved):
LGTM

Issue #38, opened 2023-01-08
More active HULKs on the field


Bendodroid:
We want to have more active HULKs on the field during a game. 
This means less falling over, less bumping into other robots, and a faster stand-up motion if a fall occurs. 
Less bumping into other robots will also help reducing our penalties.
For this to happen we need better obstacle detection (including the damned goal posts).
See also:  Robot Detection #44

This topic is closely related to 7v7 Games #37

philipniklas-r:
Philip from the team B-Human here,

to the topic ``less falls`` and ``faster stand-up`` I could give some insights if you are interested.

For ``less falls``, judging from the code for the walking, I think https://github.com/HULKs/hulk/blob/main/crates/control/src/sole_pressure_filter.rs#L34 and https://github.com/HULKs/hulk/blob/main/crates/control/src/support_foot_estimation.rs#L52 could result in some problems, which _could_ explain a handful of falls at the last competition. With the low-pass filtering a detected support foot switch will be delayed. With the already present delay of about 3 motion frames between sending requests to the joints and the measured execution of such requests, adding more delay sounds like an unintended effect.

The usage of the hysterese, to decide which foot is the supporting one, is a good idea. But from lots of testing a hysterese should only be necessary if the FSRs of both feet measure close to 0 pressure. I am not sure, if this is only because we auto calibrate the FSRs values (similar to the original rUNSWift walk, but for both min and max values).

If both things are changed, I expect the actual walk duration to be closer to your intended 260 ms one (at the last RoboCup the robots looked like walking with a lot higher walk duration). From experience I can tell that this shorter duration is on average more stable, but _can_ result in other problems. Like worn out robots walking a lot slower than intended or your really strong forward kick resulting in a lot more falls. But those problems can be easily addressed.

h3ndrk:
Moin @philipniklas-r, thank you very much for your input! We are currently looking into your proposed changes. :tada:

PR #108, opened 2023-02-27
Use unfiltered FSR in walking

JonasHolz:
## Introduced Changes

- Use unfiltered FSR in walking-engine
- Move ground check outside the function
- Adjusted walking parameters

As suggested by @philipniklas-r

## ToDo / Known Issues

None

Check benefits for using weights for the FSRs

## How to Test

Check nao while walking for stability and speed
Lift up nao in order to check if ground contact recognition is working

### DIFF
--- a/crates/control/src/motion/walking_engine.rs
+++ b/crates/control/src/motion/walking_engine.rs
@@ -197,15 +197,11 @@ impl WalkingEngine {
             WalkState::Kicking(..) => self.kick_cycle(last_cycle_duration),
         }
 
+        let left_foot_pressure = context.sensor_data.force_sensitive_resistors.left.sum();
+        let right_foot_pressure = context.sensor_data.force_sensitive_resistors.right.sum();
         let has_support_changed = match self.swing_side {
-            Side::Left => {
-                context.sensor_data.force_sensitive_resistors.left.sum()
-                    > context.sensor_data.force_sensitive_resistors.right.sum()
-            }
-            Side::Right => {
-                context.sensor_data.force_sensitive_resistors.right.sum()
-                    > context.sensor_data.force_sensitive_resistors.left.sum()
-            }
+            Side::Left => left_foot_pressure > right_foot_pressure,
+            Side::Right => right_foot_pressure > left_foot_pressure,
         };
         if has_support_changed && self.t > context.config.minimal_step_duration {
             let deviation_from_plan = self
--- a/crates/control/src/motion/walking_engine.rs
+++ b/crates/control/src/motion/walking_engine.rs
@@ -167,21 +167,18 @@ impl WalkingEngine {
 
         let is_step_started_this_cycle = self.t.is_zero();
         if is_step_started_this_cycle {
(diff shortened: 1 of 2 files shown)


crates/control/src/motion/walking_engine.rs
```diff
@@ -186,9 +197,11 @@ impl WalkingEngine {
...
         }
 
-        let has_support_changed = match context.support_foot.support_side {
-            Some(support_side) => self.swing_side.opposite() != support_side,
-            None => true,
+        let left_foot_pressure = context.sensor_data.force_sensitive_resistors.left.sum();
+        let right_foot_pressure = context.sensor_data.force_sensitive_resistors.right.sum();
+        let has_support_changed = match self.swing_side {
+            Side::Left => left_foot_pressure > right_foot_pressure,
+            Side::Right => right_foot_pressure > left_foot_pressure,
```
philipniklas-r, line 204: Citing myself from the issue #38 : `But from lots of testing a hysterese should only be necessary if the FSRs of both feet measure close to 0 pressure`.  
This version here will probably work. But for us I noticed that without some hysterese (like a minimum foot pressure) around the 0 mark, meaning both feet measure close to nothing, the robot will walk in specific situations "not that good". If I remember correctly, those situations are when the robot is walking on his heels, e.g. after a push from the front which tilts the robot backwards. Also when turning in place (meaning walk steps with a lot of rotation, but no translation) the FSRs like to measure not that much, which caused too early foot support switches. 

Another thing I noticed is that most teams use weights for the FSRs (inner pressures are weighted different than outter ones). We use the same values rUNSWift provide, but never tested if those make a difference. Just to have it mentioned in case it could improve the decision for a support foot switch.

JonasHolz: Thanks for your remarks! We have a minimal_step_duration for each step, which works like a hysterese for preventing too early support foot switches. If the FSRs values are close to zero, because we lose ground contact, our has_ground_contact check prevents the robot from unwanted foot switches.

We will look into the benefits for using weights for the FSRs in the next iteration.

crates/control/src/motion/walking_engine.rs
```diff
@@ -165,13 +165,24 @@ impl WalkingEngine {
...
+                        context.kick_steps,
+                    );
+                }
+                false => {
+                    self.current_step = Step::zero();
+                    self.step_duration = Duration::ZERO;
+                    self.swing_side = Side::Left;
+                    self.max_swing_foot_lift = 0.0;
+                }
+            }
```
knoellle, line 185: Is there a reason why you chose a match statement here over an if?

JonasHolz: I just found it more readable, because you can differentiate better between the cases of has_ground_contact.

JonasHolz: I changed it to if else.

knoellle (approved):
Thank you for your contribution!

PR #110, opened 2023-03-05
Use binaries to represent nao and webots implementation

schmidma:
## Introduced Changes

Currently, the hulk binary crate splits the webots and nao implementation by features and a single binary.
Thus, only either nao or webots can be targeted.

This PR introduces an independent binary for each target (`crates/hulk/src/bin/[nao/webots].rs`).

## ToDo / Known Issues

none

none

## How to Test

Thankfully, we have a CI to check if everything still compiles.

To test the entire pipeline, use pepsi to upload to a nao (`pepsi upload xx`).

To test the webots binary, build, run and see if webots still executes correctly.
`pepsi run`
or
`pepsi build` and use the `hulk` controller in webots


h3ndrk:
What is the benefit of this?! It does not make sense to compile multiple binaries with the same compiler settings. We either want to compile for NAO or Webots but never both.

schmidma:
The benefit is, that features are not intended to be mutual exclusive. Tools like LSPs or clippy have issues analyzing our code this way. Currently they use `--all-features` which is not possible for the hulk crate. IMHO this is much cleaner an the intended way to support multiple targets.
I adjusted pepsi to still only build either webots or nao by introducing the `--bin webots` or `--bin nao` flag

crates/repository/src/lib.rs
```diff
@@ -102,7 +102,7 @@ impl Repository {
                 "--workspace --all-features --all-targets ".to_string()
             } else {
                 format!(
-                    "--features {target} --manifest-path={} ",
+                    "--features {target} --bin {target} --manifest-path={} ",
```
knoellle, line 105: You forgot to also add this flag to `pepsi run`. That currently fails with
```
Running: cargo run --profile incremental --features webots --manifest-path=/home/mystery/hulks/hulk/crates/hulk/Cargo.toml -- 
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: nao, webots
Error: failed to execute run command

Caused by:
   0: failed to run
   1: cargo command exited with exit status: 101

Location:
    /home/mystery/hulks/hulk/crates/repository/src/lib.rs:127:13
    ```

knoellle: Oh, nevermind. I used my `cargo install`ed version of pepsi. My bad.

knoellle (approved):
LGTM

PR #112, opened 2023-03-08
Restrict map zoom

knoellle:
## Introduced Changes

Previously the map panel would always zoom when scrolling anywhere within twix, even when another scroll area is focused.
Now zooming only works when the cursor is over the map panel.

Fixes #

## How to Test

Open map panel, open another panel next to it and try scrolling.


julianschuler (approved):
LGTM

PR #113, opened 2023-03-08
Add Pepsi error message when battery state is not available

### LINKED ISSUE #105: Missing error message in pepsi when battery state is not available via aliveness
When using `pepsi aliveness`, there is only an output for the battery if it is discharging or its charging level is below 95%. 
The battery status is retrieved from HuLA and might be unavailable. In this case, `pepsi aliveness` should also output a message, e.g. `󰁽 ?` (this is the same icon as for the discharging battery, if your browser can't display it).

The function `print_summary` in the file `tools/pepsi/src/aliveness.rs` has to be changed for implementing this.

JonasHolz:
## Introduced Changes

Add Pepsi error message when battery state is not available

Fixes #105

## ToDo / Known Issues

None

None

## How to Test

Run pepsi aliveness with a robot, which has over 95% charge or without running Hula.

### DIFF
--- a/tools/pepsi/src/aliveness.rs
+++ b/tools/pepsi/src/aliveness.rs
@@ -65,6 +65,7 @@ fn print_summary(states: &AlivenessList) {
     const DISCHARGING_ICON: &str = "󰁽";
     const OS_ICON: &str = "󱑞";
     const ALL_OK_ICON: &str = "✔";
+    const UNKNOWN_CHARGE: &str = "?";
 
     for (ip, state) in states.iter() {
         let id = match ip {
@@ -82,10 +83,14 @@ fn print_summary(states: &AlivenessList) {
                 } else {
                     DISCHARGING_ICON
                 };
-
                 output.push_str(&format!("{icon} {charge}%{:SPACING$}", ""))
             }
-        }
+        } else {
+            output.push_str(&format!(
+                "{DISCHARGING_ICON}{UNKNOWN_CHARGE}{:SPACING$}",
+                ""
+            ))
+        };
 
         let version = &state.hulks_os_version;
         if version != OS_VERSION {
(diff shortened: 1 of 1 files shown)


tools/pepsi/src/aliveness.rs
```diff
@@ -82,10 +83,14 @@ fn print_summary(states: &AlivenessList) {
...
-
                 output.push_str(&format!("{icon} {charge}%{:SPACING$}", ""))
             }
-        }
+        } else {
+            output.push_str(&format!(
+                "{DISCHARGING_ICON}{UNKNOWN_CHARGE}{:SPACING$}",
+                ""
+            ))
+        };
```
julianschuler, line 93:
```suggestion
        }
```

tools/pepsi/src/aliveness.rs
```diff
@@ -82,10 +83,14 @@ fn print_summary(states: &AlivenessList) {
                 } else {
                     DISCHARGING_ICON
                 };
-
                 output.push_str(&format!("{icon} {charge}%{:SPACING$}", ""))
             }
-        }
+        } else {
+            output.push_str(&format!(
+                "{DISCHARGING_ICON}{UNKNOWN_CHARGE}{:SPACING$}",
```
julianschuler, line 90: Consider merging `{DISCHARGING_ICON}{UNKNOWN_CHARGE}` into a single constant with the same content, e.g. `{UNKNOWN_CHARGE_ICON}`. This makes changing this one independent of changing the discharging icon.

julianschuler (changes requested):
Thanks for your contribution! I only have some minor remarks, feel to implement or discuss.

tools/pepsi/src/aliveness.rs
JonasHolz: Done!

tools/pepsi/src/aliveness.rs
JonasHolz: I merged them together.

julianschuler (approved):
LGTM!

Issue #128, opened 2023-03-08
Refactor Role-assignment as preparation for 7v7


PasGl:
The code for role-assignment needs clean-up.
The result of this clean-up should be a configuration-parameter for Roles to be used and their assignment-priority.

Issue #37, opened 2023-01-08
7v7 Games


Bendodroid:
To prepare for the 7v7 games we will be facing in the upcoming RoboCup, we need to make some improvements as adding two additional robots per team on the rather small field will lead to chaos otherwise.

- Players enum
- Init positions
- Ready positions
- New Roles

(old ... partially obsolete)
- The main thing will be to implement new kick decisions to pass the ball to other players to avoid having to walk long distances with the ball.
- For the passes to work reliably, we will need a robot-detection that can tell us where the opponent's robots are to avoid passing to the enemy team.
- Our kicks need tuning, we will need very accurate kicks over long distances to make it work well.
- Maybe our role assignment needs rethinking, just adding more defenders and an additional bishop might not be the most effective strategy.
- With two additional robots sending messages, we will need to think about the message budget and what to send over Wi-Fi.

knoellle:
Current rules can be found here: https://spl.robocup.org/downloads/
More specifically: https://spl.robocup.org/wp-content/uploads/SPL-Rules-master.pdf

PR #141, opened 2023-03-12
Role assignment refactoring

PasGl:
## Introduced Changes

Fixes: #128 

## ToDo / Known Issues

None

7vs7: #37 

## How to Test

Symptomatically, nothing should have changed

Has been tested in Behavior Simulator and seems to work fine.

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -96,9 +96,9 @@ impl RoleAssignment {
         {
             role = match context.player_number {
                 PlayerNumber::One => Role::Keeper,
-                PlayerNumber::Two => context.optional_roles[0],
-                PlayerNumber::Three => context.optional_roles[1],
-                PlayerNumber::Four => context.optional_roles[2],
+                PlayerNumber::Two => *context.optional_roles.get(0).unwrap_or(&Role::default()),
+                PlayerNumber::Three => *context.optional_roles.get(1).unwrap_or(&Role::default()),
+                PlayerNumber::Four => *context.optional_roles.get(2).unwrap_or(&Role::default()),
                 PlayerNumber::Five => Role::Striker,
             };
             self.role_initialized = true;
@@ -293,7 +293,7 @@ fn process_role_state_machine(
     game_controller_state: Option<&GameControllerState>,
     player_number: PlayerNumber,
     striker_trusts_team_ball: Duration,
-    optional_roles: &Vec<Role>,
+    optional_roles: &[Role],
 ) -> (Role, bool, Option<BallPosition>) {
     if let Some(game_controller_state) = game_controller_state {
         match game_controller_state.game_phase {
@@ -546,7 +546,7 @@ fn decide_if_claiming_striker_or_other_role(
     player_number: PlayerNumber,
     cycle_start_time: SystemTime,
     game_controller_state: Option<&GameControllerState>,
-    optional_roles: &Vec<Role>,
+    optional_roles: &[Role],
 ) -> (Role, bool, Option<BallPosition>) {
     if am_better_striker(
         current_pose,
@@ -634,7 +634,7 @@ fn generate_role(
     own_player_number: PlayerNumber,
     game_controller_state: Option<&GameControllerState>,
     striker_player_number: PlayerNumber,
-    optional_roles: &Vec<Role>,
+    optional_roles: &[Role],
 ) -> Role {
     if let Some(state) = game_controller_state {
         pick_role_with_penalties(
@@ -652,7 +652,7 @@ fn pick_role_with_penalties(
     own_player_number: PlayerNumber,
     penalties: &Players<Option<Penalty>>,
     striker_player_number: PlayerNumber,
-    optional_roles: &Vec<Role>,
+    optional_roles: &[Role],
 ) -> Role {
     let mut role_assignment: Players<Option<Role>> = Players {
         one: None,
@@ -662,11 +662,7 @@ fn pick_role_with_penalties(
         five: None,
     };
 
-    // assign essential roles: Keeper xor ReplacementKeeper, and Striker
-
-    // Striker
     role_assignment[striker_player_number] = Some(Role::Striker);
-    let mut keeper_or_replacement_keeper_found = false;
     let mut unassigned_robots = 4;
 
     if penalties[PlayerNumber::One].is_some() {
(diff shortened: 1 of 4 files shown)


crates/control/src/role_assignment.rs
```diff
@@ -97,9 +98,9 @@ impl RoleAssignment {
         {
             role = match context.player_number {
                 PlayerNumber::One => Role::Keeper,
-                PlayerNumber::Two => Role::DefenderRight,
-                PlayerNumber::Three => Role::StrikerSupporter,
-                PlayerNumber::Four => Role::DefenderLeft,
+                PlayerNumber::Two => context.optional_roles[0],
+                PlayerNumber::Three => context.optional_roles[1],
+                PlayerNumber::Four => context.optional_roles[2],
```
schmidma, line 103: This might panic, if the `Vec` is too short. Consider using `Vec::get(...).unwrap_or(...)` https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get

crates/control/src/role_assignment.rs
```diff
@@ -292,6 +295,7 @@ fn process_role_state_machine(
     game_controller_state: Option<&GameControllerState>,
     player_number: PlayerNumber,
     striker_trusts_team_ball: Duration,
+    optional_roles: &Vec<Role>,
```
schmidma, line 298:
```suggestion
    optional_roles: &[Role],
```
a slice is more generic than a reference to a `Vec`

crates/control/src/role_assignment.rs
```diff
@@ -367,6 +371,7 @@ fn process_role_state_machine(
                 player_number,
                 cycle_start_time,
                 game_controller_state,
+                optional_roles,
```
schmidma, line 374: This function takes many arguments. You should consider using a struct with descriptive names. But I'm also fine with the current approach. Feel free to discuss

crates/control/src/role_assignment.rs
```diff
@@ -536,6 +548,7 @@ fn decide_if_claiming_striker_or_other_role(
     player_number: PlayerNumber,
     cycle_start_time: SystemTime,
     game_controller_state: Option<&GameControllerState>,
+    optional_roles: &Vec<Role>,
```
schmidma, line 551: same as above: use a slice instead of `&Vec`

crates/control/src/role_assignment.rs
```diff
@@ -642,6 +663,10 @@ fn pick_role_with_penalties(
         four: None,
         five: None,
     };
+
+    // assign essential roles: Keeper xor ReplacementKeeper, and Striker
+
+    // Striker
```
schmidma, line 669: Remove these comments and replace them by properly named function calls.

crates/control/src/role_assignment.rs
```diff
@@ -662,114 +687,66 @@ fn pick_role_with_penalties(
         unassigned_robots -= 1;
     }
 
-    if unassigned_robots > 0
-        && role_assignment[PlayerNumber::One].is_none()
-        && penalties[PlayerNumber::One].is_none()
-    {
+    // Keeper
+    if unassigned_robots > 0 && needs_assignment(PlayerNumber::One, penalties, &role_assignment) {
```
schmidma, line 691: If you find yourself explaining code structure with comments, think about introducing smaller function scopes by extracting code passages into function calls.

crates/control/src/role_assignment.rs
```diff
@@ -662,114 +687,66 @@ fn pick_role_with_penalties(
...
-        role_assignment[PlayerNumber::Four] = Some(Role::StrikerSupporter);
-        unassigned_robots -= 1;
-    } else if unassigned_robots > 0
-        && role_assignment[PlayerNumber::Five].is_none()
-        && penalties[PlayerNumber::Five].is_none()
-    {
-        role_assignment[PlayerNumber::Five] = Some(Role::StrikerSupporter);
-        unassigned_robots -= 1;
+    if unassigned_robots > optional_roles.len() {
+        panic!("role_assignment not possible: not enough optional_roles available")
```
schmidma, line 724: Do you really want to panic here? Consider failing more softly instead of crashing the framework

crates/control/src/role_assignment.rs
```diff
@@ -662,114 +687,66 @@ fn pick_role_with_penalties(
...
-        && role_assignment[PlayerNumber::Four].is_none()
-        && penalties[PlayerNumber::Four].is_none()
-    {
-        role_assignment[PlayerNumber::Four] = Some(Role::DefenderLeft);
-    } else if unassigned_robots > 0
-        && role_assignment[PlayerNumber::Five].is_none()
-        && penalties[PlayerNumber::Five].is_none()
-    {
-        role_assignment[PlayerNumber::Five] = Some(Role::DefenderLeft);
+    while unassigned_robots > 0 {
```
schmidma, line 727: This should be a for loop over an iterator of the optional roles. I think about something like
`optional_roles.iter().take(unassigned_robots)`. Feel free to propose a solution using iterators or discuss it here.

schmidma: Using indices is often a characteristic in the code that possibly indicates a deeper problem

schmidma (changes requested):
A first iteration of my review with some questions and remarks. Feel free to discuss ;)

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
PasGl: Do you see any advantage of wrapping the arguments in a struct, other than getting rid of the `allow` ?
Or do you have a suggestion for more descriptive names ?

Currently my preference would be to leave it as is.

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
PasGl: Removed in favor of resorting to `Role::default()` if things fail.

crates/control/src/role_assignment.rs
PasGl: done

crates/control/src/role_assignment.rs
schmidma: There is an ongoing discussion on logging at #91 . I don't know what to conclude, but in my opinion this would be a case, where one should place a warning log?

crates/control/src/role_assignment.rs
```diff
@@ -97,9 +98,9 @@ impl RoleAssignment {
         {
             role = match context.player_number {
                 PlayerNumber::One => Role::Keeper,
-                PlayerNumber::Two => Role::DefenderRight,
-                PlayerNumber::Three => Role::StrikerSupporter,
-                PlayerNumber::Four => Role::DefenderLeft,
+                PlayerNumber::Two => *context.optional_roles.get(0).unwrap_or(&Role::default()),
```
schmidma, line 101: What do you think of something like this:

```suggestion
                PlayerNumber::Two => context
                    .optional_roles
                    .get(0)
                    .copied()
                    .unwrap_or_else(|| Role::default()),
```

crates/control/src/role_assignment.rs
PasGl: Yes. I think if the role-assignment results in a Striker-role, without the Player-number matching that of the Striker, we could/should do "something" ... logging, panic, warning, error ... all fine with me.

I'd let the ongoing discussion conclude and add the desired choice in the 7vs7 PR.

crates/control/src/role_assignment.rs
PasGl: done

PasGl: Clippy gave me:  `.unwrap_or_else(Role::default),`

done

crates/control/src/role_assignment.rs
PasGl: For now, I suggest resolving this.
Tests in Behavior Simulator look good.
If you do not have more change requests, this is ready for merge.

schmidma (approved):

PR #145, opened 2023-03-15
Non linear interpolator

okiwi6:
## Introduced Changes

Replaces the current interpolator by a spline.
At the moment there is no win using this interpolator over the old implementation, however we plan to use other Interpolation schemes, especially the `CatmullRom` scheme which can be handled by the library.
Currently the interpolation uses `Linear` because the keyframe files do not work with another interpolation scheme out of the box.

Additional keys are added in the creation that ensure a gradient of 0 at the start and end key, this however will be used only if higher order interpolations like `CatmullRom` are used, they do not have a effect for `Linear` interpolation.


crates/types/src/joints.rs
```diff
@@ -60,6 +60,17 @@ impl Add for HeadJoints {
     }
 }
 
+impl Sub for HeadJoints {
+    type Output = HeadJoints;
+
+    fn sub(self, rhs: Self) -> Self::Output {
```
h3ndrk, line 66: Please don't abbreviate 

```suggestion
    fn sub(self, other: Self) -> Self::Output {
```

Also for the other occurrences.

crates/types/src/motion_file.rs
```diff
@@ -3,13 +3,14 @@ use std::{fs::File, path::Path, time::Duration};
...
 
-use crate::{Joints, LinearInterpolator};
+use crate::Joints;
 
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct MotionFile {
-    initial_positions: Joints,
-    frames: Vec<MotionFileFrame>,
+    pub initial_positions: Joints,
+    pub frames: Vec<MotionFileFrame>,
```
h3ndrk, line 13: Is it really needed to make them `pub`? Same for the `MotionFileFrame`?

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
...
-            self.interpolator_index += 1;
-        }
-        self.interpolators[self.interpolator_index].step(remaining_time_step)
+        self.current_time += time_step;
+        self.value()
     }
 
     pub fn value(&self) -> Joints {
-        self.interpolators[self.interpolator_index].value()
+        let arg: f32 = self.current_time.as_secs_f32();
```
h3ndrk, line 98: Please don't abbreviate 

```suggestion
        let argument: f32 = self.current_time.as_secs_f32();
```

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
...
 impl MotionFileInterpolator {
     pub fn reset(&mut self) {
-        self.interpolators
-            .iter_mut()
-            .for_each(|interpolator| interpolator.reset());
-        self.interpolator_index = 0;
+        self.current_time = Duration::ZERO;
     }
 
     pub fn step(&mut self, time_step: Duration) -> Joints {
```
h3ndrk, line 92: You renamed another method as `advance_by`. Why is this named `step`? Shouldn't it be consistent?

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+        }));
+
+        Self::new(keys)
+    }
+}
+
+impl SplineMotionFileInterpolator {
+    pub fn new(keys: Vec<Key<Duration, Joints>>) -> Self {
+        assert!(keys.len() >= 2, "need at least two keys to interpolate");
+        //NOTE: assume keys are sorted
```
h3ndrk, line 43: Nobody will remember this comment in a few months. Could you express this in code or guarantee it somehow?

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+        keys.extend(motion_file.frames.into_iter().map(|frame| {
+            current_time += frame.duration;
+            Key::new(current_time, frame.positions, Interpolation::Linear)
+        }));
+
+        Self::new(keys)
+    }
+}
+
+impl SplineMotionFileInterpolator {
```
h3ndrk, line 40: Does this still have something to do with a motion file or could you drop the "motion file" from the name? 🤔

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+    interpolator: Spline<f32, Joints>,
+    current_time: Duration,
+    start_time: Duration,
+    end_time: Duration,
+}
+
+#[derive(Error, Debug)]
+pub enum InterpolatorError {
+    #[error("accessed interpolator defined for [{min}s, {max}s] at {current}s")]
+    OutOfBoundsAccess { min: f32, max: f32, current: f32 },
```
h3ndrk, line 17: - Please don't abbreviate 
- Are you using `f32` for time? This smells like [Primitive Obsession](https://refactoring.guru/smells/primitive-obsession).

h3ndrk (changes requested):
Low-level review for the moment. I will review again shortly.

crates/types/src/spline_motion_interpolator.rs
okiwi6: The library has to normalize timesteps. For a time s in [a,b], the normalized time is calculated by (s-a)/(b-a). Durations do not (yet) have a stable division implementation. Therefore we decided to convert to f32 internally, but do the conversion as late as possible. If you have other ideas how to tackle this we would be open for them.

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+                .nth(1)
+                .map(|key| key.value)
+        } else {
+            self.interpolator.sample(self.current_time.as_secs_f32())
+        }
+        .ok_or(InterpolatorError::OutOfBoundsAccess {
+            min: self.start_time.as_secs_f32(),
+            max: self.end_time.as_secs_f32(),
+            current: self.current_time.as_secs_f32(),
+        })
```
h3ndrk, line 97: Why do you need the error if you have the if-cases above?

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
+use splines::{Interpolation, Key, Spline};
+use thiserror::Error;
+
+use crate::{Joints, MotionFile};
+use std::time::Duration;
+
+pub struct SplineMotionFileInterpolator {
```
h3ndrk, line 7: Could you explain how the types relate together? It looks like you added another type and kept the existing ones. If I understood it correctly, the new spline interpolation only changes the implementation but not the interface. Would it be possible to use the new spline interpolation everywhere and therefore be able to remove the existing types? Deleted code is debugged code. 😉

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+        let left_helper = Key::new(
+            2. * keys[0].t.as_secs_f32() - keys[1].t.as_secs_f32(),
+            keys[1].value,
+            Interpolation::Linear,
+        );
+        let right_helper = Key::new(
+            2. * keys[last_key_index].t.as_secs_f32() - keys[last_key_index - 1].t.as_secs_f32(),
+            keys[last_key_index - 1].value,
+            Interpolation::Linear,
+        );
```
h3ndrk, line 59: Please reevaluate if this spreads same functionality that could be gathered. For example the `Key::new()` also exists in the other type. Why is there no single point that instantiates keys? Try to apply this reasoning to the rest of the code as well.

h3ndrk, line 59: What actually happens here? What are the helpers used for? Could you extract some code into methods to document it via method names?

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
     Ok(Duration::from_secs_f32(f32::deserialize(deserializer)?))
 }
 
+impl_Interpolate!(f32, Joints, std::f32::consts::PI);
```
h3ndrk, line 54: Replace qualified path with use?

```suggestion
impl_Interpolate!(f32, Joints, PI);
```

crates/types/src/spline_motion_interpolator.rs
h3ndrk: Okay, understood. But if I saw it correctly, you are converting it just before instantiating the error. Couldn't you just keep the durations and pass them in the error? There should be no spline stuff involved anymore? 🤔

h3ndrk (changes requested):
I had another look. Please focus on the more high-level comments first. For example the question about the new vs. existing types and whether same functionality could be abstracted into a single location.

crates/types/src/joints.rs
okiwi6: the `impl Add for Joints`, etc.. also uses `rhs`instead of `other`. Isn't `rhs` (although) abbreviated much more descriptive than `other`?

crates/types/src/motion_file.rs
okiwi6: Despite creation, these structs do not contain any methods and are only used to be converted into interpolators.
Is it okay to use `pub` then?

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
...
-        }
-        self.interpolators[self.interpolator_index].step(remaining_time_step)
+        self.current_time += time_step;
+        self.value()
     }
 
     pub fn value(&self) -> Joints {
-        self.interpolators[self.interpolator_index].value()
+        let arg: f32 = self.current_time.as_secs_f32();
+        self.interpolators.clamped_sample(arg).unwrap()
```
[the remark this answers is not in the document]
okiwi6: This will panic whenever there are not enough control points to perform the interpolation, for example if there is only one control points. Since here we only use `Linear` as interpolation method, this cannot panic as long as we have `initial_positions` and at least one animation frame.

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
...
-                        end_frame.positions,
-                        end_frame.duration,
-                    )
-                }),
-        );
+
+        keys.extend(motion_file.frames.into_iter().map(|frame| {
+            current_time += frame.duration;
+            Key::new(
+                current_time.as_secs_f32(),
```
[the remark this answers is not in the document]
okiwi6: Yes, that's the root cause of the problem.

crates/types/src/joints.rs
h3ndrk: We used `other` a lot already. If we decide on another name then we should also change the other occurrences. `right` would be okay for me.

crates/types/src/motion_file.rs
h3ndrk: Yes, then it is okay.

crates/types/src/motion_file.rs
okiwi6: Done

crates/types/src/spline_motion_interpolator.rs
okiwi6: Done

crates/types/src/spline_motion_interpolator.rs
okiwi6: Renamed

crates/types/src/spline_motion_interpolator.rs
okiwi6: Done

crates/control/src/motion/stand_up_back.rs
```diff
@@ -88,7 +88,7 @@ impl StandUpBack {
         }
 
         Ok(MainOutputs {
-            stand_up_back_positions: self.interpolator.value().into(),
+            stand_up_back_positions: self.interpolator.value()?.into(),
```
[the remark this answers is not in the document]
okiwi6: Done

crates/types/src/motion_file.rs
```diff
@@ -50,68 +51,55 @@ where
...
+                frame.positions,
+                Interpolation::Linear,
+            )
+        }));
+
         Self {
-            interpolators,
-            interpolator_index: 0,
+            interpolators: Spline::from_vec(keys),
```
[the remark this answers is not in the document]
okiwi6: Done

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
...
+use thiserror::Error;
+
+use crate::{Joints, MotionFile};
+use std::time::Duration;
+
+pub struct SplineMotionFileInterpolator {
+    interpolator: Spline<f32, Joints>,
+    current_time: Duration,
+    start_time: Duration,
+    end_time: Duration,
```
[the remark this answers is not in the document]
okiwi6: No they are required in the `value()` function

crates/types/src/spline_motion_interpolator.rs
```diff
@@ -0,0 +1,107 @@
+use splines::{Interpolation, Key, Spline};
+use thiserror::Error;
+
+use crate::{Joints, MotionFile};
+use std::time::Duration;
+
+pub struct SplineMotionFileInterpolator {
+    interpolator: Spline<f32, Joints>,
```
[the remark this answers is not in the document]
okiwi6: Done

crates/types/src/spline_motion_interpolator.rs
okiwi6: The other motion interpolator is removed now

crates/types/src/spline_motion_interpolator.rs
okiwi6: Extracted into method

crates/types/src/motion_file.rs
okiwi6: Done

crates/types/src/spline_motion_interpolator.rs
okiwi6: Solved by using an extension trait on interpolation scheme to translate `Interpolation<Duration, Joints>` to `Interpolation<f32, Joints>`

h3ndrk (approved):

PR #148, opened 2023-03-16
fix for incorrect formatting

okiwi6:
## Introduced Changes
Fix incorrect formatting of nao_address

Fixes #

Before the OS error message printed "failed to get OS version of {nao_address}", now the actual address is inserted.


knoellle (approved):
LGTM

PR #149, opened 2023-03-18
Twix copy text

okiwi6:
## Introduced Changes

Needed a way to quickly copy text from twix text fields, here it is.
On hovering a text in twix, a tooltip will appear. With a click the current text is put in the clipboard.

## How to Test
* Open twix
* Connect to a robot and subscribe any text output, eg. `Control.additional_outputs.positions`
* Hover over the text, a tooltip should appear
* Clicking will put the current text in the clipboard


knoellle (approved):
Nice feature, LGTM

PR #153, opened 2023-03-20
Make all paths in Behavior Simulator Twix overlay visible

PasGl:
## Introduced Changes

What it says

## ToDo / Known Issues

none

none

## How to Test

Behavior Simulator + Twix

### DIFF
--- a/tools/twix/src/panels/map/layers/behavior_simulator.rs
+++ b/tools/twix/src/panels/map/layers/behavior_simulator.rs
@@ -54,6 +54,9 @@ impl Layer for BehaviorSimulator {
                 continue
             };
 
+            let transparent_blue = Color32::from_rgba_unmultiplied(0, 0, 255, 150);
+            let transparent_light_blue = Color32::from_rgba_unmultiplied(173, 216, 230, 150);
+
             if let MotionCommand::Walk { path, .. } = motion_command {
                 for segment in path {
                     match segment {
@@ -62,7 +65,7 @@ impl Layer for BehaviorSimulator {
                             robot_to_field * line_segment.1,
                             Stroke {
                                 width: 0.025,
-                                color: Color32::from_rgba_unmultiplied(0, 0, 255, 63),
+                                color: transparent_blue,
                             },
                         ),
                         PathSegment::Arc(arc, orientation) => painter.arc(
@@ -70,7 +73,7 @@ impl Layer for BehaviorSimulator {
                             orientation,
                             Stroke {
                                 width: 0.025,
-                                color: Color32::from_rgba_unmultiplied(0xAD, 0xD8, 0xE6, 63),
+                                color: transparent_light_blue,
                             },
                             robot_to_field,
                         ),
--- a/tools/twix/src/panels/map/layers/behavior_simulator.rs
+++ b/tools/twix/src/panels/map/layers/behavior_simulator.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use color_eyre::Result;
 use eframe::epaint::{Color32, Stroke};
 use nalgebra::Isometry2;
-use types::FieldDimensions;
+use types::{FieldDimensions, MotionCommand, PathSegment};
 
 use crate::{
     nao::Nao, panels::map::layer::Layer, players_value_buffer::PlayersValueBuffer,
@@ -12,6 +12,7 @@ use crate::{
 
 pub struct BehaviorSimulator {
     robot_to_field: PlayersValueBuffer,
+    motion_command: PlayersValueBuffer,
 }
 
 impl Layer for BehaviorSimulator {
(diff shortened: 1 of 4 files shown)


tools/twix/src/panels/map/layers/behavior_simulator.rs
```diff
@@ -12,23 +12,33 @@ use crate::{
...
+    motion_command: PlayersValueBuffer,
 }
 
 impl Layer for BehaviorSimulator {
     const NAME: &'static str = "Behavior Simulator";
 
     fn new(nao: Arc<Nao>) -> Self {
         let robot_to_field = PlayersValueBuffer::try_new(
-            nao,
+            &nao,
```
knoellle, line 23: Arcs are cheap to clone. Please use them directly instead of a reference.
```suggestion
            nao,clone(),
```

tools/twix/src/panels/map/layers/behavior_simulator.rs
```diff
@@ -39,6 +49,34 @@ impl Layer for BehaviorSimulator {
...
+                                width: 0.025,
+                                color: Color32::from_rgba_unmultiplied(0, 0, 255, 63),
+                            },
+                        ),
+                        PathSegment::Arc(arc, orientation) => painter.arc(
+                            arc,
+                            orientation,
+                            Stroke {
+                                width: 0.025,
+                                color: Color32::from_rgba_unmultiplied(0xAD, 0xD8, 0xE6, 63),
```
knoellle, line 73: Why do you mix hexadecimal and decimal numbers for your colors? Please choose one and stick to it. I'd vote for decimals.

Also, from looking at the code, it is not obvious which color this is, consider defining a named constant and using that here instead.

tools/twix/src/panels/map/layers/behavior_simulator.rs
```diff
@@ -39,6 +49,34 @@ impl Layer for BehaviorSimulator {
...
+
+            if let MotionCommand::Walk { path, .. } = motion_command {
+                for segment in path {
+                    match segment {
+                        PathSegment::LineSegment(line_segment) => painter.line_segment(
+                            robot_to_field * line_segment.0,
+                            robot_to_field * line_segment.1,
+                            Stroke {
+                                width: 0.025,
+                                color: Color32::from_rgba_unmultiplied(0, 0, 255, 63),
```
knoellle, line 65: I think the colors chosen here or rather the low opacity make the lines quite faint.
Personally I would prefer more contrast between these lines and the green field in the background.
But I would still merge this if you are happy with the current state.

tools/twix/src/players_value_buffer.rs
```diff
@@ -10,7 +10,7 @@ use crate::{nao::Nao, value_buffer::ValueBuffer};
 pub struct PlayersValueBuffer(pub Players<ValueBuffer>);
 
 impl PlayersValueBuffer {
-    pub fn try_new(nao: Arc<Nao>, prefix: &str, output: &str) -> Result<Self> {
+    pub fn try_new(nao: &Arc<Nao>, prefix: &str, output: &str) -> Result<Self> {
```
knoellle, line 13: See above, please revert this change.

knoellle (changes requested):

tools/twix/src/panels/map/layers/behavior_simulator.rs
PasGl: done

tools/twix/src/players_value_buffer.rs
PasGl: done

tools/twix/src/panels/map/layers/behavior_simulator.rs
PasGl: Opacity has been increased.

tools/twix/src/panels/map/layers/behavior_simulator.rs
knoellle, line 1: One further thing to think about: In the map layers I actively decided for rendering the paths before/under the player positions.
This makes it easier to see the robots orientation at the cost of obscuring line paths that are shorter than the robots forward indicator.

I still think the way I described is better, what is your opinion?

PasGl: I agree.
I think I should also refactor the path-painting into a painter-method to reduce redundant code, now that there is more than one layer painting paths.

knoellle: Good point, please do.

PasGl: Both done.

tools/twix/src/panels/map/layers/behavior_simulator.rs
```diff
@@ -38,6 +51,21 @@ impl Layer for BehaviorSimulator {
                 width: 0.02,
                 color: Color32::BLACK,
             };
+
+            let Ok(motion_command): Result<MotionCommand> = self.motion_command.0[player_number].parse_latest() else {
+                continue
+            };
+
+            if let MotionCommand::Walk { path, .. } = motion_command {
```
knoellle, line 59: The let-else-continue here makes it harder to add more stuff later.
What do you think of this?
```suggestion
            if let Ok(MotionCommand::Walk { path, .. }) =
                self.motion_command.0[player_number].parse_latest()
```

PasGl: done

Issue #122, opened 2023-03-08
Convert Bounding Boxes to Robot Perceptions


schmidma:
The robot detection is planned to output a bounding box of detected robots in the image.
These bounding boxes have to be filtered and transformed to actual positions in robot coordinates. This includes merging of boxes and detection of robots feet to project them onto the ground

PR #156, opened 2023-03-22
Robot Detection using B-Human network

### LINKED ISSUE #126: Integrate the B-Human robot detection
For this issue, we plan to integrate the robot detection network trained by B-Human. This issue depends on #119.

schmidma:
## Introduced Changes

This PR introduces an additional neural network to detect robots in the image.

- The `LuminanceImageExtractor` was removed, this job is now executed by the `RobotDetection`.
- The former `RobotDetection` is now called `FeetDetection`
- new overlays are introduced to plot the robot detection and the feet detection

Fixes #126

## ToDo / Known Issues

none

Filtering of boxes etc #122

## How to Test

- enable either or both the robot detection and/or the feet detection in `etc/default.json` (can also be done at runtime)
- `./pepsi upload xx`
- investigate the robot detection and feet detection results in the image overlays in twix


crates/vision/src/robot_detection.rs
```diff
@@ -1,332 +1,236 @@
...
-                .iter()
-                .map(|point| point.amount_score * point.luminance_score)
-                .sum();
-            ScoredCluster {
-                center: median,
-                score,
-            }
-        })
-        .collect()
+fn map_to_zero_one(x: f32) -> f32 {
```
julianschuler, line 157: `map_to_zero_one` sounds IMO like a function with binary output. What do you think about `logistic` or `standard_logistic`?

crates/vision/src/robot_detection.rs
```diff
@@ -1,332 +1,236 @@
...
+        center: (center + grid_position)
+            .component_div(&grid_size)
+            .component_mul(&camera_image_size)
+            .into(),
+        size: (size * 10.0)
+            .component_mul(&scaling)
+            .component_mul(&camera_image_size)
+            .component_div(&grid_size),
+        probability,
+        distance: distance * 10.0,
```
julianschuler, line 234: Where does this factor (10) come from? Is it always the same for both cases? It may be a good idea to create a constant for this.

julianschuler (changes requested):
Very promising! I have only some minor remarks, feel free to implement or discuss.

crates/vision/src/robot_detection.rs
schmidma: haha, I actually do not remember that myself :D ... i have to look it up again. I should have given it a name in the first place... @jonathan-hellwig do you still remember what this was?

crates/vision/src/robot_detection.rs
schmidma: good point, i renamed it

crates/vision/src/robot_detection.rs
jonathan-hellwig: I think we scaled it by 10 because B-Human does so as well. I think it is just a conversion of units.

schmidma: done

julianschuler (approved):
LGTM!

PR #163, opened 2023-03-26
Rework feet detection

schmidma:
## Introduced Changes

Rewrites most of the feet detection, only focusing detection on the bottom camera.

## ToDo / Known Issues

none

none

## How to Test

- Deploy a robot or load the `robot_detection` world in webots
- Observe overlays in both map and image panel for the feet detection


etc/configuration/default.json
```diff
@@ -152,48 +152,30 @@
   },
   "robot_detection": {
     "vision_top": {
+      "enable": false,
```
julianschuler, line 155: Why do you disable the robot detection here? For testing the foot detection?

tools/twix/src/panels/map/layers/feet_detection.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        let cluster_points: Vec<ClusterPoint> = [&self.segments_bottom, &self.segments_top]
+            .iter()
+            .filter_map(|buffer| buffer.parse_latest::<Vec<_>>().ok())
+            .flatten()
+            .collect();
+        for point in cluster_points {
+            painter.circle_filled(
+                robot_to_field * point.position_in_ground,
+                0.02,
```
julianschuler, line 84: Maybe consider to use a constant to describe the radius.

julianschuler (changes requested):
LGTM! Only some minor remarks left, feel free to implement or discuss.

etc/configuration/default.json
schmidma: The robot detection is currently not integrated in the filter. This will be done in a future PR. At the moment, the robot detection is disabled, as it is not used and otherwise would just burn execution time ;)

tools/twix/src/panels/map/layers/feet_detection.rs
schmidma: I extracted the `0.02` to a variable

julianschuler: Would't a constant be more appropriate?

schmidma: I think the best would be to have it configurable, but that is definitely future work. This is debug tooling and a constant moves the information far away from this specific implementation. Additionally, I think introducing a constant mainly makes sense if it is used multiple times. I propose to keep it like this.

julianschuler (approved):

PR #168, opened 2023-03-29
Additional output for positions difference

### LINKED ISSUE #118: Tooling to measure difference in interpolation
Some sort of tooling (maybe target, measurement) to analyze interpolations efficiently

julianschuler:
## Introduced Changes

This PR adds an additional output describing the difference between the set and measured joint positions. It furthermore renames all occurrences of `rhs` to `right` to avoid abbreviations.

Resolves #118.

## ToDo / Known Issues

None.

Nothing here.

## How to Test

Flash a NAO and start twix. Plot any of the additional `positions_difference` output values and observe them when the robot is in motion.

### DIFF
--- a/crates/control/src/motion/joint_command_sender.rs
+++ b/crates/control/src/motion/joint_command_sender.rs
@@ -14,6 +14,7 @@ pub struct CreationContext {}
 #[context]
 pub struct CycleContext {
     pub positions: AdditionalOutput<Joints, "positions">,
+    pub positions_difference: AdditionalOutput<Joints, "positions_difference">,
     pub stiffnesses: AdditionalOutput<Joints, "stiffnesses">,
 
     pub center_head_position: Parameter<HeadJoints, "center_head_position">,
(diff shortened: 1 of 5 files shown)


crates/control/src/motion/joint_command_sender.rs
```diff
@@ -90,6 +91,9 @@ impl JointCommandSender {
 
         context.positions.fill_if_subscribed(|| positions);
         context.stiffnesses.fill_if_subscribed(|| stiffnesses);
+        context
+            .positions_difference
+            .fill_if_subscribed(|| current_positions - positions);
```
okiwi6, line 96: More intuitive would be a SHOULD - IS, not IS - SHOULD.

crates/control/src/motion/joint_command_sender.rs
```diff
@@ -15,6 +15,7 @@ pub struct CreationContext {}
 pub struct CycleContext {
     pub positions: AdditionalOutput<Joints, "positions">,
     pub stiffnesses: AdditionalOutput<Joints, "stiffnesses">,
+    pub positions_difference: AdditionalOutput<Joints, "positions_difference">,
```
okiwi6, line 18: Sort alphabetically?

okiwi6 (changes requested):
Nice addition!

crates/control/src/motion/joint_command_sender.rs
julianschuler: Thanks, fixed.

crates/control/src/motion/joint_command_sender.rs
julianschuler: Good point, fixed.

okiwi6 (approved):

PR #171, opened 2023-04-01
Kick decision

schmidma:
## Introduced Changes

This PR introduces:

- plotting of all computed kick decisions, and the best one
- a painter method to draw 'targets' (a circle with an x inside) in twix
- a parameter to enable side-kicking on the entire field
- tuning of parameters for the obstacle filter

based on #164 #163 

## ToDo / Known Issues

none

none

## How to Test

we already observed the effects during the test game of `2023-03-29`

The plotting can be tested using the map with the kick decision overlay


crates/control/src/behavior/dribble.rs
```diff
@@ -102,7 +105,7 @@ pub fn execute(
         kick_decisions
             .iter()
             .filter(|decision| {
-                !is_inside_any_obstacle(
+                is_inside_any_obstacle(
```
knoellle, line 108: This doesn't seem right to me. Why did you invert this?

etc/configuration/default.json
```diff
@@ -230,20 +230,20 @@
...
     "hypothesis_merge_distance": 0.5,
     "process_noise": [0.005, 0.005],
-    "robot_measurement_noise": [3.0, 5.0],
+    "robot_measurement_noise": [30.0, 30.0],
     "sonar_measurement_noise": [15.0, 15.0],
     "network_robot_measurement_noise": [3.0, 5.0],
     "initial_covariance": [0.5, 0.5],
     "measurement_count_threshold": 10,
-    "use_robot_detection_measurements": false,
+    "use_robot_detection_measurements": true,
```
knoellle, line 246: This parameter has a confusing name since it actually refers to the feet detection instead of the robot detection.
Maybe rename this and the related identifiers in the obstacle filter?

knoellle (changes requested):
Looks good overall but I have some minor remarks.

etc/configuration/default.json
schmidma: This will be fixed when the robot detection is integrated in the filter. See #191

crates/control/src/behavior/dribble.rs
schmidma: uh oh... good catch, i wanted to do expose all kick decision candidates (see https://github.com/HULKs/hulk/pull/171/commits/7426749b6fde8315ceb07e4cea636f04fde38083 ). Should be fixed now

knoellle (approved):
LGTM

PR #172, opened 2023-04-03
Projection crate

schmidma:
## Introduced Changes

This removes all camera matrix projection related implementations to an external workspace crate called `projection`.
Additionally, it changes all function signatures to take nalgebra types by value instead of reference. These types are `Copy`.

This is necessary refactoring, to work on projection related stuff. Upcoming robot detection implementation and filtering is adding additional functionality to check robot sizes etc.

## ToDo / Known Issues

none

move more implementation stuff out of `types`

## How to Test

compile the code, sanity checking, by e.g. looking at some projection related nodes (i.e. ball detection).
run tests (already done by the CI)


crates/projection/tests/projection.rs
```diff
@@ -0,0 +1,304 @@
...
+        optical_center,
+        image_size,
+        Isometry3::identity(),
+        Isometry3::identity(),
+        Isometry3::identity(),
+    )
+}
+
+#[test]
+fn pixel_to_camera_default_center() {
```
h3ndrk, line 22: Did you copy all tests? I would like to not check them manually if they only have been copied.

crates/types/src/horizon.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde::{Deserialize, Serialize};
+use serialize_hierarchy::SerializeHierarchy;
+
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize, SerializeHierarchy)]
+pub struct Horizon {
+    pub left_horizon_y: f32,
+    pub right_horizon_y: f32,
+}
+
+impl Horizon {
```
h3ndrk, line 11: Did you change the implementation despite copying?

crates/projection/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn robot_to_pixel(&self, robot_coordinates: Point3<f32>) -> Result<Point2<f32>, Error>;
+    fn get_pixel_radius(
+        &self,
+        radius_in_robot_coordinates: f32,
+        pixel_coordinates: Point2<f32>,
+        resolution: Vector2<u32>,
+    ) -> Result<f32, Error>;
+}
+
+impl Projection for CameraMatrix {
```
h3ndrk, line 39: Did you change the implementation despite copying?

crates/vision/src/line_detection.rs
```diff
@@ -112,10 +113,10 @@ impl LineDetection {
                 continue;
             }
             let (start_point_in_image, start_point_in_robot) =
-                match points_with_projection_onto_line.iter().find_map(
+                match points_with_projection_onto_line.iter().copied().find_map(
```
h3ndrk, line 116: You are now copying all the points vs. only the found point. Does that have an impact?

h3ndrk (dismissed):
Only some questions, haven't tested but the code looks fine if you only refactored and didn't change implementations.

knoellle (changes requested):
Looks good for the most part, just have one minor comment about variable naming.

crates/vision/src/line_detection.rs
knoellle: It still only copies until a value was found since the `Copied` iterator is still lazy.

h3ndrk: Yes, but it ranges from `1..n` copies, worst case is `n`.

knoellle: Sure, but the point type is only two `f32`s, which is hardly much more expensive than passing a reference.
The scaling behavior wasn't changed either, since neither copying nor taking a reference scales with the length of the vec.

I don't see why the change to copying here would be worse than copying anywhere else.

crates/types/src/horizon.rs
schmidma: no, the idea of this PR is to refactor first to ease the review. I did not touch any implementation. Implementation changes are coming in a next pr

crates/vision/src/line_detection.rs
h3ndrk: Ok, thumbs up from Schmiddy probably means that we want to keep it as implemented. Approval from my side, knoellle please approve if you are happy.

crates/control/src/ball_filter.rs
```diff
@@ -367,12 +369,12 @@ fn is_visible_to_camera(
...
     projected_limbs_bottom: &[Limb],
 ) -> bool {
-    let position_in_image = match camera_matrix
-        .ground_with_z_to_pixel(&Point2::from(hypothesis.filter.state().xy()), ball_radius)
-    {
-        Ok(position_in_image) => position_in_image,
-        Err(_) => return false,
-    };
+    let ground_coordinates = Point2::from(hypothesis.filter.state().xy());
+    let position_in_image =
```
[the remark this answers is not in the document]
schmidma: Ah, yes. I'm sorry. I changed that.
The naming comes from the function signature, as I used the rust-analyzer code action.
```rust
fn ground_with_z_to_pixel(&self, ground_coordinates: Point2<f32>, z: f32) -> Result<Point2<f32>, Error>
```
I renamed the variable to `position_on_ground` as you suggested.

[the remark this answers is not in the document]
knoellle: There is a similar change about 20 lines above. Please do the same there too.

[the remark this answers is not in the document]
schmidma: done, applied the same naming

knoellle (approved):
LGTM

PR #173, opened 2023-04-03
Upgrade GameController messages

### LINKED ISSUE #160: Update Game RoboCup Headers
New header files exists from the RoboCup-SPL League, they need to be added to our repo and compatibility needs to be checked.
RoboCup-SPL/GameController/examples/c

h3ndrk:
## Introduced Changes

`SplMessage` has been renamed to `GameControllerReturnMessage` and all usages have been adjusted. The added fields are also represented in the new messages.

Fixes #160

## ToDo / Known Issues

None

None

## How to Test

Test with a real GameController in a test game.


schmidma:
If I understand your changes correctly, we are using the `GameControllerReturnMessage` for the (now unspecified) team communication messages?
I propose to introduce another message type, e.g. `TeamMessage`

h3ndrk:
> If I understand your changes correctly, we are using the `GameControllerReturnMessage` for the (now unspecified) team communication messages?

Yes, the old `SplMessage` was very similar to the `GameControllerReturnMessage` and the TC therefore removed the `SplMessage` type. They are only distinguished by the port they are sent from/to. The robots are now exchanging return messages. I wanted to align our naming of the types with how they are named officially. Yes, this naming is not optimal but we are following it currently.

> I propose to introduce another message type, e.g. `TeamMessage`

It is the question if we want to use different naming in our code base.

knoellle:
> Yes, the old `SplMessage` was very similar to the `GameControllerReturnMessage` and the TC therefore removed the `SplMessage` type.

Is that true though? In my understanding of the rules, the standardization of inter-robot communication has been dropped completely. From rule diff to 2022, chapter 2.4.2 "Wireless":
image

Sure, we can just use the same fields that the return message has for our own messages, but I agree with @schmidma that we should introduce a separate type for this.

h3ndrk:
> Is that true though? In my understanding of the rules, the standardization of inter-robot communication has been dropped completely.

Correct, I misunderstood this. I thought we still need to send `GameControllerReturnMessages` but are free to use the remaining bytes. Thanks for clarification.

> but I agree with @schmidma that we should introduce a separate type for this.

So, duplicate the type with all serialization code?



Or introduce an alias? What is the benefit of that?

Or do we want to send different information and serialize differently? IMHO, changing the message is out-of-scope for upgrading the messages and propagating the name through our code from the header files.

schmidma:
IMHO changing the message itself is out of scope of this PR. I only propose to introduce a different type/symbol for the messages for team communication and the messages for game controller communication. This way, future changes to the team communication, which we are free to implement in any way, is easy. Currently you are merging both use-cases by introducing the same symbol/type.
A type alias would be totally sufficient as far as i can see. This way, a independent type for team communication can be introduced in the future without major changes in its usage.

knoellle (approved):
LGTM

PR #176, opened 2023-04-04
Show whistle detection status via ear LEDs

knoellle:
## Introduced Changes

Ear LEDs are currently unused. With this PR, they will be used to indicate whether the robots detected a whistle or not.

Mainly implemented for the NAO Devils image, but we might as well use it ourselves too.

## How to Test

- Upload
- Blow whistle
- Inspect photons emerging from the "ear" area of the robot. You should seen an increase of those commonly associated with the color blue.

### DIFF
--- a/crates/control/src/led_status.rs
+++ b/crates/control/src/led_status.rs
@@ -3,7 +3,7 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use color_eyre::Result;
 use context_attribute::context;
 use framework::{MainOutput, PerceptionInput};
-use types::{Ball, CycleTime, Eye, Leds, PrimaryState, Rgb};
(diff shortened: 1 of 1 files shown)


crates/control/src/led_status.rs
```diff
@@ -147,9 +148,17 @@ impl LedStatus {
             last_ball_data_top_too_old,
             last_ball_data_bottom_too_old,
         );
+
+        let ears = if context.filtered_whistle.is_detected {
```
h3ndrk, line 152: How long is `is_detected` true? If only one cycle, is this noticeable?

h3ndrk (approved):

crates/control/src/led_status.rs
knoellle: When testing this the ears just flashed for a couple cycles for the shortest whistle burst I could create but were lit up continuously for blowing the whistle for longer.

PR #177, opened 2023-04-05
Webotsfalling

MaikRe:
## Introduced Changes

Enables automatic testing for the NAO being pushed and falling with easy repeatability.
Based on work from @PasGl for starting the controller automatically.

Fixes #

Simulated pushing during walking, i.e. not while standing still to test falling parameters in motion.

## How to Test

Recompile with pepsi build, launch the webots/worlds/falling.wbt file. The controller is started automatically with simulated chest button presses and shortly after the robot is knocked over. Pressing the circular refresh button or Ctrl+Shift+R resets the simulation and allows for test in succession with automatically reloaded scene and config files.


webots/controllers/falling/falling.py
```diff
@@ -0,0 +1,40 @@
+from controller import Supervisor
+from simple_websocket_server import WebSocketServer, WebSocket
```
knoellle, line 2: This package is not part of the standard library nor is it provided by webots automagically.
Please add it to the `requirements.txt` file.

webots/controllers/falling/falling.py
```diff
@@ -0,0 +1,40 @@
...
+    scene_control_server = WebSocketServer("localhost", 9980, SceneControl)
+    scene_control_server.serve_forever()
+
+websocket_thread = Thread(target=run_scene_control_server)
+websocket_thread.start()
+
+count = 0
+pressed = 0
+
+nao_node=supervisor.getFromDef("NAO")
```
knoellle, line 27: Please format this file

crates/hulk/src/webots/keyboard_device.rs
```diff
@@ -1,28 +1,38 @@
...
+                None
+            }
+        };
+
         let control_shift_c_pressed = if let Some(key) = key {
             key == Keyboard::CONTROL | Keyboard::SHIFT | 'C' as u32
         } else {
-            false
+            received_message.is_some()
         };
```
knoellle, line 36: Doesn't this logic drop messages when any key is pressed?

crates/hulk/src/webots/keyboard_device.rs
```diff
@@ -1,28 +1,38 @@
...
         let key = self.keyboard.get_key();
+        let received_message = match self.receiver.get_next_packet() {
+            Ok(message) => message,
+            Err(error) => {
+                println!("error in received message: {error:?}");
+                None
+            }
+        };
+
         let control_shift_c_pressed = if let Some(key) = key {
```
knoellle, line 32: This name is no longer accurate. Rename to `chest_button(_pressed)`?

crates/control/Cargo.toml
```diff
@@ -20,6 +21,7 @@ projection = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serialize_hierarchy = { workspace = true }
+simple-websockets = { workspace = true }
```
knoellle, line 24: Are these two dependencies added here used at all?
I don't see them being used anywhere.

Cargo.toml
```diff
@@ -112,6 +112,7 @@ serde_test = "1.0.152"
 serde-transcode = "1.1.1"
 serialize_hierarchy = { path = "crates/serialize_hierarchy" }
 serialize_hierarchy_derive = { path = "crates/serialize_hierarchy_derive" }
+simple-websockets = "0.1.4"
```
knoellle, line 115: I think this is unused

knoellle (changes requested):
Works as advertised but there are some things left to clean up. 😉

webots/controllers/falling/falling.py
MaikRe: Removed unnecessary imports and unused references

webots/controllers/falling/falling.py
MaikRe: Done

crates/control/Cargo.toml
MaikRe: removed

Cargo.toml
MaikRe: removed

crates/hulk/src/webots/keyboard_device.rs
MaikRe: Technically yes, but irrelevant because if the keys are manually pressed then we can ignore presses caused by received messages

knoellle: You say "the keys", but we don't go into the else case even if the user is holding down some completely different key.

crates/control/Cargo.toml
knoellle: There's still the byteorder dependency.
I don't think that's used either.

MaikRe: removed byteorder dependency from control crate

crates/hulk/src/webots/keyboard_device.rs
```diff
@@ -29,10 +29,8 @@ impl KeyboardDevice {
             }
         };
 
-        let chest_button_pressed = if let Some(key) = key {
-            key == Keyboard::CONTROL | Keyboard::SHIFT | 'C' as u32
-        } else {
-            received_message.is_some()
+        let chest_button_pressed = received_message.is_some() | {
+            key == Some(Keyboard::CONTROL | Keyboard::SHIFT | 'C' as u32)
         };
```
knoellle, line 34: `|` is the bitwise or operation.
In the line below it is used to combine the modifier values with the character to get a keycode, are you sure you want to use that here to combine booleans?

Also, why are there curly braces here?

MaikRe: Both mistakes are linked, I didn't use the proper || operator hence rust analyzer complained and wanted curly braces, I didn't think twice as to why, is fixed now

knoellle (approved):
LGTM

Issue #71, opened 2023-01-29
Ideas for next iteration


h3ndrk:
Things that @h3ndrk has on his list but may not have time for them. Feel free to remove them here and create individual issues.

- Visual semantic segmentation
- Opponent robot orientation estimation
- Datatype/module recording and replay
- ~~Extract communication server into crate and integrate into behavior simulator (allow Flora to connect)~~
- Integrate logging into framework
- ~~Support getting the next data item without subscribing (registers temporary subscriber and waits for single data item, then unsubscribes automatically)~~
- Deduplicate configuration for vision cyclers (`$this_cycler` should be optional and also usable in the middle of the path)
- Audio playback
- Fix change notification for children configuration parameters
- Automatically generate databases and configuration from module attributes
- Statically check validity/completeness of to be uploaded configuration files before uploading/at compile time
- ~~Reduce code duplication in communication server~~
- Integrate aliveness message receival into the `nao` crate #27 
- Parallelize Pepsi for multiple robots (dependency graph) #46 
- Readd remote control support via WiFi and smartphone browser (needs to be deployed on all NAOs, NAOs need to open a WiFi network when e.g. pressing a button)
- Write more documentation
- Custom clippy checks for formatting and other rules that the HULKs want to enforce in the repository
- Replace Yocto with Nix

PR #182, opened 2023-04-09
Re-Framework

schmidma:
## Introduced Changes

This refactors the existing framework parts of source analyzation and code generation.

  - introduction of `framework.toml` (`crates/hulk/framework.toml`)
  - rewrite of `crates/source_analyzer`
    - central `Cyclers` struct containing all info in a single tree
    - no bidirectional mappings
    - no code generation related information in source analyzation step
    - error handling using `thiserror` including span information
    - circular dependencies are identified and lead to errors
  - refactor code generation
    - improve readability
    - use `quote!{ ... }`
    - error handling using `thiserror`

## ToDo / Known Issues

- recheck the `rerun-if-changed` calls, it seems the binary is always recompiled
- selection of different `framework.toml` configurations to allow for different run-configurations.
  - reintegration of the behavior simulator
- Check allows

- #21
- #71
- Errors have to be improved, when hitting them. I tried to test many of them, but definitely missed corner cases. Additionally, they are not shown in the rust analyzer diagnostics. I did not find a way to do that yet.
- parameter as value not as reference by default
- default should not be required for parameters

## How to Test

`./pepsi run` (for webots)
`./pepsi upload xx` (to test the nao)
`./pepsi analyze xxxxxx` to query framework information


crates/source_analyzer/src/struct_hierarchy.rs
```diff
@@ -0,0 +1,113 @@
...
+    },
+}
+
+impl StructHierarchy {
+    pub fn new_struct() -> Self {
+        Self::Struct {
+            fields: Default::default(),
+        }
+    }
+}
```
h3ndrk, line 25: Implement `Default` instead?

crates/source_analyzer/src/structs.rs
```diff
@@ -99,190 +77,86 @@ impl Structs {
...
-    BeginStruct,
-    AppendDataType { data_type: Type },
-}
-
-fn path_to_insertion_rules(path: &[PathSegment], data_type: &Type) -> Vec<InsertionRule> {
-    path.iter()
+fn path_to_insertion_rules<'a>(
+    path: &'a Path,
+    data_type: &Type,
+) -> impl 'a + Iterator<Item = InsertionRule> {
```
h3ndrk, line 157:
```suggestion
fn path_to_insertion_rules<'path>(
    path: &'path Path,
    data_type: &Type,
) -> impl 'path + Iterator<Item = InsertionRule> {
```

crates/source_analyzer/src/structs.rs
```diff
@@ -304,30 +178,29 @@ fn path_to_insertion_rules(path: &[PathSegment], data_type: &Type) -> Vec<Insert
         .chain(once(InsertionRule::AppendDataType {
             data_type: data_type.clone(),
         }))
-        .collect()
 }
 
-fn unwrap_option_data_type(data_type: Type) -> Result<Type> {
+fn unwrap_option_type(data_type: Type) -> Type {
```
h3ndrk, line 183: Why do you now panic instead of returning errors?

crates/source_analyzer/src/uses.rs
```diff
@@ -1,11 +1,11 @@
...
 
-use syn::{Ident, Item, UseTree};
+use syn::{File, Ident, Item, UseTree};
 
 pub type Uses = HashMap<Ident, Vec<Ident>>;
 
-pub fn uses_from_items(items: &[Item]) -> Uses {
-    items
+pub fn uses_from_file(file: &File) -> Uses {
+    file.items
```
h3ndrk, line 8: What was the reason to make this less generic?

crates/types/src/camera_position.rs
```diff
@@ -1,5 +1,11 @@
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+use serde::{Deserialize, Serialize};
+use serialize_hierarchy::SerializeHierarchy;
+
+#[derive(
+    Default, Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize, SerializeHierarchy,
+)]
 pub enum CameraPosition {
+    #[default]
```
h3ndrk, line 8: Is this required?

crates/vision/src/camera_matrix_extractor.rs
```diff
@@ -13,8 +11,9 @@ pub struct CreationContext {}
 #[context]
 pub struct CycleContext {
     pub camera_matrices: RequiredInput<Option<CameraMatrices>, "Control", "camera_matrices?">,
+    pub camera_position:
+        Parameter<CameraPosition, "image_receiver.$cycler_instance.camera_position">,
     pub image: Input<YCbCr422Image, "image">, // required for correct node order
```
h3ndrk, line 16: Remove `image` input because you have now the concept of setup nodes?

tests/behavior/golden_goal.lua
h3ndrk, line 1: Revert changes in this file because only formatting?

tools/behavior_simulator/src/cycler.rs
h3ndrk, line 1: @knoellle Please review the behavior simulator changes. Resolve if read.

tools/pepsi/src/analyze.rs
```diff
@@ -1,103 +1,50 @@
...
-use source_analyzer::{parse_rust_file, Contexts, CyclerInstances, CyclerTypes, Nodes, Structs};
+use source_analyzer::{
+    contexts::Contexts, cycler::Cyclers, manifest::FrameworkManifest, node::parse_rust_file,
+    structs::Structs,
+};
 
 #[derive(Subcommand)]
 #[allow(clippy::enum_variant_names)]
 pub enum Arguments {
-    DumpBuildScriptOutput {
```
h3ndrk, line 13: Why did you remove this command?

tools/pepsi/src/analyze.rs
```diff
@@ -1,103 +1,50 @@
...
-        /// File name to dump (may contain wildcard characters usable by glob())
-        file_name: String,
-    },
     DumpContexts {
         /// File path to a Rust file containing a module with context structs
         file_path: PathBuf,
     },
-    DumpCyclerInstances,
-    DumpCyclerTypes,
-    DumpLatest {
```
h3ndrk, line 25: Why did you remove this command?

crates/framework/Cargo.toml
```diff
@@ -7,11 +7,9 @@ homepage = "https://github.com/hulks/hulk"
 
 [dependencies]
 parking_lot = { workspace = true }
-structs = { workspace = true }
 types = { workspace = true }
 
 [build-dependencies]
```
h3ndrk, line 12: Remove `[build-dependencies]` because there is no `build.rs` anymore?

crates/control/src/behavior_simulator_setup.rs
h3ndrk, line 1: Is this file dead code?

crates/control/src/ball_filter.rs
```diff
@@ -45,7 +45,6 @@ pub struct CycleContext {
     pub current_odometry_to_last_odometry:
         HistoricInput<Option<Isometry2<f32>>, "current_odometry_to_last_odometry?">,
     pub historic_camera_matrices: HistoricInput<Option<CameraMatrices>, "camera_matrices?">,
-    pub projected_limbs: HistoricInput<Option<ProjectedLimbs>, "projected_limbs?">,
```
h3ndrk, line 48: What did you do here? It is probably the easiest if you shortly explain instead of me digging deep into the details and trying to understand.

crates/code_generation/src/structs.rs
```diff
@@ -0,0 +1,119 @@
...
+        #configuration
+        #(#cyclers)*
+    }
+}
+
+fn hierarchy_to_token_stream(
+    hierarchy: &StructHierarchy,
+    struct_name: Ident,
+    derives: &TokenStream,
+) -> TokenStream {
```
h3ndrk, line 62: Why does this panic now?

crates/code_generation/src/run.rs
```diff
@@ -3,210 +3,43 @@ use std::iter::repeat;
...
-                })
-                .collect::<Vec<_>>()
-        })
-        .collect();
     quote! {
-        #[allow(unused_imports, unused_variables)]
-        pub fn run<Interface>(
-            hardware_interface: std::sync::Arc<Interface>,
+        #[allow(dead_code, unused_mut, unused_imports, unused_variables, clippy::too_many_arguments, clippy::redundant_clone, clippy::needless_question_mark)]
+        pub(crate) fn run(
```
h3ndrk, line 19: Should this even be public?

crates/code_generation/src/accessor.rs
```diff
@@ -1,383 +1,310 @@
...
-            prefix_token_stream,
-            path,
-            reference_type,
-            None,
-        )
+        path_to_accessor_token_stream_with_cycler_instance(prefix, path, reference_type, None)
     }
 }
 
 fn path_to_accessor_token_stream_with_cycler_instance(
```
h3ndrk, line 52: This is a one of the success stories where a high code coverage leads to less effort in the review. :100:

Resolve if read.

crates/hulk_nao/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+serde = { workspace = true }
+serde_json = { workspace = true }
+serialize_hierarchy = { workspace = true }
+spl_network = { workspace = true }
+spl_network_messages = { workspace = true }
+tokio = { workspace = true }
+tokio-util = { workspace = true }
+types = { workspace = true }
+v4l = { workspace = true }
+vision = { workspace = true }
```
h3ndrk, line 36: Why do we have more dependencies than the original `hulk` crate's `Cargo.toml`?

crates/hulk_webots/src/hardware_interface.rs
h3ndrk, line 1: Did you test that Webots still works with the new changes?

crates/hulk_webots/src/hardware_interface.rs
```diff
@@ -7,26 +7,31 @@ use std::{
...
-    hardware::{self, Ids},
+    hardware::Ids,
     messages::{IncomingMessage, OutgoingMessage},
     samples::Samples,
     ycbcr422_image::YCbCr422Image,
     CameraPosition, Joints, Leds, SensorData,
 };
 use webots::Robot;
 
+use spl_network::endpoint::{Endpoint, Ports};
```
h3ndrk, line 33: Why is this separate from the other uses?

crates/hulk_webots/src/hardware_interface.rs
```diff
@@ -7,26 +7,31 @@ use std::{
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
+use ::hardware::{
```
h3ndrk, line 10: Do you need a colon prefix?

h3ndrk (changes requested):

knoellle (changes requested):

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+pub type CyclerName = String;
+pub type InstanceName = String;
+pub type ModulePath = String;
+
+#[derive(Debug)]
+pub struct Instance {
+    pub name: InstanceName,
+}
+
+impl Display for Instance {
```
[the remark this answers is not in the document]
h3ndrk: Understood, maybe remove the `Display` stuff because it is a lot of visual noise and only used rarely and we might ditch the Pepsi dump commands in the future?

tests/behavior/golden_goal.lua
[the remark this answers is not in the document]
h3ndrk: Duplicated comment?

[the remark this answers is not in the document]
knoellle: Ah, yes. I had created this comment on saturday but only submitted the  review today. Resolving.

Cargo.toml
```diff
@@ -41,7 +40,6 @@ alsa = "0.7.0"
 aliveness = { path = "crates/aliveness" }
 approx = "0.5.1"
 audio = { path = "crates/audio" }
-behavior_simulator = { path = "tools/behavior_simulator" }
```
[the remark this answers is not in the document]
schmidma: It is a binary crate. There is no need to have it in the dependencies?! Nobody can depend on it

crates/code_generation/src/accessor.rs
```diff
@@ -1,383 +1,310 @@
...
-        if segment.is_variable && segment.name != "cycler_instance" {
-            unimplemented!("only $cycler_instance is implemented");
-        }
+    if let Some(segment) = path
+        .segments
+        .iter()
+        .find(|segment| segment.is_variable && segment.name != "cycler_instance")
+    {
+        let name = &segment.name;
+        unimplemented!("unexpected `${name}` only $cycler_instance can be used as variable in path",)
```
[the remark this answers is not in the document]
schmidma: I think this is coding style an neither is clearly (and objectively) better than the other. I like it the way I implemented it, as the output message is readable in a single sentence and I do not have to parse and understand the pattern matching when reading the message in the code.
I suggest that you grant me this room of personal coding style as long as the only reason is fewer lines of code

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
schmidma: I grasp two different aspects in this discussion:

1. Having a `type Instance = String` instead of a `struct String`
This might be an artifact of previous iterations (I agree). While writing the code I understood the concept of an instance as more than just its name. In the end it boiled down to having just one property (`name`). Being strict this might additionally be YAGNI as an `Instance` might have additional properties. The reason to already use two distinct types for the instance itself and the name is that in the source analysis and code generation these concepts are treated differently.

2. Removing Display
I agree that this is only for debug purposes using `pepsi dump[...]` I am going to move the implementation to `pepsi` but vote to not delete it as the information about sorted nodes is otherwise not available to the user

knoellle:
Tested recompilation times with the code generation extraction from [4e7000e](https://github.com/HULKs/hulk/pull/182/commits/4e7000e4eabeffbde142ce8081a9bed22f3a3ad3).

# Hot builds

Recompilation time when touching `walking_engine.rs` without changing the code.

```sh
pepsi build # --profile release
touch crates/control/src/motion/walking_engine.rs
pepsi build # --profile release
```

|           | now  | previous |
|-----------|------|----------|
|release    | 18s  | 15.5s    |
|incremental| 5.5s | 15.5s    |

# Cold builds

Using a locally installed pepsi to unfair reuse of crates already built for pepsi in incremental mode.

```sh
cargo clean
pepsi build # --profile release
```

|           | now  | previous |
|-----------|------|----------|
|release    | 42s  | 49s      |
|incremental| 45s  | 52.5s    |

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+            kind: cycler_manifest.kind,
+            instances,
+            setup_nodes,
+            cycle_nodes,
+        })
+    }
+
+    pub fn sort_nodes(&mut self) -> Result<(), Error> {
+        let output_to_setup_node: HashMap<_, _> = self
+            .setup_nodes
```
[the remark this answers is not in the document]
schmidma: It is always hard to answer this kind of questions. In my opinion `setup` is the better name. This is why I renamed it. Which kind of discussion did you want to have on this topic? Should we involve more people and vote for naming? How do you want to proceed?

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+            name: cycler_manifest.name.clone(),
+            kind: cycler_manifest.kind,
+            instances,
+            setup_nodes,
+            cycle_nodes,
+        })
+    }
+
+    pub fn sort_nodes(&mut self) -> Result<(), Error> {
+        let output_to_setup_node: HashMap<_, _> = self
```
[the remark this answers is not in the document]
schmidma: renamed to `output_name_to_setup_node`

tools/pepsi/src/analyze.rs
```diff
@@ -51,3 +59,100 @@ pub async fn analyze(arguments: Arguments) -> Result<()> {
...
+
+trait ToWriterPretty {
+    fn to_writer_pretty(&self, f: &mut impl Write) -> fmt::Result;
+}
+
+impl ToWriterPretty for Instance {
+    fn to_writer_pretty(&self, f: &mut impl Write) -> fmt::Result {
+        write!(f, "{}", self.name)
+    }
+}
```
h3ndrk, line 77: `f`?! Please don't abbreviate. Also below.

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
h3ndrk: 1. Vote for `type Instance = String`
2. Ok, yes, moving it near the usage is also good.

Looks good now, resolving.

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
h3ndrk: Ok, I have no hard feelings between "setup" and "prepare". Then let's go with "setup". Then second question: Did you alter all existing names? Damals, we had some commments with "prepare". Maybe there are still left-overs?

crates/source_analyzer/src/path.rs
```diff
@@ -37,7 +38,7 @@ impl Path {
                     .map(|segment| {
                         if segment.is_variable {
                             PathSegment {
-                                name: instance.name.to_case(Case::Snake),
+                                name: instance.name.clone(),
```
h3ndrk: Why did you change the case style?

crates/hulk/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#![recursion_limit = "256"]
```
h3ndrk, line 1: Is this needed?

h3ndrk (changes requested):

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+                    .instances
+                    .iter()
+                    .map(move |instance| (cycler, instance))
+            })
+    }
+}
+
+fn sort_nodes(
+    nodes: &[Node],
+    output_to_node: &HashMap<String, &Node>,
```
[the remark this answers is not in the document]
h3ndrk: Implemented

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
h3ndrk: Implemented

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
h3ndrk: Checked, done

crates/source_analyzer/src/node.rs
```diff
@@ -0,0 +1,88 @@
...
+                Item::Impl(implementation) if has_new_and_cycle_method(implementation) => {
+                    match *implementation.self_ty {
+                        Type::Path(ref path) => path.path.get_ident(),
+                        _ => None,
+                    }
+                }
+                _ => None,
+            })
+            .ok_or_else(|| wrap_error(ParseError::new_spanned(&rust_file, "cannot find node declaration, expected a type with new(...) and cycle(...) method")))?
+            .to_string();
```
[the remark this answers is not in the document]
h3ndrk: Readded and made better than in the `main`.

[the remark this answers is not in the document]
h3ndrk: This is not relevant anymore because the topological sort has now nodes as items and not raw strings (node names). Collisions should not happen anymore.

crates/source_analyzer/src/contexts.rs
```diff
@@ -75,29 +81,26 @@ impl Contexts {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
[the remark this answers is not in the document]
h3ndrk: Sorted all of the derives.

crates/source_analyzer/src/structs.rs
```diff
@@ -304,30 +178,29 @@ fn path_to_insertion_rules(path: &[PathSegment], data_type: &Type) -> Vec<Insert
...
                             "unexpected generic argument, expected type argument in data type"
                         ),
                     }
                 }
-                _ => bail!("expected exactly one generic type argument in data type"),
+                _ => panic!("expected exactly one generic type argument in data type"),
             }
         }
-        _ => bail!("execpted Option<T> as data type"),
+        _ => panic!("execpted Option<T> as data type"),
```
knoellle, line 203: Typo

knoellle:
Rebased and resolved conflicts.
Had to downgrade the `toml` crate to make it compile because of https://github.com/toml-rs/toml/issues/490.
We should undo this when removing toml parsing.

crates/hulk_webots/src/hardware_interface.rs
knoellle: I just tested it, seems to work fine. The robot can walk, see the ball, and kick it.

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+pub struct Cycler {
+    pub name: CyclerName,
+    pub kind: CyclerKind,
+    pub instances: Vec<Instance>,
+    pub setup_nodes: Vec<Node>,
+    pub cycle_nodes: Vec<Node>,
+}
+
+impl Cycler {
+    fn try_from_manifest(cycler_manifest: &CyclerManifest, root: &Path) -> Result<Cycler, Error> {
```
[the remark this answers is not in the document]
knoellle: Doesn't remove all clones in the file but still seems better, done.

crates/hulk_webots/src/hardware_interface.rs
knoellle: Seems to have been removed already.

crates/hulk_webots/src/hardware_interface.rs
knoellle: Presumably because it is from a local crate.

crates/hulk/src/lib.rs
knoellle: Yes, compilation fails with very ugly error messages otherwise.

crates/source_analyzer/src/cycler.rs
```diff
@@ -15,21 +15,14 @@ use crate::{
     node::Node,
 };
 
-pub type CyclerName = String;
-pub type InstanceName = String;
-pub type ModulePath = String;
```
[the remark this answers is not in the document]
knoellle: The module path is now stored as a `syn::Path` directly.
If we want to add a type alias for it, it should probably live somewhere else though. Maybe next to the `Node` type?

Cargo.toml
```diff
@@ -110,34 +109,34 @@ regex = "1.6.0"
...
 syn = { version = "1.0.101", features = ["full", "extra-traits"] }
 tempfile = "3.3.0"
 thiserror = "1.0.37"
+threadbound = "0.1.6"
 tokio = { version = "1.21.2", features = ["full"] }
-surge-ping = "0.8.0"
 tokio-tungstenite = "0.19.0"
 tokio-util = "0.7.4"
-toml = "0.7.4"
+toml = "0.5.8"
```
h3ndrk, line 132: Do we still need TOML?

crates/code_generation/src/perception_databases.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut timestamp_array_items = cyclers
+        .instances_with(CyclerKind::Perception)
+        .map(|(_cycler, instance)| {
+            let field_name = &instance.name;
+            quote! {
+                self.#field_name.first_timestamp_of_non_finalized_database
+            }
+        })
+        .peekable();
+    let find_min_timestamp = if timestamp_array_items.peek().is_some() {
```
h3ndrk, line 24: Since there is a difference between minimum and minimal we should not abbreviate:

```suggestion
    let find_minimum_timestamp = if timestamp_array_items.peek().is_some() {
```

crates/hulk/framework.toml
h3ndrk, line 1: Why does this file still exist when we already have removed TOML parsing?

crates/hulk/src/lib.rs
h3ndrk: Can we somehow fix the root cause instead of silencing the symptom?

crates/hulk_webots/src/hardware_interface.rs
h3ndrk: From this crates view it is a third-party crate so we should put it into the appropriate group.

crates/source_analyzer/src/cyclers.rs
```diff
@@ -0,0 +1,225 @@
...
+};
+
+pub type CyclerName = Ident;
+pub type InstanceName = Ident;
+pub type OutputName = Ident;
+
+#[derive(Debug)]
+pub struct Instance {
+    pub name: InstanceName,
+}
```
h3ndrk, line 25: Remove struct as discussed in a previous discussion.

h3ndrk (changes requested):

crates/hulk_webots/src/hardware_interface.rs
knoellle: What would be a first party crate then?
I like to separate uses from registry crates from those from crates belonging to our project.

crates/source_analyzer/src/path.rs
knoellle: Undone

crates/hulk/src/lib.rs
knoellle: The error occurs when the compiler tries to parse the serialize_hierarchy macro output on the generated main outputs type of the control crate.
I don't think we should optimize for this recursion limit.

crates/code_generation/src/run.rs
knoellle: It has to be, since the run function is what is being called in the binary main functions.

crates/control/src/ball_filter.rs
knoellle: The limb projector was moved from Control to Vision. I think both are fine and I have no strong opinion regarding this.

crates/hulk/build.rs
```diff
@@ -12,7 +16,7 @@ fn main() -> Result<()> {
         cyclers: vec![
             CyclerManifest {
                 name: "Vision",
-                kind: CyclerKind::Perception,
+                kind: Perception,
```
h3ndrk: Do we really remove the enum prefix? For me it is more natural to have the enum type as prefix. Rare exceptions are some of Rust's prelude enum variants.

tools/pepsi/Cargo.toml
```diff
@@ -14,6 +14,7 @@ color-eyre = { workspace = true }
 constants = { workspace = true }
 futures-util = { workspace = true }
 indicatif = { workspace = true }
+itertools = { workspace = true }
```
h3ndrk, line 17: Is this needed? If not: remove?

h3ndrk (changes requested):

crates/source_analyzer/src/contexts.rs
```diff
@@ -1,69 +1,75 @@
...
+            writeln!(f, "  {field}")?;
+        }
+        Ok(())
+    }
+}
+
 impl Contexts {
-    pub fn try_from_file(file_path: impl AsRef<Path>, file: &File) -> Result<Self> {
-        let uses = uses_from_items(&file.items);
+    pub fn try_from_file(file: &File) -> Result<Self, ParseError> {
```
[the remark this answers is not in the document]
h3ndrk: Do we need better error messages?

crates/source_analyzer/src/cycler.rs
```diff
@@ -0,0 +1,255 @@
...
+        let instance_names = cycler_manifest
+            .instances
+            .clone()
+            .unwrap_or_else(|| vec![String::new()]);
+        let instances = instance_names
+            .iter()
+            .map(|instance_name| Instance {
+                name: format!("{}{}", cycler_manifest.name, instance_name),
+            })
+            .collect();
```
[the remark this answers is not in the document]
h3ndrk: Doesn't work easily. Skip.

crates/source_analyzer/src/contexts.rs
```diff
@@ -433,571 +350,16 @@ fn extract_three_arguments(
...
-            arguments.span(),
+        _ => Err(ParseError::new_spanned(
+            arguments,
             "expected exactly three generic parameters",
-            file_path,
-        ),
-    }
-}
-
-#[cfg(test)]
```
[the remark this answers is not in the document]
h3ndrk: They are gone and should be added if there is no reason to remove them. Maybe they need to be written more cleanly. But messy tests are at least better than removed tests.

crates/hulk_webots/src/hardware_interface.rs
h3ndrk: That is a new policy. Until now we formatted completely crate-local. Crates that are not the current crate are considered third-party. Since rustfmt and rust-analyzer handle crates this way, we should not define our own rule here. Please group them together.

tools/behavior_simulator/src/cycler.rs
```diff
@@ -10,81 +17,82 @@ use control::{
...
-    ball_state_composer: BallStateComposer,
     active_vision: ActiveVision,
-    kick_selector: KickSelector,
-    world_state_composer: WorldStateComposer,
+    ball_state_composer: BallStateComposer,
     behavior: Behavior,
+    kick_selector: KickSelector,
+    role_assignment: RoleAssignment,
     rule_obstacle_composer: RuleObstacleComposer,
+    world_state_composer: WorldStateComposer,
```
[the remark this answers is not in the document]
h3ndrk: Alphabetical sorting allows to see which nodes exist or are missing.

crates/control/src/hulk_message_filter.rs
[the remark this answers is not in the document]
h3ndrk: Since this is a framework refactor, refactorings in the robotics domain are out-of-scope IMHO. Feel free to revert. We can introduce this again (also in a refined way) in a subsequent PR.

crates/source_analyzer/src/cycler.rs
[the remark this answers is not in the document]
h3ndrk: Let's leave it out. Resolving.

crates/source_analyzer/src/manifest.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use serde::{
+    de::{value::MapAccessDeserializer, MapAccess, Visitor},
+    Deserialize, Deserializer,
+};
+
+use crate::{cycler::CyclerKind, error::Error};
+
+#[derive(Deserialize, Debug)]
+pub struct FrameworkManifest {
```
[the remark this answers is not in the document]
knoellle: Keep in mind that there are other manifests too. Having a `FrameworkManifest` that contains `CyclerManifest`s seems more sensible to me than having a generic `Manifest` that contains `CyclerManifest`s.

crates/hulk_webots/src/hardware_interface.rs
knoellle: AFAIK rustfmt and rust-analyzer don't handle import group order at all (by default).
rustfmt sorts within groups and I know there is an option to do some grouping, but its off by default.

Nevertheless I think it makes sense to split off imports belonging to our project from those belonging to third-party crates.
But that's something for after robocup, merged the imports for now for now.

crates/source_analyzer/src/contexts.rs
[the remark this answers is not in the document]
knoellle: I played around with this a little bit and error messages seem perfectly fine (and much better than on main).
As we discussed previously, the current implementation focuses on generating good error messages for framework users at the cost of usefulness (through backtraces, etc.) for framework developers.
I think it's fine like this.

crates/source_analyzer/src/manifest.rs
```diff
@@ -0,0 +1,133 @@
...
+                V: MapAccess<'de>,
+            {
+                let deserializer = MapAccessDeserializer::new(map);
+                DetailedNodeSpecification::deserialize(deserializer)
+                    .map(TomlNodeSpecification::Detailed)
+            }
+        }
+        deserializer.deserialize_any(TomlNodeSpecificationVisitor)
+    }
+}
```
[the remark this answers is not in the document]
knoellle: TOML parsing has been replaced.

crates/source_analyzer/src/manifest.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl FrameworkManifest {
+    pub fn try_from_toml(path: impl AsRef<Path>) -> Result<Self, Error> {
+        let path = path.as_ref();
+        toml::from_str(&read_to_string(path).map_err(|error| Error::Io {
+            source: error,
+            path: path.to_path_buf(),
+        })?)
+        .map_err(|error| Error::ConfigurationParsing {
```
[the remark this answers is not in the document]
knoellle: Error kind has been removed.

crates/source_analyzer/src/manifest.rs
```diff
@@ -0,0 +1,133 @@
...
+            source: error,
+            path: path.to_path_buf(),
+        })
```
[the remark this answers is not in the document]
knoellle: All parsing and error generation has been removed from this file.

crates/source_analyzer/src/struct_hierarchy.rs
knoellle: Done, also changed to derived default at all the call sites.

crates/code_generation/src/perception_databases.rs
knoellle: Renamed to `find_earliest_timestamp` to make the meaning clearer.
The fact that we search for the smallest number is just a detail, we really want the earliest timestamp.

crates/hulk/framework.toml
knoellle: Removed

crates/source_analyzer/src/cyclers.rs
knoellle: Done

Cargo.toml
knoellle: Yes, `depp` still requires it.

crates/hulk/build.rs
knoellle: Considering it says "kind" right next to it already, I think its more readable this way.

crates/source_analyzer/src/structs.rs
```diff
@@ -1,85 +1,63 @@
...
+}
+
+impl Default for Structs {
+    fn default() -> Self {
+        Self {
+            configuration: StructHierarchy::new_struct(),
+            cyclers: Default::default(),
+        }
+    }
 }
```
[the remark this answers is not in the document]
knoellle: It does now.

crates/source_analyzer/src/cyclers.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    let mut sorted_nodes = Vec::new();
+    while let Some(node) = topological_sort.pop() {
+        sorted_nodes.push(node.clone());
+    }
+    if !topological_sort.is_empty() {
+        return Err(Error::CircularDependency);
+    }
```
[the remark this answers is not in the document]
knoellle: 1. While TopologicalSort implements iterator, the item type we use is an `&Node`, so we have to clone the items.
2. `from_iter`, `collect`, etc. all move or consume the underlying iterator. We can get around this by cloning the iterator as you correctly pointed out in your suggestion. However, in this case we are interested in the state of the object after the `next` call returns `None`.
3. If there were no errors, `topological_sort.len()` should be 0, as `pop` actually removes the elements.
  I also think checking whether there are any leftover nodes whose dependencies weren't fulfilled is easier to understand than checking if the number of nodes emitted matches the input. What do you think?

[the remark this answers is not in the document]
knoellle: Regarding 2.:
Just remembered [`by_ref`](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.by_ref) is a thing that exists.
Using that, `cloned`, and `collect` now but left the cycle check untouched.

crates/source_analyzer/src/error.rs
```diff
@@ -0,0 +1,67 @@
...
+        node: String,
+        path: PathBuf,
+    },
+    #[error("invalid module path")]
+    InvalidModulePath,
+    #[error("`{node}` requires output `{output}`, but it is never produced")]
+    MissingOutput { node: String, output: String },
+    #[error("failed to sort nodes, circular dependency detected")]
+    CircularDependency,
+}
```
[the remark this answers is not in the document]
knoellle: Done

tools/pepsi/Cargo.toml
knoellle: Done

crates/source_analyzer/src/contexts.rs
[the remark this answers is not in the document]
knoellle: Readded the tests.

Removed the one regarding cartesian products when having multiple variables in a path since that feature was apparently removed.
Modified the variable expansion function interface on the `Path` type slightly. It doesn't need a whole cycler, just the instances.
Readded a check for AdditionalOutput paths containing optionals to make one of the tests pass again.

crates/source_analyzer/src/structs.rs
knoellle: Done

crates/source_analyzer/src/manifest.rs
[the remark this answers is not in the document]
h3ndrk: Please remove it from the `crates/source_analyzer/Cargo.toml`.

crates/source_analyzer/src/manifest.rs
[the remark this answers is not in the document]
h3ndrk: Yes, but please apply this suggestion to `crates/source_analyzer/nodes.rs`, s.t. all originating errors are named `source` because this will be used by `thiserror` as context information.

crates/source_analyzer/src/contexts.rs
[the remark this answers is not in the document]
h3ndrk: I now noticed that this multi-variable feature is only partially removed and half-done. We should rethink what the goal is:

For now we do not need to support additional variables because I currently see no use-case where we need them (so it is okay that Schmiddy dropped this feature). This may change in the future, but YAGNI.

In addition, variables restrict the scope in a way because they allow conditional branches in the type path. Once a branch is taken by expanding the variable, subsequent branches of the same variable in deeper path segments become obsolete. We therefore only need to support a single occurrence of any variable in the path. In combination that we only need to support cycler instances leads to a single cycler instance per path.

Please adapt the `Path` type s.t. it fits this design decisions best. IMHO, we should also enforce that the cycler instance variable is named `cycler_instance` to communicate a consistent style for our users.

The current implementation is insufficient, because it allows you to write any variable name if you have a `$` character in front. It will still be replaced with cycler instances. In addition, what happens if we have multiple variables in the path? This should also be not possible as discussed above.

crates/hulk_webots/src/hardware_interface.rs
h3ndrk: Thanks, last remark: I meant that if you quick-fix then it will place them in the correct groups, not if you are formatting. But let's discuss this again after RoboCup. Resolve if read.

crates/hulk/build.rs
h3ndrk: But think about how others might use it if it is not used already in your scope. Maybe they rather think in my mentioned "more natural" way and will use `CyclerKind` which results in inconsistent style. Yes, you are right, that this is shorter, but I would weight the risk of future inconsistency higher.

crates/code_generation/src/accessor.rs
```diff
@@ -287,14 +287,7 @@ mod tests {
...
             kind: CyclerKind::RealTime,
-            instances: vec![
-                Instance {
-                    name: "InstanceA".to_string(),
-                },
-                Instance {
-                    name: "InstanceB".to_string(),
-                },
-            ],
+            instances: vec!["InstanceA".to_string(), "InstanceB".to_string()],
```
h3ndrk, line 290: Maybe we already talked about this, is it possible to take `&'static str` instead of `String` here?

knoellle: We talked about this for the manifest, however, this isn't the `CyclerManifest` type but the `Cycler` type.
It shouldn't normally be constructed manually anyways.

Besides, since the instance names are compounds of cycler name and user provided instance names, the new string has to be stored somewhere. If we were to to use string slices here, something else would have to own that string.

crates/code_generation/src/run.rs
```diff
@@ -3,210 +3,43 @@ use std::iter::repeat;
 use convert_case::{Case, Casing};
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
+use source_analyzer::cyclers::{CyclerKind, Cyclers};
 
-use super::{cycler::Cycler, other_cycler::OtherCycler};
+pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
```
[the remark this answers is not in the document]
knoellle: I disagree. Both `mod run_function` and `fn generate_run` sound wrong to me.
I'm for leaving this as is.

crates/hulk/build.rs
knoellle: Not quite convinced but done.

tools/pepsi/src/analyze.rs
knoellle: Done. First expanded to `formatter` but realized that actually doesn't fit here, likely a leftover from a `Display` implementation.
Named `writer` instead

crates/source_analyzer/src/structs.rs
knoellle: Done

crates/control/src/hulk_message_filter.rs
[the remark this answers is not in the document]
knoellle: Reverted.

crates/source_analyzer/src/manifest.rs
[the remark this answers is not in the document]
knoellle: Ah, didn't know that was a magic name. Renamed.

crates/source_analyzer/src/manifest.rs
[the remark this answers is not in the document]
knoellle: Removed the dependency.

crates/source_analyzer/src/structs.rs
knoellle: Returning errors here does not actually improve error messages.
We also don't have spans to the original source location anymore at this point, so we'd have to hand that all the way through. Alternatively, these checks could be duplicated to where the context structs are parsed and we just continue unwrapping here.
Ideally we wouldn't need any checks here because the data types would be designed in such a way that makes invalid states impossible to represent.

crates/source_analyzer/src/uses.rs
knoellle: Undone, also removed the unnecessary trait below.

crates/types/src/camera_position.rs
knoellle: Yes, because this type is now used in the generated configuration structs which require `Default`.

crates/framework/Cargo.toml
knoellle: Done

tests/behavior/golden_goal.lua
knoellle: Done

crates/vision/src/camera_matrix_extractor.rs
knoellle: Done

crates/code_generation/src/accessor.rs
```diff
@@ -1,383 +1,310 @@
...
-                }
-                _ => {
-                    token_stream.append(TokenTree::Ident(format_ident!("{}", segment.name)));
-                }
-            }
+    });
+    if path.contains_optional() {
+        quote! {
+            (|| Some(#prefix . #(#segments).*))()
         }
```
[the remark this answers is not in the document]
knoellle: Why do you want to get rid of it?
What did you have in mind?

In my opinion this is the correct way to implement it, since it's the emitted code is much closer to what the user wrote. We only need to convert some options to references and restrict the "return scope" of the `?` operator.

crates/control/src/behavior_simulator_setup.rs
knoellle: For now, yes.
Removed.

crates/source_analyzer/src/contexts.rs
```diff
@@ -799,50 +756,50 @@ mod tests {
...
+        // // optionals are supported
+        // let field = "PersistentState<usize, \"a.b?.c\">";
+        // let fields = format!("{{ name: {field} }}");
+        // let named_fields: FieldsNamed = parse_str(&fields).unwrap();
+        // assert!(Field::try_from_field(
+        //     named_fields.named.first().unwrap(),
+        //     &empty_uses,
+        //     "CycleContext"
+        // )
+        // .is_ok());
```
knoellle, line 793: @h3ndrk should `PersistentState` support optional path segments?
I added checks for this which broke this test.

crates/code_generation/src/structs.rs
knoellle: All failure cases where this function panics are caused by bugs in the source analyzer. I.e. an improper hierarchy such as a `struct` directly containing another `struct` without a `field` in between.
I think its fine to panic in these cases. I tried to reach these panics with malformed paths but never succeeded.

crates/source_analyzer/src/contexts.rs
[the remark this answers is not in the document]
knoellle: The `Path` implementation now checks for (dis)allowed optional segments, no more than one variable segment as well as the names of variable segments.

crates/context_attribute/src/lib.rs
```diff
@@ -8,16 +8,36 @@ use syn::{
...
+            "PersistentState",
+            "RequiredInput",
+        ]
+        .as_slice(),
+        "MainOutputs" => ["MainOutput"].as_slice(),
+        _ => abort!(
+            struct_item.ident,
+            "unexpected context name, try one of `CreationContext`, `CycleContext`, `MainOutputs`"
+        ),
+    };
```
h3ndrk, line 39: Deduplicate with the function in `source_analyzer`s `contexts.rs`?

h3ndrk (changes requested):

crates/context_attribute/src/lib.rs
knoellle: I tried that, but proc macro crates cannot export anything other than proc macros and having the context attribute crate depend on source analyzer didn't seem right to me either. Do you think that would be better?

crates/hulk/build.rs
```diff
@@ -0,0 +1,118 @@
...
+use color_eyre::eyre::{Result, WrapErr};
+use source_analyzer::{
+    cyclers::{CyclerKind, Cyclers},
+    manifest::{CyclerManifest, FrameworkManifest},
+    pretty::to_string_pretty,
+    structs::Structs,
+};
+
+fn main() -> Result<()> {
+    println!("cargo:rerun-if-changed=framework.toml");
```
h3ndrk, line 11: Remove and check for similar mistakes?

crates/hulk/build.rs
```diff
@@ -0,0 +1,118 @@
...
+                setup_nodes: vec!["audio::microphone_recorder"],
+                nodes: vec!["audio::whistle_detection"],
+            },
+        ],
+    };
+    let root = "..";
+
+    let mut cyclers = Cyclers::try_from_manifest(manifest, root)?;
+    for path in cyclers.watch_paths() {
+        println!("cargo:rerun-if-changed={}", path.display());
```
h3ndrk, line 107: Recheck why the binary rebuilds every time.

crates/code_generation/src/perception_databases.rs
```diff
@@ -0,0 +1,93 @@
...
+        pub struct Updates {
+            #(#updates_fields,)*
+        }
+
+        impl framework::Updates<Databases> for Updates {
+            fn first_timestamp_of_temporary_databases(&self) -> Option<std::time::SystemTime> {
+                #find_earliest_timestamp
+            }
+
+            #[allow(unused_variables)]
```
h3ndrk, line 65: Recheck `#[allow]`s

h3ndrk (changes requested):

crates/hulk/build.rs
knoellle: Removed from both build.rs files.

crates/hulk/build.rs
knoellle: This was indeed caused by the `rerun-if-changed` print for the missing framework.toml.
Now that that's removed it nothing has to rebuilt if no files changed and pepsi build finishes within a few milliseconds.

crates/code_generation/src/perception_databases.rs
knoellle: Checked every `allow` in the repo for whether it can be removed.
Was able to remove most from framework as well as a whole bunch from other crates, mostly no longer dead `dead_code` ones.

h3ndrk (approved):
Thanks @schmidma and @knoellle! 🎉

Let's test this once more and then feel free to merge @knoellle! 🤩

PR #183, opened 2023-04-09
Resubscribe Parameter Panels in Twix

knoellle:
## Introduced Changes

Correctly initialize parameter panels at twix startup.
Previously the previous path would be restored, but the the values wouldn't be subscribed requiring manually clicking on the input field and pressing enter for the current parameter values to show up and be editable.
Also some minor refactoring.

Fixes #

## How to Test

- Open parameter panel and connect to a hulk instance
- Subscribe some path with the parameter panel
- Restart twix, everything should be restored instead of just the path text.

### DIFF
--- a/tools/twix/src/panels/parameter.rs
+++ b/tools/twix/src/panels/parameter.rs
@@ -24,11 +24,14 @@ impl Panel for ParameterPanel {
             Some(Value::String(string)) => string.clone(),
             _ => String::new(),
         };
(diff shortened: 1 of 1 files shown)


tuxbotix:
I encountered the same issue while doing #184, and ended up doing something similar at the initialising time.
LGTM in my opinion.

tuxbotix (approved):
LGTM

PR #189, opened 2023-04-15
Introduce double buffered HULA reader

h3ndrk:
## Introduced Changes

This PR introduces Unix socket draining when reading from the HULA socket. This is needed because our HULK process connects very early to the Unix socket and HULA/LoLA immediately starts sending state messages to HULK which fills our read buffer. After the connection, we are further initializing the HardwareInterface, e.g. doing camera initialization/resetting. This may take a long time, filling up around 1 second of historic state messages in HULKs read buffer. Once the HardwareInterface has been initialized, the cyclers begin to run and the control cycler will rush to drain the accumulated messages, because the sensor data receiver has a lot of data available to feed the cycle. This results in a lot of messages being sent to HULA/LoLA containing historic (1 second old) control messages. But since HULA/LoLA only updates its shared memory and does not communicate directly with the hardware, this is usually not noticed and it will converge quickly to nearly no latency and drained buffers. This bug was discovered because the new HULA will communicate synchronously with the hardware thus needing this fix.

Changes in HULA are not needed. No worries, there is a lot of test code which accumulates a lot of added lines. :wink: 

## ToDo / Known Issues

None

None

## How to Test

The double buffering should have nearly 100% test coverage, but please test on a real NAO if everything works as intended. The chest button LED blinking frequency should be exactly one second immediately on startup which is a good indicator that everything works as intended.

### DIFF
--- a/crates/hulk/src/nao/double_buffered_reader.rs
+++ b/crates/hulk/src/nao/double_buffered_reader.rs
@@ -1,76 +1,161 @@
 use std::{
     io::{self, ErrorKind, Read},
-    mem::size_of,
+    mem::{size_of, MaybeUninit},
     os::unix::{io::AsRawFd, prelude::RawFd},
+    ptr::null_mut,
 };
+
+use libc::{fd_set, select, FD_SET, FD_ZERO};
 
 use super::hula::StateStorage;
 
-pub struct DoubleBufferedReader {
-    buffers: [[u8; size_of::<StateStorage>()]; 2],
+const NUMBER_OF_BUFFERS: usize = 2;
+
+pub struct DoubleBufferedReader<Reader, Poller> {
+    reader: Reader,
+    poller: Poller,
+    buffers: [[u8; size_of::<StateStorage>()]; NUMBER_OF_BUFFERS],
     active_buffer_index: usize,
-    incomplete_offset: usize,
+    number_of_read_bytes_in_active_buffer: usize,
 }
 
-impl Default for DoubleBufferedReader {
-    fn default() -> Self {
+impl<Reader, Poller> DoubleBufferedReader<Reader, Poller>
+where
+    Reader: AsRawFd + Read,
+    Poller: Poll,
+{
+    pub fn from_reader_and_poller(reader: Reader, poller: Poller) -> Self {
         Self {
-            buffers: [[0; size_of::<StateStorage>()]; 2],
+            reader,
+            poller,
+            buffers: [[0; size_of::<StateStorage>()]; NUMBER_OF_BUFFERS],
             active_buffer_index: Default::default(),
-            incomplete_offset: Default::default(),
+            number_of_read_bytes_in_active_buffer: Default::default(),
         }
     }
-}
 
-impl DoubleBufferedReader {
-    fn swap_buffers(&mut self) {
-        self.incomplete_offset = 0;
-        self.active_buffer_index = (self.active_buffer_index + 1) % 2;
+    fn previous_active_buffer_index(&self) -> usize {
+        (self.active_buffer_index + NUMBER_OF_BUFFERS - 1) % NUMBER_OF_BUFFERS
     }
 
-    pub fn read(
-        &mut self,
-        reader: &mut (impl AsRawFd + Read),
-        mut poll_reader: impl FnMut(RawFd) -> io::Result<()>,
-    ) -> io::Result<()> {
-        let mut buffers_swapped = false;
+    fn next_active_buffer_index(&self) -> usize {
+        (self.active_buffer_index + 1) % NUMBER_OF_BUFFERS
+    }
+
+    fn activate_next_buffer(&mut self) {
+        self.active_buffer_index = self.next_active_buffer_index();
+        self.number_of_read_bytes_in_active_buffer = 0;
+    }
+
+    pub fn drain(&mut self) -> io::Result<&StateStorage> {
+        let mut is_at_least_one_buffer_complete = false;
         loop {
-            match reader.read(&mut self.buffers[self.active_buffer_index][self.incomplete_offset..])
-            {
+            match self.reader.read(
+                &mut self.buffers[self.active_buffer_index]
+                    [self.number_of_read_bytes_in_active_buffer..],
+            ) {
                 Ok(number_of_read_bytes) => {
-                    self.incomplete_offset += number_of_read_bytes;
-                    assert!(self.incomplete_offset <= self.buffers[self.active_buffer_index].len());
-                    if self.incomplete_offset == self.buffers[self.active_buffer_index].len() {
-                        self.swap_buffers();
-                        buffers_swapped = true;
+                    self.number_of_read_bytes_in_active_buffer += number_of_read_bytes;
+                    assert!(
+                        self.number_of_read_bytes_in_active_buffer
+                            <= self.buffers[self.active_buffer_index].len()
+                    );
+                    let is_active_buffer_complete = self.number_of_read_bytes_in_active_buffer
+                        == self.buffers[self.active_buffer_index].len();
+                    if is_active_buffer_complete {
+                        self.activate_next_buffer();
+                        is_at_least_one_buffer_complete = true;
                     }
                 }
                 Err(ref error) if error.kind() == ErrorKind::WouldBlock => {
-                    if buffers_swapped {
-                        return Ok(());
+                    if is_at_least_one_buffer_complete {
+                        return Ok(unsafe {
+                            &*(self.buffers[self.previous_active_buffer_index()].as_ptr()
+                                as *const StateStorage)
+                        });
                     }
-                    poll_reader(reader.as_raw_fd())?;
+                    self.poller.poll(self.reader.as_raw_fd())?;
                 }
                 Err(error) => return Err(error),
             }
         }
     }
+}
 
-    /// Precondition: `read()` has been called once
-    pub fn get_last(&self) -> &StateStorage {
-        let inactive_buffer_index = (self.active_buffer_index + 1) % 2;
-        unsafe { &*(self.buffers[inactive_buffer_index].as_ptr() as *const StateStorage) }
+pub trait Poll {
+    fn poll(&mut self, file_descriptor: RawFd) -> io::Result<()>;
+}
+
+pub struct SelectPoller;
+
+impl Poll for SelectPoller {
+    fn poll(&mut self, file_descriptor: RawFd) -> io::Result<()> {
+        unsafe {
+            let mut set = MaybeUninit::<fd_set>::uninit();
+            FD_ZERO(set.as_mut_ptr());
+            let mut set = set.assume_init();
+            FD_SET(file_descriptor, &mut set);
+            if select(
+                file_descriptor + 1,
+                &mut set,
+                null_mut(),
+                null_mut(),
+                null_mut(),
+            ) < 0
+            {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(())
+        }
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use std::slice::from_raw_parts;
+    use std::{
+        slice::from_raw_parts,
+        sync::{
+            atomic::{AtomicUsize, Ordering},
+            Arc,
+        },
+    };
 
     use super::*;
 
+    struct PanickingPoller;
+
+    impl Poll for PanickingPoller {
+        fn poll(&mut self, _file_descriptor: RawFd) -> io::Result<()> {
+            panic!("should not be called");
+        }
+    }
+
+    struct ErroringPoller;
+
+    impl Poll for ErroringPoller {
+        fn poll(&mut self, file_descriptor: RawFd) -> io::Result<()> {
+            assert_eq!(file_descriptor, FIXED_FILE_DESCRIPTOR);
+            Err(ErrorKind::ConnectionAborted.into())
+        }
+    }
+
+    const FIXED_FILE_DESCRIPTOR: RawFd = 42;
+
+    #[derive(Default)]
+    struct CountingPoller {
+        number_of_polls: Arc<AtomicUsize>,
+    }
+
+    impl Poll for CountingPoller {
+        fn poll(&mut self, file_descriptor: RawFd) -> io::Result<()> {
+            assert_eq!(file_descriptor, FIXED_FILE_DESCRIPTOR);
+            self.number_of_polls.fetch_add(1, Ordering::SeqCst);
+            Ok(())
+        }
+    }
+
     #[test]
-    fn error_is_returned() {
+    fn read_error_is_returned() {
         struct Reader;
         impl AsRawFd for Reader {
             fn as_raw_fd(&self) -> RawFd {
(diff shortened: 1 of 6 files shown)


crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
+use std::{
+    io::{self, ErrorKind, Read},
+    mem::size_of,
+    os::unix::{io::AsRawFd, prelude::RawFd},
+};
+
+use super::hula::StateStorage;
+
+pub struct DoubleBufferedReader {
```
schmidma, line 9: What do you think of this with some generics in an extra crate? Or is it overkill?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+            buffers: [[0; size_of::<StateStorage>()]; 2],
+            active_buffer_index: Default::default(),
+            incomplete_offset: Default::default(),
+        }
+    }
+}
+
+impl DoubleBufferedReader {
+    fn swap_buffers(&mut self) {
+        self.incomplete_offset = 0;
```
schmidma, line 27: `number_of_incomplete_bytes` ?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+        self.active_buffer_index = (self.active_buffer_index + 1) % 2;
```
schmidma, line 28: `2` -> `NUMBER_OF_BUFFERS` ?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    pub fn read(
+        &mut self,
+        reader: &mut (impl AsRawFd + Read),
```
schmidma, line 33: An idea:
Is it possible to split the `UnixStream` into an reader and writer part?
I want to reach a point where the reader can be consumed/owned by the `DoubleBufferedReader`.
This means, moving a hypothetical reading part to `Self` and drop the argument here.

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+        mut poll_reader: impl FnMut(RawFd) -> io::Result<()>,
+    ) -> io::Result<()> {
+        let mut buffers_swapped = false;
```
schmidma, line 36: Naming things: This boolean, is used to express whether a full message was received since the last read, correct? Maybe something like `is_buffer_complete`?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+        loop {
+            match reader.read(&mut self.buffers[self.active_buffer_index][self.incomplete_offset..])
+            {
+                Ok(number_of_read_bytes) => {
+                    self.incomplete_offset += number_of_read_bytes;
+                    assert!(self.incomplete_offset <= self.buffers[self.active_buffer_index].len());
+                    if self.incomplete_offset == self.buffers[self.active_buffer_index].len() {
```
schmidma, line 43:
```suggestion
                    let is_message_complete (?) = self.incomplete_offset == self.buffers[self.active_buffer_index].len();
                    if  is_message_complete {
```
not sure about the naming, giving this comparison a name leads to better documented behavior

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+                    poll_reader(reader.as_raw_fd())?;
+                }
+                Err(error) => return Err(error),
+            }
+        }
+    }
+
+    /// Precondition: `read()` has been called once
+    pub fn get_last(&self) -> &StateStorage {
+        let inactive_buffer_index = (self.active_buffer_index + 1) % 2;
```
schmidma, line 61: This `[...] + 1 % 2` toggle computation is used multiple times? This could be a function? Or even more explicit, use an `enum` representing a tri-state instead of indices, which also covers the case of being uninitialized?

crates/hulk/src/nao/hula.rs
```diff
@@ -363,7 +366,7 @@ impl From<types::Ear> for Ear {
     }
 }
 
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
```
schmidma, line 369: why is this needed?

crates/hulk/src/nao/hula.rs
```diff
@@ -377,10 +380,30 @@ pub struct ControlStorage {
...
+    stream: &mut UnixStream,
+    reader: &mut DoubleBufferedReader,
+) -> Result<StateStorage> {
+    reader
+        .read(stream, |file_descriptor| unsafe {
+            let mut set = MaybeUninit::<fd_set>::uninit();
+            FD_ZERO(set.as_mut_ptr());
+            let mut set = set.assume_init();
+            FD_SET(file_descriptor, &mut set);
+            if select(
```
schmidma, line 393: Can you elaborate what this does? Is there nothing that already implements such a behavior as an API function on some file descriptor type?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+                    }
+                    poll_reader(reader.as_raw_fd())?;
+                }
+                Err(error) => return Err(error),
+            }
+        }
+    }
+
+    /// Precondition: `read()` has been called once
+    pub fn get_last(&self) -> &StateStorage {
```
schmidma, line 60: I'm not a fan of `get_[...]` maybe just drop the `get`-part?

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+}
+
+impl DoubleBufferedReader {
+    fn swap_buffers(&mut self) {
+        self.incomplete_offset = 0;
+        self.active_buffer_index = (self.active_buffer_index + 1) % 2;
+    }
+
+    pub fn read(
```
schmidma, line 31:
```suggestion
    pub fn draining_read(
```

??? What do you think?

crates/hulk/src/nao/double_buffered_reader.rs
schmidma: Another suggestion is `buffered()`? Because `last` indicates, there is more to get, but the user can actually always only get one complete message out of the buffer. So there is no first or last, there is just one `buffered` message.

I'm open to for suggestions on the actual name, but i would like to see this concept represented.

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    /// Precondition: `read()` has been called once
+    pub fn get_last(&self) -> &StateStorage {
+        let inactive_buffer_index = (self.active_buffer_index + 1) % 2;
+        unsafe { &*(self.buffers[inactive_buffer_index].as_ptr() as *const StateStorage) }
+    }
+}
+
+#[cfg(test)]
```
schmidma, line 66: I did not read the tests yet, but think about splitting the tests into "unit-tests" in this module file, and integration tests, which test the API available to the public user.

schmidma (changes requested):
Some general remarks and mostly nit-picking on naming.

One general question:
You are concerned a lot about partial messages and stitching them back together afterwards. Is this actually happening? I assumed the unix socket to have extremely high memory bounds on the internal buffers? Or why are you concerned with partial messages? Please elaborate.

h3ndrk:
> I assumed the unix socket to have extremely high memory bounds on the internal buffers?

I [read that Unix sockets have around 100-200 KB of buffer space](https://unix.stackexchange.com/a/424381). HULA sends messages of 652 bytes which means the buffer is full after around 300 messages which is approximately 3-4 seconds at a rate of 80 Hertz. When the buffer gets full, the likelihood that only a partial message still fits is very high. In my experiments, I saw around 80 KB of buffered messages to be usual during initialization. Hopefully, this explains the message stitching efforts.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: I wanted to define `pub struct DoubleBufferedReader<T>` but this does not work because I used `size_of::<T>()` in the type which is [currently unstable](https://github.com/rust-lang/rust/issues/60551). We could try but I thought, I don't need to spend effort generalizing for this single instantiation. But yes, if you have an idea, how we can implement it, I'd be happy to do so.

crates/hulk/src/nao/hula.rs
h3ndrk: Mainly to compare the types in assertions in the tests.

crates/hulk/src/nao/hula.rs
h3ndrk: Since this is Unix only functionality and Rust's standard library does not provide such high-level constructs, there is nothing available officially. I decided to use the libc call directly which is unsafe but has is a well-known API and does not introduce heavy abstractions. My goal was to find a thing that allows me to simply wait for a single file descriptor becoming readable. The other Linux polling methods would also work but `select()` is one of the simpler ways to do it. We could search for a safe abstraction over `select()` to get rid of the `unsafe {}` here. What is your opinion or do you know another way to use standard functionality that I didn't see?

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: IIRC testing public API is not the separation argument between unit and integration tests. Integration tests are testing a unit together with other units. Since I used traits to only implement the functionality that I need, I would argue that the `DoubleBufferedReader` is a unit and all tests in this section mock all I/O and therefore fall into the unit-test category. Correct me if I'm wrong. :wink:

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: It is now named `number_of_read_bytes_in_active_buffer`.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: Abstracted.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: The reader and poller are now wrapped in the `DoubleBufferedReader`.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: It is now named `is_at_least_one_buffer_complete`.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: Is extracted as `is_active_buffer_complete` to align with the `is_at_least_one_buffer_complete` variable.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: The index calculations are now extracted as methods.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: The `DoubleBufferingReader::drain()` now returns the read value directly. The precondition is also enforced by that design. `comments -= 1`

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: I called it not `drain()`. I'm not 100% happy with this name because it actually drains and returns the last complete value. `drain()` does not reflect the value returning but maybe that is okay.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: We could also add integration tests but I think this piece of code will be tested enough in using the NAO in everyday life situations.

crates/hulk/src/nao/double_buffered_reader.rs
schmidma: Hmm.. i'm fine with either approach. One solution might be to drop `size_of::<T>()` and replace it with `const N: usize` or similar.
Another idea, can you just make an array of `T` and reinterpret it to a `[u8]` when reading from the socket?

crates/hulk/src/nao/hula.rs
schmidma: No, I just assumed, there already is a safe abstraction. I now, that for example https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html#method.readable does something like you did manually. Maybe there is a crate for these functionalities out there?!

crates/hulk/src/nao/double_buffered_reader.rs
schmidma: but `draining_read` does? What is your objection to that name?

crates/hulk/src/nao/double_buffered_reader.rs
schmidma: Maybe my current neovim setup just lacks easy folding of parts, but I am always a little distracted by these huge parts of testing code in the implementation. But I think there is not a pretty fix to it. Feel free to resolve when read or start a discussion ;)

schmidma:
But did you ever observe partial messages in the socket buffer? And what is HULA doing then? Is it blocking? Crashing?!

crates/hulk/src/nao/hula.rs
schmidma: Ok, I was shortly thinking about a way to represent the fact, that it is only needed for testing. But never mind. I'm fine, if you have no further ideas.
(Feel free to resolve when read)

crates/hulk/src/nao/double_buffered_reader.rs
```diff
@@ -0,0 +1,513 @@
...
+
+use libc::{fd_set, select, FD_SET, FD_ZERO};
+
+use super::hula::StateStorage;
+
+const NUMBER_OF_BUFFERS: usize = 2;
+
+pub struct DoubleBufferedReader<Reader, Poller> {
+    reader: Reader,
+    poller: Poller,
```
schmidma, line 16: Why is it generic over the `poller`?

crates/hulk/src/nao/double_buffered_reader.rs
schmidma: What do you think about the most explicit approach with `enum`s? I'm not a fan of indices

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: Well we could move the code into its own file but then we would do it differently than in the other files. Do we want this Präzedenzfall? Do we want this everywhere? I think it is okay to have it near the implementations since it is always at the bottom and noone forces you to scroll that far. Now you resolve again 😉

crates/hulk/src/nao/hula.rs
h3ndrk: Maybe this? https://github.com/DoumanAsh/selecting Sadly there is no `readable()` in non-async land as far as I know. IMHO the benefit for such simple functionality is debatable.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: At some point we need the indices again if we want to peek into the `Self::buffers`. IMHO it is okay to use indices consistently and as a bonus we are general about the number of buffers.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: For dependency injection for testability.

h3ndrk:
> But did you ever observe partial messages in the socket buffer?

No, the scenario would happen if HULK pauses reading for 3-4 seconds, then reads faster than HULA being able to write the remaining message. This could happen if the scheduler decides to not schedule HULA for a short moment in time.

> And what is HULA doing then? Is it blocking?

Yes

> Crashing?!

No, but it will wait for everything to be written until reading more messages from LoLA.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: It is now generic over the read type. It is still located in the `hulk` crate because this is the single usage for now. We can extract it into a crate later.

crates/hulk/src/nao/double_buffered_reader.rs
h3ndrk: Renamed to `draining_read()`.

crates/hulk/src/nao/hula.rs
```diff
@@ -377,10 +378,16 @@ pub struct ControlStorage {
     pub stiffness: JointsArray,
 }
 
-pub fn read_from_hula(stream: &mut UnixStream) -> Result<StateStorage> {
-    let mut read_buffer = [0; size_of::<StateStorage>()];
-    stream.read_exact(&mut read_buffer)?;
-    Ok(unsafe { read(read_buffer.as_ptr() as *const StateStorage) })
+pub fn read_from_hula<Reader, Poller>(
```
schmidma, line 381: Does this need to be generic over the `Poller`?

h3ndrk: Removed generic parameters of reader and poller.

h3ndrk:
@schmidma Tested successfully on a NAO with legacy HULA.

schmidma (approved):

PR #190, opened 2023-04-16
Merge Robots

### LINKED ISSUE #122: Convert Bounding Boxes to Robot Perceptions

schmidma:
## Introduced Changes

Closes #122 

Merge robot perceptions in the robot detection. The approaches are very easy and probably have to be improved when testing extensively.

## ToDo / Known Issues

none

none

## How to Test

Deploy a robot, use twix to tune filtering parameters of the robot detection and view it via the respective image overlay


crates/projection/src/lib.rs
```diff
@@ -101,6 +102,28 @@ impl Projection for CameraMatrix {
...
+    ) -> Result<Point3<f32>, Error> {
+        if x <= 0.0 {
+            return Err(Error::BehindCamera);
+        }
+
+        let camera_ray = self.pixel_to_camera(pixel_coordinates);
+        let camera_ray_over_robot = self.camera_to_robot.rotation * camera_ray;
+
+        let elevation = x - self.camera_to_robot.translation.x;
+        let slope = elevation / camera_ray_over_robot.x;
```
knoellle, line 118: It took me quite a while to understand what this does.
While I think these names are correct, I would suggest variable names that better express the greater geometrical meaning.
Essentially we are intersecting a ray with an axis-aligned plane so perhaps call it something along the lines of `distance_to_plane` and `ray_scale`. Not sure about the second one, if the ray was always of length 1, I would call it ray length, but it doesn't. `ray_repetitions_to_reach_plane`?

The same applies to the `pixel_to_robot_with_z` function.

knoellle (changes requested):
Code looks good except for some naming stuff, will test tomorrow.

crates/projection/src/lib.rs
```diff
@@ -101,6 +102,28 @@ impl Projection for CameraMatrix {
...
+
+        Ok(point![
+            x,
+            self.camera_to_robot.translation.y + camera_ray_over_robot.y * slope,
+            self.camera_to_robot.translation.z + camera_ray_over_robot.z * slope
+        ])
```
knoellle, line 124:
```suggestion
        let intersection_point = self.camera_to_robot.translation + camera_ray_over_robot * slope;
        Ok(point![x, intersection_point.y, intersection_point.z])
```

knoellle (approved):
LGTM

PR #194, opened 2023-04-17
Add MotionType to execute an energy saving stand

### LINKED ISSUE #129: Energy saving stand
Implement an additional standing motion to reduce joint heat and stress.
This involves implementing additional states in the walking engine and evaluating how such a stand motion should look like. This can be a controller gradually reducing stiffness but also a standing position that results in low energy demand.

schmidma:
## Introduced Changes

This introduces a new boolean flag, whether a standing motion should be "energy-saving".
An energy saving stand is executed by an independent motion node, which currently only publishes penalized angles. A normal stand is still done by walking.

Fixes #129 

## ToDo / Known Issues

none

- a more sophisticated version which actually uses control to save energy (e.g. gradually decreasing stiffness)
- lean the robot a bit to the front when penalized as it tends to fall backwards much easier than to the front

## How to Test

Testgame? Alternatively deploy the nao as the keeper, as it is standing with energy saving in the goal center without a ball

### DIFF
--- a/crates/control/src/motion/energy_saving_stand.rs
+++ b/crates/control/src/motion/energy_saving_stand.rs
@@ -1,0 +1,50 @@
+use color_eyre::Result;
+use context_attribute::context;
+use framework::MainOutput;
+use types::{
+    ArmJoints, BodyJoints, BodyJointsCommand, CycleTime, Joints, LegJoints, MotionSelection,
+    SensorData,
+};
+
+pub struct EnergySavingStand {}
+
+#[context]
+pub struct CreationContext {}
+
+#[context]
+pub struct CycleContext {
+    pub cycle_time: Input<CycleTime, "cycle_time">,
+    pub motion_selection: Input<MotionSelection, "motion_selection">,
+    pub sensor_data: Input<SensorData, "sensor_data">,
+
+    pub arm_stiffness: Parameter<f32, "energy_saving_stand.arm_stiffness">,
+    pub leg_stiffness: Parameter<f32, "energy_saving_stand.leg_stiffness">,
+    pub penalized_pose: Parameter<Joints<f32>, "penalized_pose">,
+}
+
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {
+    pub energy_saving_stand_command: MainOutput<BodyJointsCommand<f32>>,
+}
+
+impl EnergySavingStand {
+    pub fn new(_context: CreationContext) -> Result<Self> {
+        Ok(Self {})
+    }
+
+    pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+        Ok(MainOutputs {
+            energy_saving_stand_command: BodyJointsCommand {
+                positions: BodyJoints::from(*context.penalized_pose),
+                stiffnesses: BodyJoints {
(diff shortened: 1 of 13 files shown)


schmidma:
successfully tested in today's testgame. We observed some possible improvements, like slightly leaning forward in the standing position, but this may also be improved in future iterations

crates/control/src/motion/energy_saving_stand.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct EnergySavingStand {}
+
+#[context]
+pub struct CreationContext {}
+
+#[context]
+pub struct CycleContext {
+    pub motion_selection: Input<MotionSelection, "motion_selection">,
+    pub sensor_data: Input<SensorData, "sensor_data">,
+    pub cycle_time: Input<CycleTime, "cycle_time">,
```
okiwi6, line 18: sort alphabetically?

crates/control/src/motion/energy_saving_stand.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    pub penalized_pose: Parameter<Joints, "penalized_pose">,
+    pub arm_stiffness: Parameter<f32, "energy_saving_stand.arm_stiffness">,
+    pub leg_stiffness: Parameter<f32, "energy_saving_stand.leg_stiffness">,
```
okiwi6, line 22: sort alphabetically?

crates/control/src/behavior/fall_safely.rs
```diff
@@ -1,6 +1,9 @@
-use types::{FallState, MotionCommand, WorldState};
+use types::{FallState, MotionCommand, PrimaryState, WorldState};
 
 pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+    if world_state.robot.primary_state == PrimaryState::Initial {
+        return None;
+    }
```
okiwi6, line 6: Why is this required now?

crates/control/src/behavior/stand_up.rs
```diff
@@ -1,6 +1,9 @@
-use types::{FallState, MotionCommand, WorldState};
+use types::{FallState, MotionCommand, PrimaryState, WorldState};
 
 pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+    if world_state.robot.primary_state == PrimaryState::Initial {
+        return None;
+    }
```
okiwi6, line 6: Why is this required now?

okiwi6 (changes requested):
Looks nice, only have two questions and noticed non-alphabetical ordering :)

crates/control/src/motion/energy_saving_stand.rs
schmidma: done

crates/control/src/motion/energy_saving_stand.rs
schmidma: done

crates/control/src/behavior/stand_up.rs
schmidma: I don't remember, removed now

okiwi6 (approved):

PR #195, opened 2023-04-17
Add serialize_hierarchy(bound="...") attribute

schmidma:
## Introduced Changes

Adds a new type attribute to specify where predicates for generics.

## ToDo / Known Issues

none

- add some logic to infer the trait bounds based on the given type parameters and generics and use `bound = ...` only to overwrite this guess.

## How to Test

Derive `SerializeHierarchy` to any generic struct and use `#[serialize_hierarchy(bound="...")` to specify the necessary trait bounds.

### DIFF
--- a/crates/serialize_hierarchy_derive/src/lib.rs
+++ b/crates/serialize_hierarchy_derive/src/lib.rs
@@ -1,15 +1,12 @@
 use std::collections::HashSet;
 
 use proc_macro2::TokenStream;
-use proc_macro_error::abort;
-use proc_macro_error::proc_macro_error;
-use quote::quote;
-use quote::ToTokens;
+use proc_macro_error::{abort, proc_macro_error};
+use quote::{quote, ToTokens};
 use syn::{
     parse_macro_input, punctuated::Punctuated, Data, DataStruct, DeriveInput, Generics, Ident, Lit,
-    Meta, NestedMeta, Token, WherePredicate,
+    Meta, MetaNameValue, NestedMeta, Token, Type, WherePredicate,
 };
-use syn::{MetaNameValue, Type};
 
 #[proc_macro_derive(SerializeHierarchy, attributes(serialize_hierarchy))]
 #[proc_macro_error]
--- a/crates/serialize_hierarchy_derive/src/lib.rs
+++ b/crates/serialize_hierarchy_derive/src/lib.rs
@@ -158,7 +158,8 @@ fn extend_where_clause_from_attributes(
                 TypeAttribute::Bounds { predicates } => Some(predicates),
                 _ => None,
             })
-            .flat_map(|predicates| predicates.to_vec())
+            .flatten()
+            .cloned()
     });
 }
 
--- a/crates/serialize_hierarchy_derive/src/lib.rs
+++ b/crates/serialize_hierarchy_derive/src/lib.rs
@@ -5,8 +5,11 @@ use proc_macro_error::abort;
 use proc_macro_error::proc_macro_error;
 use quote::quote;
 use quote::ToTokens;
-use syn::Type;
-use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Ident, Meta, NestedMeta};
+use syn::{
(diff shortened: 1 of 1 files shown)


crates/serialize_hierarchy_derive/src/lib.rs
```diff
@@ -266,6 +287,26 @@ fn parse_attributes(attrs: &[syn::Attribute]) -> HashSet<TypeAttribute> {
         .flat_map(parse_meta_items)
         .map(|meta| match meta {
             NestedMeta::Meta(Meta::Path(word)) if word.is_ident("as_jpeg") => TypeAttribute::AsJpeg,
+            NestedMeta::Meta(Meta::NameValue(MetaNameValue {
+                path, lit: literal, ..
+            })) if path.is_ident("bound") => {
```
h3ndrk, line 292: Instead of acquiring the trait bounds as string, couldn't you just parse the trait bounds on the struct? Yes, that would mean that you will have more strict trait bounds on the struct if you are not using the `SerializeHierarchy` trait but I think there is a trade-off between more strict trait bounds and a string Freitextfeld.

crates/serialize_hierarchy_derive/src/lib.rs
```diff
@@ -145,6 +150,21 @@ fn process_input(input: DeriveInput) -> TokenStream {
...
+    type_attributes: HashSet<TypeAttribute>,
+) {
+    generics.make_where_clause().predicates.extend({
+        type_attributes
+            .iter()
+            .filter_map(|attribute| match attribute {
+                TypeAttribute::Bounds { predicates } => Some(predicates),
+                _ => None,
+            })
+            .flat_map(|predicates| predicates.to_vec())
```
h3ndrk, line 164: Does that work?

```suggestion
            .flatten().cloned()
```

h3ndrk (changes requested):

crates/serialize_hierarchy_derive/src/lib.rs
schmidma: The string is not a string. It is parsed as a list of trait bounds and reports the correct errors. I think this is already very pretty.
The trait bounds on the struct are too strict. The bounds should only be on the respective implementation

schmidma: In the future, we might be able to automatically guess the appropriate bounds and the user does not have to pass this attribute anymore (like `serde` does)

crates/serialize_hierarchy_derive/src/lib.rs
schmidma: It actually does. Thanks

crates/serialize_hierarchy_derive/src/lib.rs
```diff
@@ -5,8 +5,11 @@ use proc_macro_error::abort;
 use proc_macro_error::proc_macro_error;
 use quote::quote;
 use quote::ToTokens;
-use syn::Type;
-use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Ident, Meta, NestedMeta};
+use syn::{
+    parse_macro_input, punctuated::Punctuated, Data, DataStruct, DeriveInput, Generics, Ident, Lit,
+    Meta, NestedMeta, Token, WherePredicate,
+};
+use syn::{MetaNameValue, Type};
```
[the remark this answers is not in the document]
schmidma: Is it possible to configure `rustfmt` or `rust-analyer` to do that automatically on format?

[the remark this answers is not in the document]
h3ndrk: There is the `merge_imports` setting which is deprecated and you should now use `imports_granularity` which is unstable 🤡

crates/serialize_hierarchy_derive/src/lib.rs
h3ndrk: Ok, I predicted your answer. Fine for me.

h3ndrk (approved):

PR #205, opened 2023-04-20
Sit down fix

jonathan-hellwig:
## Introduced Changes

This PR fixes the state transition to unstiff. Previously, by pressing the head buttons we used PrimaryState::Finish to request a sit down. Now, we request unstiff and transition to sit down before going to unstiff.

## How to Test

- deploy the code and check whether the sit down motion gets executed correctly in various states


schmidma (approved):
LGTM, tested in today's test game

PR #207, opened 2023-04-21
Refactor active vision

PasGl:
## Introduced Changes

Non-Strikers:
 - The "look forward" position (currently [1.0,0.0]) is turned into a parameter
 - The order of "positions of interest" is set to make sure every other position is either ball or forward
- replaces index with `enum`
- store last location (e.g. the closest obstacle/field mark may change, which otherwise leads to weird behavior)

Striker:
- Fixate on ball

## ToDo / Known Issues

none

Switch Strikers from fixation on ball to a look-action that makes him around the ball while keeping the ball in the field of view

## How to Test

Probably best tested in a real test-game


crates/control/src/active_vision.rs
```diff
@@ -146,3 +157,65 @@ fn generate_field_mark_positions(field_dimensions: &FieldDimensions) -> Vec<Poin
...
+            let field_mark_of_interest =
+                closest_field_mark_visible(field_mark_positions, parameters, robot_to_field);
+
+            if let Some(field_mark_position) = field_mark_of_interest {
+                PointOfInterest::FieldMark {
+                    absolute_position: robot_to_field * field_mark_position,
+                }
+            } else if let Some(..) = ball {
+                PointOfInterest::Ball
+            } else {
```
julianschuler, line 180: What do you think about the following to replace the `if else if else` pattern?

```suggestion
            match (field_mark_of_interest, ball) {
                (Some(field_mark_position), _) => PointOfInterest::FieldMark {
                    absolute_position: robot_to_field * field_mark_position,
                },
                (None, Some(_)) => PointOfInterest::Ball,
                (_, _) => {
```

PasGl: Done

PR #209, opened 2023-04-22
Improve fall motion

Vivituhh:
## Introduced Changes

When a NAO is falling, it currently doesn't do much except reducing the stiffness and moving the head slightly.

The new falling motion contains bending the knees and unrolling. This should reduce the strain on impact.

Create the motion such that it brings the NAO into a position more favourable for the following stand-up motion.

## How to Test

Test in Webots: Start Webots with `webots/worlds/falling.wbt` and start the simulation. Robot should then do the motion in the next 5s.

Test in Lab with Robot: Push the robot. Moreover it was tested in a test game.


crates/control/Cargo.toml
```diff
@@ -7,6 +7,7 @@ homepage = "https://github.com/hulks/hulk"
 
 [dependencies]
 approx = { workspace = true }
+byteorder = { workspace = true }
```
MaikRe, line 10: Make sure to rebase, this dependency was removed from the webots falling branch before it was merged as it is unused

etc/configuration/default.json
```diff
@@ -633,7 +633,7 @@
     "angular_velocity_low_pass_factor": 0.2,
     "roll_pitch_low_pass_factor": 0.2,
     "gravitational_acceleration_threshold": 4.0,
-    "falling_angle_threshold": [0.5, 0.45],
+    "falling_angle_threshold": [0.5, 0.4],
```
MaikRe, line 636: These are changed by #212 which is already approved for merge, you may need to adjust accordingly and potentially retest to make sure functionality remains as expected

crates/control/Cargo.toml
Vivituhh: Done

etc/configuration/default.json
Vivituhh: Done. It is changed and the motion is adapted to this parameter.

julianschuler (approved):

PR #213, opened 2023-04-23
Improve SitDown keyframes

okiwi6:
## Introduced Changes

I noticed the roboter was hitting its arms on its knees during sitdown.
This effect is greatly reduced in this PR and on another positive note the sitdown animation duration was reduced by 0.5s.

Fixes #

## ToDo / Known Issues

* We also noticed that the roboter hits itself during the transition to Penalized, this should also be fixed

## How to Test
* Upload and observe the Sit Down motion, the robot should hit itself less

### DIFF
--- a/etc/motions/sit_down.json
+++ b/etc/motions/sit_down.json
@@ -5,20 +5,20 @@
       "pitch": 0.1993780136108398
     },
     "left_arm": {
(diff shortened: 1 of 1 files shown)


knoellle (approved):
Nice, LGTM

PR #218, opened 2023-04-25
Remove incremental target in CI

julianschuler:
## Introduced Changes

When e.g. a format runner fails, the CI terminates the other runners as well. In the case of the incremental target, this can result in an invalid cache resulting in runs of e.g. other PRs failing.

This PR removes the incremental targets, since the CI already builds for release and dev.

## ToDo / Known Issues

None.

None.

## How to Test

Observe the CI targets.


schmidma (approved):
yep

PR #225, opened 2023-04-26
Hotfix for dispatching transition

okiwi6:
## Introduced Changes
Hotfix for falling robots in the dispatching interpolator using the maximum velocity approach.
For know switch back to static duration 1.0s for every transition

Fixes #

## ToDo / Known Issues

This will be dealt with properly when I have some time again :)

Konrad mentioned that I also wrongly assumed linear interpolation, instead an interpolation mode aware maximum velocity is required.

## How to Test
Transition Penalized to Walking, Robot should not be falling anymore.
We saw this in action in the testgame on April 26 at CFEL.

### DIFF
--- a/crates/control/src/motion/dispatching_interpolator.rs
+++ b/crates/control/src/motion/dispatching_interpolator.rs
@@ -104,10 +104,10 @@ impl DispatchingInterpolator {
                 ),
             };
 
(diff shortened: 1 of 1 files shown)


julianschuler (approved):
LGTM!

PR #226, opened 2023-04-26
Add separate pose for energy saving stand

julianschuler:
## Introduced Changes

With the new energy saving stand introduced in #194 more robots tend to fall.

This PR introduces a pose separate from the penalized pose with the `hip_pitch` angle rotated 5 degrees forward.This improves stability for the energy saving stand itself and the transitions to and from it.

## ToDo / Known Issues

None.

The positions of the energy saving stand may be tuned further.

## How to Test

Flash a NAO as keeper and observe the behaviour when transitioning in and out of the energy saving stand.


schluis (approved):
Very nice!

PR #253, opened 2023-04-27
Move kick off defender further away from center circle

### LINKED ISSUE #243: Increase Kick-off Defense Distance
The kick off defender is standing too close to the center circle, making to slight delocalizations result in illegal position penalties.

knoellle:
## Introduced Changes

Title.
This is to prevent a slightly delocalized kick off defending striker from accidentally entering the center circle, violating the rules.
Also makes the position relative to the center circle border instead of relative to the center of the field.

Fixes #243 

## How to Test

Do a kick off where the opponent has kick off. The defending striker should stand about 30cm away from the center circle.


schluis (approved):
LGTM

PR #254, opened 2023-04-28
move striker further back in penalty kick

### LINKED ISSUE #252: Penaltykick Striker position
Penaltykick striker position is too close to penalty spot when hulks are kicking.
Value should either be statically increased or use new kick decisions when available

MaikRe:
## Introduced Changes

Move striker back by another penalty marker size to make sure he doesn't block the penalty mark during a penalty kick.

Fixes #252 

## ToDo / Known Issues

Could be changed to using a new parameter, but optimally should eventually use kick decisions and thus a new parameter would then be removed again. I'm open to discuss.


knoellle (approved):
LGTM

PR #248, opened 2023-04-27
Move JSON + parameter matters out of communication server crate

tuxbotix:
## Introduced Changes

See the issue description for more information. Also some discussion happened in #184.

Fixes #219
**Might be a partial fix?**

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

*Describe how to test your changes. (For the reviewer)*

PR #267, opened 2023-04-28
Improve twix parameter save diff

### LINKED ISSUE #240: Parameter save to disk does not do diff against thefull parameter tree.
Related: #219 

At the moment the twix facility for save-to-disk only read the configuration of a robot's head and check the difference against what is set.

But this is incomplete compared to what is done in the communication server where it checks against the complete parameter tree constructed from defaults, head IDs, body IDs, location, etc.

fix:
Expose "deserialize" at communication server so that the diff can be done against the full tree.  It is not accessible for Twix due to a cargo feature activating it only for framework* compilation.

Once exposed:

```rust
let stored_value = deserialize();

let diff = stored_value.clone();
json_merge(diff, incoming_value);

prune_equal_branches(diff, stored_value);
```

tuxbotix:
REQUIRES #248 
## Introduced Changes

Fullfills the following.

```rust
let stored_value = json!({"a":{"b":[1,2,3], "c":10}, "x":1000});
let incoming_sparse_value = json!({"a":{"b":[1,4,3], "c":10}});

// Only "b" has changes.
let expected_diff = json!({"a":{"b":[1,4,3]}});

assert_eq!(
    get_diff_against_stored_value(&stored_value, &incoming_sparse_value),
    expected_diff
);
```
See the issue for further details.

Fixes #240 

## ToDo / Known Issues

We still directly write to head json of a particular robot. Locations are not respected (unlike `serialize()` of parameters crate (Moved from communication server in #248 ))

* Further refactoring on these crates and how they are used could be done.

## How to Test

Cargo test on twix should do the trick ;)
OR
Open twix open manual calibration, change only extrinsic parameters.
Once you click save, only extrinsic parameter changes should show up in the json file. (Before this fix, both intrinsic and extrinsic showed up.)

### DIFF
--- a/crates/parameters/src/directory.rs
+++ b/crates/parameters/src/directory.rs
@@ -130,7 +130,7 @@ where
 
     let mut location_head_parameters = from_path(&location_head_file_path)
         .await
-        .map_err(DirectoryError::BodyParametersOfLocationNotGet)?;
+        .map_err(DirectoryError::HeadParametersOfLocationNotGet)?;
     merge_json(&mut location_head_parameters, &parameters);
 
     to_path(location_head_file_path, location_head_parameters)
--- a/crates/parameters/src/directory.rs
+++ b/crates/parameters/src/directory.rs
@@ -7,7 +7,7 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_json::{error, from_str, from_value, to_string_pretty, to_value, Value};
 use tokio::fs::{read_to_string, write};
 
-use super::json::{copy_nested_value, merge_json, prune_equal_branches};
+use super::json::{clone_nested_value, merge_json, prune_equal_branches};
 
 #[derive(Debug, thiserror::Error)]
 pub enum DirectoryError {
@@ -53,7 +53,7 @@ where
     Parameters: DeserializeOwned,
 {
     let default_file_path = parameters_root_path.as_ref().join("default.json");
-    let mut parameters = from_path(default_file_path)
+    let mut parameters = read_from_file(default_file_path)
         .await
         .map_err(DirectoryError::DefaultParametersNotGet)?;
 
@@ -63,7 +63,7 @@ where
 
     let location_default_file_path = location_directory.join("default.json");
     if location_default_file_path.exists() {
-        let location_default_parameters = from_path(location_default_file_path)
+        let location_default_parameters = read_from_file(location_default_file_path)
             .await
             .map_err(DirectoryError::DefaultParametersOfLocationNotGet)?;
         merge_json(&mut parameters, &location_default_parameters);
@@ -73,7 +73,7 @@ where
         .as_ref()
         .join(format!("body.{}.json", body_id));
     if body_file_path.exists() {
-        let body_parameters = from_path(body_file_path)
+        let body_parameters = read_from_file(body_file_path)
             .await
             .map_err(DirectoryError::BodyParametersNotGet)?;
         merge_json(&mut parameters, &body_parameters);
@@ -83,7 +83,7 @@ where
         .as_ref()
         .join(format!("head.{}.json", head_id));
     if head_file_path.exists() {
-        let head_parameters = from_path(head_file_path)
+        let head_parameters = read_from_file(head_file_path)
             .await
             .map_err(DirectoryError::HeadParametersNotGet)?;
         merge_json(&mut parameters, &head_parameters);
@@ -91,7 +91,7 @@ where
 
     let location_body_file_path = location_directory.join(format!("body.{}.json", body_id));
     if location_body_file_path.exists() {
-        let location_body_parameters = from_path(location_body_file_path)
+        let location_body_parameters = read_from_file(location_body_file_path)
             .await
             .map_err(DirectoryError::BodyParametersOfLocationNotGet)?;
         merge_json(&mut parameters, &location_body_parameters);
@@ -99,7 +99,7 @@ where
 
     let location_head_file_path = location_directory.join(format!("head.{}.json", head_id));
     if location_head_file_path.exists() {
-        let location_head_parameters = from_path(location_head_file_path)
+        let location_head_parameters = read_from_file(location_head_file_path)
             .await
             .map_err(DirectoryError::HeadParametersOfLocationNotGet)?;
         merge_json(&mut parameters, &location_head_parameters);
@@ -133,13 +133,13 @@ where
 
     prune_equal_branches(&mut parameters, &stored_parameters);
 
-    let Some(sparse_parameters_from_scope_path) = copy_nested_value(&parameters, path) else {
+    let Some(sparse_parameters_from_scope_path) = clone_nested_value(&parameters, path) else {
         return Ok(());
     };
     let serialization_file_path =
         file_path_from_scope(scope, parameters_root_path, body_id, head_id);
     let mut parameters = if serialization_file_path.exists() {
-        from_path(&serialization_file_path)
+        read_from_file(&serialization_file_path)
             .await
             .map_err(DirectoryError::HeadParametersOfLocationNotGet)?
     } else {
@@ -147,7 +147,7 @@ where
     };
     merge_json(&mut parameters, &sparse_parameters_from_scope_path);
 
-    to_path(serialization_file_path, parameters)
+    write_to_file(serialization_file_path, parameters)
         .await
         .map_err(DirectoryError::HeadParametersOfLocationNotSet)
 }
@@ -177,20 +177,16 @@ fn file_path_from_scope(
     body_id: &str,
     head_id: &str,
 ) -> PathBuf {
-    let location_directory = parameters_root_path
-        .as_ref()
-        .join(location_directory_from_head_id(head_id));
-    match (scope.location, scope.id) {
-        (Location::All, Id::All) => parameters_root_path.as_ref().join("default.json"),
-        (Location::All, Id::Body) => parameters_root_path
+    let directory = match scope.location {
+        Location::All => parameters_root_path.as_ref().to_path_buf(),
+        Location::Current => parameters_root_path
             .as_ref()
-            .join(format!("body.{}.json", body_id)),
-        (Location::All, Id::Head) => parameters_root_path
-            .as_ref()
-            .join(format!("head.{}.json", head_id)),
-        (Location::Current, Id::All) => location_directory.join("default.json"),
-        (Location::Current, Id::Body) => location_directory.join(format!("body.{}.json", body_id)),
-        (Location::Current, Id::Head) => location_directory.join(format!("head.{}.json", head_id)),
+            .join(location_directory_from_head_id(head_id)),
+    };
+    match scope.id {
+        Id::All => directory.join("default.json"),
+        Id::Body => directory.join(format!("body.{}.json", body_id)),
+        Id::Head => directory.join(format!("head.{}.json", head_id)),
     }
 }
 
@@ -206,7 +202,7 @@ fn location_directory_from_head_id(head_id: &str) -> &'static str {
     }
 }
 
-async fn from_path(file_path: impl AsRef<Path>) -> Result<Value, SerializationError> {
+async fn read_from_file(file_path: impl AsRef<Path>) -> Result<Value, SerializationError> {
     let file_contents =
         read_to_string(&file_path)
             .await
@@ -220,7 +216,10 @@ async fn from_path(file_path: impl AsRef<Path>) -> Result<Value, SerializationEr
     })
 }
 
-async fn to_path(file_path: impl AsRef<Path>, value: Value) -> Result<(), SerializationError> {
+async fn write_to_file(
+    file_path: impl AsRef<Path>,
+    value: Value,
+) -> Result<(), SerializationError> {
     let file_contents =
         to_string_pretty(&value).map_err(|source| SerializationError::FileNotSerialized {
             source,
--- a/crates/parameters/src/directory.rs
+++ b/crates/parameters/src/directory.rs
@@ -3,12 +3,11 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use log::debug;
-use serde::{de::DeserializeOwned, Serialize};
-use serde_json::{error, from_str, from_value, to_value, Value};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use serde_json::{error, from_str, from_value, to_string_pretty, to_value, Value};
 use tokio::fs::{read_to_string, write};
 
-use super::json::{merge_json, prune_equal_branches};
+use super::json::{copy_nested_value, merge_json, prune_equal_branches};
 
 #[derive(Debug, thiserror::Error)]
 pub enum DirectoryError {
@@ -39,6 +38,8 @@ pub enum SerializationError {
     FileNotRead { source: io::Error, path: PathBuf },
     #[error("failed to parse {path:?}")]
     FileNotParsed { source: error::Error, path: PathBuf },
+    #[error("failed to serialize {path:?}")]
+    FileNotSerialized { source: error::Error, path: PathBuf },
     #[error("failed to write {path:?}")]
     FileNotWritten { source: io::Error, path: PathBuf },
 }
@@ -58,7 +59,7 @@ where
 
     let location_directory = parameters_root_path
         .as_ref()
-        .join(get_location_directory(head_id));
+        .join(location_directory_from_head_id(head_id));
 
     let location_default_file_path = location_directory.join("default.json");
     if location_default_file_path.exists() {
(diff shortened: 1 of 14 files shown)


tools/twix/src/repository_parameters.rs
```diff
@@ -32,46 +32,52 @@ impl RepositoryParameters {
...
-            merge_json(&mut stored_value, &nested_value_to_be_added);
+            let stored_value: Value = parameters::directory::deserialize(
+                repository.root_directory(),
+                &hardware_ids.body_id,
+                &hardware_ids.head_id,
+            )
+            .await
+            .unwrap_or_default();
+
+            let diff = get_diff_against_stored_value(&stored_value, &nested_value_to_be_added);
```
h3ndrk, line 52: What was the reason not to use `serialize()` from the configuration directory?

h3ndrk (changes requested):

tools/twix/src/repository_parameters.rs
tuxbotix: Serialise directly writes to the location directory, whereas we generally want to write a level above.
I didn't want to modify the behaviour of serialise either as it is a critical component for deployment, etc parts.

crates/repository/src/lib.rs
```diff
@@ -46,6 +46,10 @@ impl Repository {
         }
     }
 
+    pub fn root_directory(&self) -> PathBuf {
+        self.root.clone()
+    }
```
h3ndrk, line 51: You should be able return a `Path` to save the `.clone()`:

```suggestion
    pub fn root_directory(&self) -> Path {
        &self.root
    }
```

If the caller needs an owned value he/she can clone.

tools/twix/src/repository_parameters.rs
```diff
@@ -32,46 +32,61 @@ impl RepositoryParameters {
...
+            .hardware_ids_from_address(address)
         else {
             error!("failed to get head ID from address {address}");
             return
         };
         self.runtime.spawn(async move {
-            let mut stored_value = repository
-                .read_configuration(&head_id)
+            let stored_complete_parameter_tree: Value = parameters::directory::deserialize(
+                repository.root_directory(),
```
h3ndrk, line 43: Is this root directory correct? Shouldn't this be

```suggestion
                repository.root_directory().join("etc/configuration"),
```

?

tools/twix/src/repository_parameters.rs
```diff
@@ -80,7 +95,7 @@ fn last_octet_from_ip_address(ip_address: Ipv4Addr) -> u8 {
 }
 
 // Create tree structure from path and value points to the last key i.e. a.b.c -> { a: { b: { c: value } } }
-fn nest_value_at_path(path: &str, value: &Value) -> Value {
+fn make_sparse_value_tree_from_path(path: &str, value: &Value) -> Value {
```
h3ndrk, line 98: What was the reasoning to change the name?

tools/twix/src/repository_parameters.rs
```diff
@@ -32,46 +32,61 @@ impl RepositoryParameters {
...
-                .read_configuration(&head_id)
+            let stored_complete_parameter_tree: Value = parameters::directory::deserialize(
+                repository.root_directory(),
+                &hardware_ids.body_id,
+                &hardware_ids.head_id,
+            )
+            .await
+            .unwrap_or_default();
+
+            // value is just the "leaf" of the "path", make the tree structure to diff/ merge against complete parameter tree.
```
h3ndrk, line 50: This comment explains ***what*** happens, which the code already expresses. I propose to either remove the comment to not violate DRY or explain the ***why***.

tools/twix/src/repository_parameters.rs
```diff
@@ -32,46 +32,61 @@ impl RepositoryParameters {
...
     }
 
-    fn head_id_from_nao_number(&self, nao_number: u8) -> Result<String> {
+    fn hardware_ids_from_nao_number(&self, nao_number: u8) -> Result<HardwareIds> {
         self.ids
             .get(&nao_number)
             .ok_or_else(|| eyre!("no IDs known for NAO number {nao_number}"))
-            .map(|id| id.head_id.clone())
+            .cloned()
     }
```
h3ndrk, line 90: Same here: You could return a borrowed value and remove the `.cloned()`. The remaining code only takes references from your newly cloned `HardwareIds` instance. Therefore it does not matter if you already return a borrow.

tools/twix/src/repository_parameters.rs
```diff
@@ -90,11 +105,20 @@ fn nest_value_at_path(path: &str, value: &Value) -> Value {
         })
 }
 
+fn get_diff_against_stored_value(stored_value: &Value, incoming_sparse_value: &Value) -> Value {
+    let mut diff = stored_value.clone();
+    merge_json(&mut diff, incoming_sparse_value);
+    prune_equal_branches(&mut diff, stored_value);
+    diff
+}
```
h3ndrk, line 113: Why not take `stored_value` as owned value and make it `mut` inside this function. No need to clone. The borrowed value is not used anywhere else so you should be able to move it in this function.

```suggestion
fn get_diff_against_stored_value(mut stored_value: Value, incoming_sparse_value: &Value) -> Value {
    merge_json(&mut stored_value, incoming_sparse_value);
    prune_equal_branches(&mut stored_value, stored_value);
    stored_value
}
```

h3ndrk (changes requested):

tools/twix/src/repository_parameters.rs
h3ndrk: We need more functionality in the deserialization to select where to deserialize to (location, no location, head/body, ...). The communication interface would also benefit from this feature. @knoellle Do we want to make the serialization more intelligent in this PR?

tuxbotix: Yes this would remove some code duplication.

tools/twix/src/repository_parameters.rs
tuxbotix: I'll need to double check, as my memory serves deserialize takes the repo root.

h3ndrk: https://github.com/HULKs/hulk/blob/main/crates/communication/src/server/parameters/directory.rs#L54

tools/twix/src/repository_parameters.rs
knoellle: I think that's a good idea, yes.
Parameter reading/writing should be implemented only once and then used both for tools running on dev machines as well as for code running on the nao.
I had already suggested unifying them in a previous PR during GORE but decided to postpone it.

crates/repository/src/lib.rs
tuxbotix: This function is removed as it is no longer needed.

tools/twix/src/repository_parameters.rs
tuxbotix: Damnit! you are right, I mixed up the two as repository crate and this have different meanings.
I renamed `root_path` to `parameters_root_path` for directory.rs. Hope it is not a problem.

I made the existing `configuration_root` method of repository crate public and used it.

tools/twix/src/repository_parameters.rs
tuxbotix: With the merge and prune operations done, it felt clearer to refer this as a sparse tree, but in retrospection I think this might not convey what this parameter does better than the old one.
What's your opinion? I can revert if you think the same.

tools/twix/src/repository_parameters.rs
tuxbotix: Makes sense, the variable naming could be clearer if needed instead of that comment.

tools/twix/src/repository_parameters.rs
tuxbotix: I'll still have to add a `.cloned()` in `hardware_ids_from_address` as that may return a temporary;
```rust
    fn hardware_ids_from_address(&self, address: &str) -> Result<HardwareIds> {
        if address == "localhost" {
            return Ok(HardwareIds { // temporary
                body_id: "webots".to_string(),
                head_id: "webots".to_string(),
            });
        }
        let nao_number =
            last_octet_from_ip_address(address.parse().wrap_err("failed to parse IP address")?);
        self.hardware_ids_from_nao_number(nao_number)
            .wrap_err_with(|| format!("failed to get head ID from NAO number {nao_number}"))
            .cloned() // clone
    }
```

tools/twix/src/repository_parameters.rs
tuxbotix: Unfortunately, this wouldn't work as I need the original stored value to perform the pruning `prune_equal_branches(&mut stored_value, stored_value);`.

tools/twix/src/repository_parameters.rs
tuxbotix: In that case, we can implement these functions into this PR and get it done.

My proposal for this:
- Decouple serialize function and write to the disk.
- Or indicate where to write by passing an argument to serialize

If this sounds reasonable, I can implement one of the solutions.

tuxbotix:
Moved back to "In Progress" to reduce code duplication and possibly move the extra functionality in `twix/repository_parameters` into `parameters` crate.

crates/repository/src/lib.rs
h3ndrk: Your PR moves the function and adds a `pub`. I would vote for not moving it because I see no reason for it.

tools/twix/src/repository_parameters.rs
h3ndrk: Interestingly, during the refactorings back then to your that I did, I once renamed it from `make_sparse_value_tree_from_path()` to `nest_value_at_path()`. You are now changing it back. :joy:

What I find suboptimal with the *make*-function is that it does not describe how the value tree is created. I thought that the *nest*-function better describes that you're just putting this value at a specific depth and the tree nodes are named after path segments. :thinking:

tools/twix/src/repository_parameters.rs
h3ndrk: Yes, I overlooked this. Then it needs to stay as is.

tools/twix/src/repository_parameters.rs
h3ndrk: Ah, I see. Mist.

tools/twix/src/repository_parameters.rs
knoellle: I'm not sure what you mean by decoupling serialization from writing.
The actual serialization part is trivial and does not need to be extracted. On the other hand, merging of an updated branch or leaf with the current configuration and generating a diff already needs information about where the configuration will be saved to.

Thus, I argue for the second option.

h3ndrk: The second option also allows easy integration in the communication protocol.

h3ndrk: My current idea is to change the parameter directory API and make the API smarter by hiding complexity. Currently the serialization/deserialization are free functions and the idea is to store the parsed parameters with location information as state and convert the free functions into methods on this state. The goal of this design is to hide the location complexity inside of this new type.

The deserialization will be converted into a method that returns the flattened parameters (e.g. for usage in Communication).

For the serialization it will be possible to decide whether to add or remove a value at a path. Additions must contain the location (all locations, all locations with head ID, all locations with body ID, current location, current location with head ID, current location with body ID). Removals also must contain the location (it should also be possible to pass "don't care" which removes any value in any parameter location). With this design, we can more easily update parameter files and pass some hard decisions to the user (e.g. where to store the parameters).

This adjusted parameter directory API will be made available via Communication API und used in this PR for Twix parameter saving.

One drawback of this design is that the user can only get a flattened snapshot of the parameters but needs to provide location information while updating values. This may lead to situations where the user stores a value but a higher priority value at the same path still takes precedence resulting in no change in the flattened parameters. One mitigation for the user in this scenario is to first remove the value in all locations and then add it.

Open questions:

1. Only allow to store to the current location or to arbitrary locations?
2. Have a flat enum for the location decision or divide into multiple orthogonal enums e.g. whether it should be in a location and whether it should be default/head/body?

knoellle: I generally like the approach.

Regarding your questions:
1. Which "location" do you mean here? The term is overloaded by the physical location as well as the combination of physical location + head/body ID for the filed location.
2. I'd argue for having separate enums. Since they are - as you say - orthogonal.

I also have some questions of my own. I will continue your numbering to avoid confusion.

3. On the communication server side, would you keep this state in addition to what the server already stores or did you intend to replace it? In the former case the new state object would be in sync with the state on disk. That would have to advantage of being able to save parameters regardless of what has been temporarily set at runtime by a communication client.
4. In the "remove: don't care" case mentioned above, what do you mean by "any value in any location"? 
   1. Why "any value"? Would that delete the whole file at the specified location/all locations?
   2. "any location" meaning the `Location` enum above or across all physical locations?

h3ndrk: 1. I meant the combination of location and IDs. The question is if we want to limit the user that it will be all or the current location (regardless of IDs) or we allow that it can be all or a given location (e.g. by string, regardless of IDs).
2. ACK
3. We currently have this buffering semantic and we can keep it. The buffering would be in the form of the state within the new type.
4. The idea with "any" is that we delete all values (and children) at a given path in all combinations of locations and IDs. This would be for user convenience that the user does not need to trigger removal calls for each combination but rather the server iterates and deletes everything under a given path for all combinations.
5. I just noticed in 4. that with the removal feature, we could produce invalid configurations because we don't "default" the value at the path but completely nuke it. If there is a bug and we are not adding something back, this will be a problem. I'm not sure how to mitigate this. 🤔

h3ndrk: The whole idea of an incremental update API will only work for the Twix case because there we are directly writing JSON values into the JSON object. But for communication, this will not work because there we are using `SerializeHierarchy`. `SerializeHierarchy` works on the flattened JSON object and has no knowledge about locations and IDs. The proposed parameter directory API does not help with that because we need to somehow recover location information.

New idea similar what Darshana already started: We could diff the stored parameters on disk with the flattened parameters in RAM. This diff is then stored to the combination of location and ID that the user requests.

One scenario could be that we manually modify the calibration parameters via Twix in the Communication server. Afterwards we are storing the parameters to disk, i.e., the diff results in the calibration parameters and the user specified to store this diff in the root head parameters. Should work.

Another scenario could be that we tweak some vision parameters and also the calibration parameters. We would like to store the vision parameters in the root parameters for all robots but the calibration parameters only for the particular robot. A solution would be to let the user specify the path to store to disk in addition to the combination of location and IDs.

Btw, let's call the combination of location and IDs from now on *scope*.

This would result in the following API:

```rust
pub async fn deserialize<Parameters>(
    parameters_root_path: impl AsRef<Path>,
    body_id: &str,
    head_id: &str,
) -> Result<Parameters, DirectoryError>
where
    Parameters: DeserializeOwned,
{}

pub async fn serialize<Parameters>(
    parameters: &Parameters,
    scope: Scope,  // <-------------------------- new
    parameters_root_path: impl AsRef<Path>,
    body_id: &str,
    head_id: &str,
) -> Result<(), DirectoryError>
where
    Parameters: Serialize,
{}

struct Scope {
    location: Location,
    id: Id,
    path: String, // <----- path in the JSON diff object to store in this serialization
}                 //        will be similar to the path used in SerializeHierarchy
enum Location { All, Current }
enum Id { All, Body, Head }
```

h3ndrk: Works as planned: https://github.com/HULKs/hulk/pull/267/commits/9a851f668f3d9e7efa1c5f5cb74547177591a1cd

crates/parameters/src/json.rs
```diff
@@ -35,9 +35,32 @@ pub fn prune_equal_branches(own: &mut Value, other: &Value) {
     }
 }
 
+pub fn copy_nested_value(value: &Value, path: &str) -> Option<Value> {
```
knoellle, line 38: `clone_nested_value`?

knoellle (changes requested):

tools/twix/src/repository_parameters.rs
h3ndrk: You are correct. Not done yet.

h3ndrk: Now integrated into Twix and repository crate.

crates/parameters/src/directory.rs
```diff
@@ -160,8 +221,11 @@ async fn from_path(file_path: impl AsRef<Path>) -> Result<Value, SerializationEr
 }
 
 async fn to_path(file_path: impl AsRef<Path>, value: Value) -> Result<(), SerializationError> {
```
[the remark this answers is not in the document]
h3ndrk: Renamed

crates/parameters/src/directory.rs
```diff
@@ -103,36 +110,91 @@ where
...
+        (Location::All, Id::Body) => parameters_root_path
+            .as_ref()
+            .join(format!("body.{}.json", body_id)),
+        (Location::All, Id::Head) => parameters_root_path
+            .as_ref()
+            .join(format!("head.{}.json", head_id)),
+        (Location::Current, Id::All) => location_directory.join("default.json"),
+        (Location::Current, Id::Body) => location_directory.join(format!("body.{}.json", body_id)),
+        (Location::Current, Id::Head) => location_directory.join(format!("head.{}.json", head_id)),
+    }
```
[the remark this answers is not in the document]
h3ndrk: Orthogonalized

crates/parameters/src/json.rs
h3ndrk: Renamed

knoellle (approved):
LGTM

PR #278, opened 2023-04-28
Decay the ball velocity in the ball filter

### LINKED ISSUE #273: Increase ball model roll resistance
We observed the real ball slowing down much faster than the ball model, causing robots to look at the wrong place, losing track of the ball.

jonathan-hellwig:
## Introduced Changes
During the games we had issues with the ball position, because the prediction step in the ball filter did not account for friction. In this PR, a decay factor for the velocity is introduced to model the friction of the ball.

Fixes #273 

## ToDo / Known Issues
- the velocity decay parameter has not been tested on the field

None

## How to Test

Deploy this branch to a robot and check the ball behavior through the map panel in Twix.


knoellle (approved):
LGTM

PR #284, opened 2023-04-29
Add instant kick decisions to own database to prevent action panic

MaikRe:
## Introduced Changes

As in title, instant kick decisions were not output to own_database in behavior simulator.

## How to Test

Behavior sim works again as current main does not.


schmidma (approved):
good catch, sorry for introducing that bug

PR #293, opened 2023-04-29
Increase ready phase length 

### LINKED ISSUE #214: Lua Motion Files
Quick brainfart.
Not sure if this is a good idea, investigate.

While thinking about resetting the standup motion when falling during standup, I had the idea of using lua for motion files.
That would make it easier to implement resets, waiting conditions and different types of interpolators without having to design an overly complex format covering all of these usecases and state machines.

MaikRe:
## Introduced Changes

Change ready phase length to closer match real time while keeping it short for faster testing

Fixes #214 

Create some lua helper functions used in multiple files to prevent having to repeat function creation and allow for easier changes in behavior testing across multiple scenarios. Configs should be use for certain parameter changes.

## How to Test

Run behavior sim with the changed files, robots now have time to reach proper positions before Set state is active as the lookaround motion caused this to not be the case.


h3ndrk:
Is this still In Progress?

tests/behavior/ingamepenalty_attacking.lua
```diff
@@ -20,7 +20,7 @@ function on_goal()
   print("Ball: " .. inspect(state.ball))
   print("Ball was at x: " .. state.ball.position[1] .. " y: " .. state.ball.position[2])
   state.ball = nil
-  game_end_time = state.cycle_count + 200
+  state.finished = true
```
knoellle, line 23: While this fixes the crash, I would prefer if we simulated the game controller correctly.
So instead of aborting the simulation immediately prior to the crash, the game controller state should be set to ready after a goal was scored.

MaikRe: Done, next iteration this should be moved into a helper function to be shared across lua files

knoellle (approved):
LGTM

PR #306, opened 2023-04-29
Abort StandUp if Fallen

Vivituhh:
## Introduced Changes

Abort stand up motion when robot falls again in the stand up motion and start again the right motion.
This required some changes to the motion_selector since a repeated dispatch is necessary in the transition from StandUpBack to StandUpBack or StandUpFront to StandUpFront if the robot falls.

- Instead of checking if the robot is fallen only in discrete time points, maybe it is better to have a concept of milestones in the motionfile. In this context after reaching the FallenAbort milestone, the robot will always abort the motion once it falls instead of waiting for the next condition.

## How to Test

Test on robot: In stand up motion make it fall again forwards/backwards, motion should be aborted as soon as it runs over an abort condition (takes some frames).


crates/control/src/motion/motion_selector.rs
```diff
@@ -103,15 +104,17 @@ fn motion_type_from_command(
 fn transition_motion(
     from: MotionType,
     to: MotionType,
-    is_safe_to_exit: bool,
+    motion_finished: bool,
```
h3ndrk, line 107: What was your reasoning for the renaming?

crates/control/src/motion/stand_up_back.rs
```diff
@@ -45,53 +37,41 @@ pub struct MainOutputs {
...
+        self.interpolator.value()
+    }
+
+    pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+        let current_position =
+            if let MotionType::StandUpBack = context.motion_selection.current_motion {
+                self.compute_stand_up_back(context)
+            } else {
+                self.interpolator.reset();
+                dbg!("initial positions");
```
h3ndrk, line 70: Remove this `dbg!()`?

crates/control/src/motion/stand_up_back.rs
```diff
@@ -45,53 +37,41 @@ pub struct MainOutputs {
...
-                        < Vector2::new(
-                            *context.gyro_low_pass_filter_tolerance,
-                            *context.gyro_low_pass_filter_tolerance,
-                        )
-                    {
-                        context.motion_safe_exits[MotionType::StandUpBack] = true;
-                    }
-                }
-            };
+            dbg!("interpolator finished");
```
h3ndrk, line 57: Remove this `dbg!()`?

crates/control/src/motion/stand_up_back.rs
```diff
@@ -45,53 +37,41 @@ pub struct MainOutputs {
...
+
+    pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+        let current_position =
+            if let MotionType::StandUpBack = context.motion_selection.current_motion {
+                self.compute_stand_up_back(context)
+            } else {
+                self.interpolator.reset();
+                dbg!("initial positions");
+                self.interpolator.initial_positions()
+            };
```
h3ndrk, line 72: Why not always return the current interpolator value instead of the initial positions? If the interpolator is reset and not advanced further, it should return the initial positions, right?

crates/control/src/motion/stand_up_back.rs
```diff
@@ -45,53 +37,41 @@ pub struct MainOutputs {
...
+    }
+
+    pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+        let current_position =
+            if let MotionType::StandUpBack = context.motion_selection.current_motion {
+                self.compute_stand_up_back(context)
+            } else {
+                self.interpolator.reset();
+                dbg!("initial positions");
+                self.interpolator.initial_positions()
```
h3ndrk, line 71: Why does the interpolator have a "initial positions" getter? Shouldn't the interpolator be generic over the type it interpolates? Having a getter with "positions" in the name means this interpolator can only interpolate over positions?

etc/motions/stand_up_back.json
```diff
@@ -1,3 +1,4 @@
+
```
h3ndrk, line 1: Remove empty line?

crates/motionfile/src/condition.rs
```diff
@@ -1,18 +1,25 @@
...
 use types::ConditionInput;
 
+pub enum Response {
+    Abort,
+    Continue,
+    Wait,
+}
+
 #[enum_dispatch(ConditionType)]
 pub trait Condition: Clone {
```
h3ndrk, line 16: Do you need to require `Clone` here?

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -38,6 +39,16 @@ enum State {
         time_since_start: Duration,
     },
     Finished,
+    Aborted {
+        frame_side: Side,
```
h3ndrk, line 43: Is it necessary to remember the `frame_side`? I don't know the details, please explain shortly. Thanks!

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -57,21 +68,23 @@ impl<T: Debug + Interpolate<f32>> MotionInterpolator<T> {
                 time_since_start,
             } => {
                 let current_frame = &self.frames[current_frame_index];
-                if current_frame
+                match current_frame
```
h3ndrk, line 71: I'm looking at the motion file conditions for the first time. If I understood it correctly, you are checking the conditions before entering the next frame. Couldn't you model it in the way that the conditions are checked in each advancement step. This decoupling from the frames would allow to immediately abort or wait in the middle of frames which sounds more powerful to me. What was your reasoning? :thinking:

etc/motions/stand_up_back_dortmund_2022.json
```diff
@@ -408,7 +410,14 @@
...
             }
           }
-        },
+        }
+      ]
+    },
+    {
+      "entry_condition": {
+        "FallenAbort": {}
+      },
```
h3ndrk, line 419: Does this work?

```suggestion
      "entry_condition": "FallenAbort",
```

old_stand_up.json
h3ndrk, line 1: Empty file?!

h3ndrk (changes requested):
Some understanding questions and remarks. :sweat_smile:

crates/control/src/motion/motion_selector.rs
okiwi6: The meaning of the variable changed. In the standUpBack, standUpFront motions this will now also be set when the motion is aborted. This allows for a repeated dispatch to the initial_positions of the standup motion.

crates/control/src/motion/stand_up_back.rs
okiwi6: In this case `interpolator.value() == interpolator.initial_positions()`, however I find the latter more intuitive to read.

crates/control/src/motion/stand_up_back.rs
okiwi6: Agree. What do you propose? Sth. like `initial_value`?

crates/control/src/motion/stand_up_back.rs
okiwi6: Oops, removed

crates/control/src/motion/stand_up_back.rs
okiwi6: Oops, removed

crates/motionfile/src/motion_interpolator.rs
okiwi6: Yes, this was the origin of the dying robot bug. In case the motion is aborted, the interpolator will be asked for joint angles one last time. In this case the correct joint angles have to be calculated. If the entry condition failed, the initial values of a segment are correct, if the exit condition failed the end values of a segment need to be output. Alternatively `Aborted` could save the `Joints` directly, however this would make the enum variant very large.

crates/motionfile/src/motion_interpolator.rs
okiwi6: Well this sort of depends on the `Condition`. For the new `FallenAbort`, this would be much better, since we can save some time again by restarting the motion earlier. However for other conditions like `Stabilized` which waits until the angular velocity is small it only makes sense to do this at distinct time points (otherwise no dynamic motions would be possible).

etc/motions/stand_up_back_dortmund_2022.json
okiwi6: No, unfortunately not:
`invalid type: unit variant, expected newtype variant at line 417 column 39`

old_stand_up.json
okiwi6: Hehe, this should not have been commited :sweat_smile:, removed now

crates/motionfile/src/condition.rs
okiwi6: I thought so, but apparently it works without it. Removed now.

crates/control/src/motion/motion_selector.rs
h3ndrk: But shouldn't the motion also be safe to exit on abort? The motion is not finished when it is aborted. Therefore I'm wondering whether the renaming is better. :thinking:

crates/control/src/motion/stand_up_back.rs
h3ndrk: Hmmm, I find it more readable to always return the current positions because there is a single source/responsibility where the angles are coming from and you can also remove your `initial_positions()` getter because it would be dead code. One could argue that currently you are accessing/leaking internal implementation detail of the interpolator. :thinking:

crates/control/src/motion/stand_up_back.rs
h3ndrk: Would be better but let's first agree on the discussion above.

crates/motionfile/src/motion_interpolator.rs
h3ndrk: Maybe we want to have two classes of conditions? Conditions that are evaluated in each cycle of a frame and conditions that are a special kind of frame which can wait for a condition. Try to challenge the design and see whether it can be made more general s.t. you have more clear responsibilities and abstractions. We could also move this to Ideas for Next Iterations.

crates/control/src/motion/stand_up_back.rs
okiwi6: Removed

crates/motionfile/src/motion_interpolator.rs
okiwi6: I will experiment a bit with this

okiwi6: I've now implemented sth. like that and it works great. However the state transitions in the MotionInterpolator are very messy (and it doesn't even yet correct automatically for non-existent conditions). What do you think of how I implemented the second condition type and do you know a way to make the state transitions cleaner?

crates/control/src/motion/stand_up_back.rs
h3ndrk: Looks way better, thanks! I think, we can improve even more: If we would extract the `interpolator.value()` from the `compute_...()` function and rename the function to e.g. `advance_interpolator()` or something else, we could move the `interpolator.value()` out of the if-condition in the `cycle()` function. This results in code that only has a single `interpolator.value()` and is clearly separated between 1. mutating the interpolator and 2. getting the value and returning it.

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -12,19 +13,20 @@ use types::ConditionInput;
 
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct ConditionedSpline<T> {
-    pub entry_condition: Option<ConditionType>,
+    pub entry_condition: Option<DiscreteConditionType>,
+    pub enable: Option<Vec<ContinuousConditionType>>,
```
h3ndrk, line 17: Instead of `enable` maybe `spline_conditions` or similar? IMHO `enable` is more confusing than helpful.

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -51,27 +56,54 @@ impl Default for State {
...
+                    condition
+                        .evaluate(condition_input)
+                        .with_timeout(condition.timeout(time_since_start))
+                }) {
+                    Some(Response::Abort) => State::Aborted {
+                        at_position: self.value(),
+                    },
+                    Some(Response::Wait) => State::CheckEntry {
                         current_frame_index,
                         time_since_start: time_since_start + time_step,
```
h3ndrk, line 101: Why are you advancing the time in the wait state? Maybe handle it similar to below?

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -12,19 +13,20 @@ use types::ConditionInput;
 
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct ConditionedSpline<T> {
-    pub entry_condition: Option<ConditionType>,
+    pub entry_condition: Option<DiscreteConditionType>,
+    pub enable: Option<Vec<ContinuousConditionType>>,
     pub spline: TimedSpline<T>,
-    pub exit_condition: Option<ConditionType>,
+    pub exit_condition: Option<DiscreteConditionType>,
 }
```
h3ndrk, line 20: Why can there be multiple `enable` conditions but only a single `entry_` or `exit_` condition?

Also why is it an `Option<Vec<...>>`? Why is `None` not just an empty vector? Or does `Option` allow you to omit the key in the JSON file? Then just use vectors and annotate it as serde-skip-field?

crates/motionfile/src/motion_interpolator.rs
h3ndrk: The `advance_by()` function is already very long and complex. It also violates the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle). It does too many low-level things. One mitigation would be to split it into multiple functions that handle different parts of the function. That would also allow to reuse code and reduce code duplication.

Other than that the state transitions are probably needed if you want to support your aimed features (aka your requirements) where each condition may have different outcomes. That requires you to react to all outcomes. Maybe you could extract the timeout-functionality and put it into a more central place s.t. you can also check it at a more central place. Currently each condition can implement special functionality which is maybe not needed and would make your code simpler.

crates/control/src/motion/stand_up_back.rs
okiwi6: Done

crates/motionfile/src/motion_interpolator.rs
okiwi6: I've split up the condition checking and state transitions now

crates/motionfile/src/motion_interpolator.rs
```diff
@@ -51,27 +56,54 @@ impl Default for State {
...
+                ..
+            }
+            | State::CheckExit {
+                current_frame_index,
+                ..
+            } => {
+                let current_frame = &self.frames[current_frame_index];
+                if let Some(continuous_conditions) = current_frame.enable.as_ref() {
+                    if continuous_conditions.iter().any(|condition| {
+                        matches!(condition.evaluate(condition_input), Response::Abort)
```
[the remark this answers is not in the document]
okiwi6: Made things easier, but certainly not correct. Now should be implemented correctly

crates/motionfile/src/motion_interpolator.rs
okiwi6: I don't know what you mean with "below". The time advancement is required for the timeout functionality of conditions

crates/motionfile/src/motion_interpolator.rs
okiwi6: `None` allows to omit the key in the json. I don't understand what you mean by `serde(skip)`. If I used that, I wouldnt be able to (de-)serialize that attribute anymore?

okiwi6: About `Vec`: I thought it wouldn`t be too hard to implement this for enhanced functionality. However, you are right that it isn't used at the moment. Do you think I should drop `Vec` support?

crates/motionfile/src/motion_interpolator.rs
okiwi6: What do you think of `motion_interrupts`? Since they can occur at any time in a segment to abort/pause the animation.

h3ndrk (approved):

PR #316, opened 2023-04-30
Lightning fast walking

schmidma:
## Introduced Changes

improve stability and fix a bug in step size computation


JonasHolz (approved):
lgtm

PR #321, opened 2023-04-30
Add `pepsi aliveness --watch`

Narcha:
## Introduced Changes

Add a `--watch` parameter to continually refresh the output of `pepsi aliveness`.
The refresh interval is configurable with `--interval` | `-i`.

For consistency, the `--timeout` parameter is changed to be in seconds instead of milliseconds.

## ToDo / Known Issues

None

None.

## How to Test

Run `./pepsi aliveness --watch`, try the `--interval` parameter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ constants = { path = "crates/constants" }
 context_attribute = { path = "crates/context_attribute" }
 control = { path = "crates/control" }
 convert_case = "0.6.0"
+crossterm = { git = "https://github.com/crossterm-rs/crossterm.git", rev = "03c25178af94ae986677dfc178f94c931b96de66", default-features = false }
 ctrlc = { version = "3.2.3", features = ["termination"] }
 cyclers = { path = "crates/cyclers" }
 eframe = { version = "0.21.0", features = ["persistence"] }
--- a/tools/pepsi/src/aliveness.rs
+++ b/tools/pepsi/src/aliveness.rs
@@ -1,7 +1,13 @@
-use std::{collections::BTreeMap, net::IpAddr, num::ParseIntError, time::Duration};
+use std::{collections::BTreeMap, io::stdout, net::IpAddr, num::ParseFloatError, time::Duration};
 
 use clap::{arg, Args};
-use color_eyre::owo_colors::OwoColorize;
+use color_eyre::Result;
+use crossterm::{
+    cursor,
+    style::{PrintStyledContent, Stylize},
+    terminal::{Clear, ClearType},
+    QueueableCommand,
+};
 
 use crate::parsers::NaoAddress;
 use aliveness::{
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -63,6 +63,7 @@ constants = { path = "crates/constants" }
 context_attribute = { path = "crates/context_attribute" }
 control = { path = "crates/control" }
 convert_case = "0.6.0"
+crossterm = { git = "https://github.com/crossterm-rs/crossterm.git", rev = "03c25178af94ae986677dfc178f94c931b96de66", default-features = false }
```
h3ndrk, line 66: Why do we need this special version?

tools/pepsi/src/aliveness.rs
```diff
@@ -19,16 +25,22 @@ pub struct Arguments {
     /// Output aliveness information as json
     #[arg(long, short = 'j')]
     json: bool,
-    /// Timeout in ms for waiting for responses
-    #[arg(long, short = 't', value_parser = parse_duration, default_value = "200")]
+    /// Timeout in seconds for waiting for responses
+    #[arg(long, short = 't', value_parser = parse_duration, default_value = "0.2")]
     timeout: Duration,
+    /// Refresh output every 2 seconds
```
h3ndrk, line 31: This help text is not correct. The interval is configurable, not 2 seconds.

tools/pepsi/src/aliveness.rs
```diff
@@ -19,16 +25,22 @@ pub struct Arguments {
...
+    #[arg(long, short = 'w')]
+    watch: bool,
+    /// Refresh interval for --watch, in seconds
+    #[arg(long, short = 'i', value_parser = parse_duration, default_value = "1.0")]
```
h3ndrk, line 35: Make it shorter?

```suggestion
    #[arg(long, short = 'i', value_parser = parse_duration, default_value = "1")]
```

tools/pepsi/src/aliveness.rs
```diff
@@ -41,8 +53,32 @@ pub enum Error {
...
+        loop {
+            stdout
+                .queue(Clear(ClearType::All))?
+                .queue(cursor::MoveTo(0, 0))?
+                .queue(PrintStyledContent(
+                    format!("Aliveness (every {:?})", arguments.interval).dark_grey(),
+                ))?
+                .queue(cursor::MoveTo(0, 1))?;
+            query_and_print_aliveness(&arguments).await?;
+            tokio::time::sleep(arguments.interval).await;
```
h3ndrk, line 71: Replace qualified path with use?

```suggestion
            sleep(arguments.interval).await;
```

tools/pepsi/src/aliveness.rs
```diff
@@ -41,8 +53,32 @@ pub enum Error {
...
+            .queue(cursor::MoveTo(0, 0))?;
+        loop {
+            stdout
+                .queue(Clear(ClearType::All))?
+                .queue(cursor::MoveTo(0, 0))?
+                .queue(PrintStyledContent(
+                    format!("Aliveness (every {:?})", arguments.interval).dark_grey(),
+                ))?
+                .queue(cursor::MoveTo(0, 1))?;
+            query_and_print_aliveness(&arguments).await?;
```
h3ndrk, line 70: This "scrolls" my whole terminal screen for each iteration. Isn't there a mode that all other modal terminal applications use e.g. nano, vim, or htop? In my terminal, the whole terminal is used for the application and at teardown, my whole prompt is restored from where I started the terminal application. Would be cool if our aliveness would feel more native. Is that possible?

tools/pepsi/src/aliveness.rs
```diff
@@ -41,8 +53,32 @@ pub enum Error {
...
 }
 
-pub async fn aliveness(arguments: Arguments) -> Result<(), Error> {
-    let states = query_aliveness_list(&arguments)
+pub async fn aliveness(arguments: Arguments) -> Result<()> {
+    if arguments.watch {
+        let mut stdout = stdout();
+        stdout
+            .queue(Clear(ClearType::All))?
+            .queue(cursor::MoveTo(0, 0))?;
```
h3ndrk, line 61: You are clearing in each iteration, why do you also clear here?

h3ndrk (changes requested):
Tested with a real NAO, seems to work but I would really love to see improvements in the raw terminal mode.

PR #323, opened 2023-05-01
Home Sweet Home

schmidma:
## Introduced Changes

back at SMD -> Sets the nao location to SMD


knoellle (approved):
LGTM

etc/configuration/nao_location
```diff
@@ -1 +1 @@
-gore/
```
h3ndrk, line 1: Delete GORE parameters?

h3ndrk (changes requested):

h3ndrk (approved):

PR #326, opened 2023-05-05
Interpolation Mode and Name

okiwi6:
## Introduced Changes
This PR adds two optional fields to the motionfile json.
* In the top scope it is now possible to set the interpolation mode to something else than linear
* In motion segments it is now possible to set a name for the current segment. Since there currently is no visualization of motionfiles this greatly helps remembering what different parts do. The value is not used anywhere in the rust code, however it will be used in future tooling regarding editing motionfiles.

Fixes #

## ToDo / Known Issues

## How to Test

In the top scove of the motionfile, set the `interpolator_mode` key to e.g. `catmullrom`, the motion will be using this interpolation mode instead then.
In a motion segment, the `name` key can be set to any arbitrary string.


knoellle (approved):
LGTM

PR #330, opened 2023-05-06
Nuke the Flower Power

schmidma:


knoellle (changes requested):
While you're at it, please also delete `tools/sprite` - the old behavior simulator interface

knoellle (approved):
LGTM

Issue #15, opened 2022-11-30
Post Open Source Release TODOs


h3ndrk:
- [x] SSH key removal
- [x] SSH key automatic login @schmidma
- [x] Make meta-hulks public (and the Rattenschwanz that is pulled by that)
- [x] All links to the old `HULKsCodeRelease` repo need to be updated to point to `hulk`
  - [x] Internal pages, Readmes
  - [x] hulks.de
  - [x] External RoboCup, pages etc.
- [x] Merge queue cannot be activated because `hulk` is not part of the closed beta. It will be generally available in Q4 2022, see https://github.com/github/roadmap/issues/370 **Edit: It is now available again for this repository and has been activated.**
- [x] CI Checks need to be added to branch protection rule once the first check had run
- [x] Migrate issues: **We are moving issues over if we see fit.**
- [x] Migrate PRs: **We are moving and rebasing PRs over if we see fit.**
- [x] Migrate labels
- [x] Work out a policy/workflow how to resubmit PRs to the new repository
  - Add `hulk` remote in your repository
  - Fetch it
  - Checkout the `main` branch of the `hulk` remote
  - Create the new branch that will contain the "rebased" changes
  - Then cherry-pick the individual commits e.g. `git cherry-pick beginning_hash..ending_hash` where `beginning_hash` and `ending_hash` are commit IDs. Careful, the `beginning_hash` has to be the base of your branch, not the first commit on top of that.
  - Push your new branch
  - Open a PR
  - Profit
- [x] readme title #331
- [x] Kanban board https://github.com/orgs/HULKs/projects/2
- [x] Was ist das? https://github.com/HULKs/hulk/blob/3e6e14baf99bf9b4f806a8787c117e7a8041513b/.github/workflows/documentation.yml#L22
  - is automatically filled by GitHub Actions with a token that has write access
- [x] Update Documentation hulks.de/hulk
- [x] Decide what to do with the wiki: The important stuff has been migrated
- [x] archive hulks/nao
- [ ] CI security considerations
  - Are there any keys/secrets accessible from inside the docker containers?
  - Do the containers have access to any system resources that could be a problem?

Bendodroid:
Labels have been migrated.

knoellle:
The only remaining point on this list are the security considerations.
I don't think our CI runners contain any secrets worth stealing and since I doubt anyone is going to slap a stamp of "yes, this is secure" on it, I suggest closing this issue.

h3ndrk:
[...]

PR #331, opened 2023-05-06
Fix title and remove copyright year

h3ndrk:
## Introduced Changes

What it says, originated from #15. The copyright year has been incremented.

## ToDo / Known Issues

None

None

## How to Test

Check the README...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# NAO
+# HULK
 
 This repository contains the robot control program and associated tools of the RoboCup SPL team HULKs.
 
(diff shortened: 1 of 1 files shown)


Bendodroid:
The GNU howto explicitly says not to do this and update the years. It also says we should have a license notice in each file and we're not doing that either so what do I even know... 

https://www.gnu.org/licenses/gpl-howto.html

h3ndrk:
> The GNU howto explicitly says not to do this and update the years. It also says we should have a license notice in each file and we're not doing that either so what do I even know...
> 
> https://www.gnu.org/licenses/gpl-howto.html

Thanks for your answer! What a pity to have to do that manually...

julianschuler (approved):
LGTM!

Bendodroid:
Slightly confusing commit message and PR title but ok 🥴

PR #332, opened 2023-05-07
Twixfix

okiwi6:
## Introduced Changes

Two small fixes for twix I noticed.
* The completion edit hint always showed `Subscription Key`. I changed this, so that e.g. the Panel completion edit shows `Panel` as hint.
* The arrow up functionality in the completion edit dialog allowed for negative indices, since in rust modulo != remainder (meaning -2 % 5 = -2). Now the correct function is used to compute the remainder.

Fixes #

## ToDo / Known Issues

## How to Test

* Make sure that the mentioned changes are correct.

### DIFF
--- a/tools/twix/src/completion_edit.rs
+++ b/tools/twix/src/completion_edit.rs
@@ -26,13 +26,15 @@ impl CompletionState {
 }
 
 pub struct CompletionEdit<'key> {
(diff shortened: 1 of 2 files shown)


knoellle (approved):
LGTM

PR #336, opened 2023-05-11
Resize all textures to 256x256

schmidma:
## Introduced Changes

Resizes the field textures for webots to 256x256


schluis:
This seems really small 🤔

schmidma:
This seems small: On what do you base your judgement? "Seems small" sounds like "too small", which needs a reference point. Too small for what?
Why: We experienced a non negligible runtime improvement during webots simulation.

PasGl:
The speed-up seems to vary a lot from system to system. I'd encourage everybody to do a quick comparison.

For me (Laptop, no dedicated GPU):
from ~0.40x to ~0.45x

On other systems we saw more significant speed-ups.

MaikRe:
RTX 3060 Laptop
Previously:
1.5x-1.85x
After:
1.55x-1.95x

knoellle:
On my laptop
main:
0.4x-0.45x
with smaller textures:
0.45x-0.55x
with smaller textures and removing normal map and occlusion map from `Field.proto`:
0.5x-0.6x

knoellle (changes requested):
I like these changes but maybe we can go even further.

I think we only need the base color texture anyways.
When removing the other two, I get even slightly more performance (see my comment) and it doesn't seem to alter the visuals much if at all.

I'd suggest removing the other two files and their references in `Field.proto`.

knoellle:
What is the state here?
We can merge this as is, but I would prefer to also add the further optimizations I suggested.

knoellle (approved):
Lets merge this for now.

PR #339, opened 2023-05-13
Pepsi unstiff

knoellle:
## Introduced Changes

Simulates pressing the head buttons prior to upload.
If the robot is standing upright, this will cause it to sit down, otherwise it becomes unstiff immediately.
`hulk stop` is postponed until the robot is unstiff making the upload process wait until the sit down motion completes.
If the communication server on the robot cannot be reached for a given timeout duration, the upload process continues as usual.
This can happen if the hulk service is not running or communication is inactive.

Fixes #

## ToDo / Known Issues

- [ ] Add dedicated `pepsi sitdown`? `pepsi unstiff`? command.
- [ ] Discuss how to handle communication failure
  - Currently pepsi attempts to connect to the communication server and silently skips sit down if there was no response for one second.
- [ ] Allow skipping sit down via command line argument. May also be a solution to the point below.
- [ ] Figure out how to cleanly handle uploads to robots running older code
  - Since setting parameters is currently a fire-and-forget operation, the headbutton press is sent but fails because the parameter doesn't exist causing pepsi to wait for unstiff indefinitely.

- Replace all our `injected_.*` with with injected outputs, see https://github.com/HULKs/nao/pull/4065.
- Use RPCs instead of fiddling with parameters/injections, requires communication protocol changes.
- Make hula make the robot sit down gracefully if the robot is standing and the hulk process is stopped/crashes

## How to Test

Upload to nao without unstiffing first, it should sit down on its own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2768,7 +2768,9 @@ name = "nao"
 version = "0.1.0"
 dependencies = [
  "color-eyre",
+ "communication",
  "constants",
+ "serde_json",
  "tokio",
 ]
 
--- a/crates/control/src/primary_state_filter.rs
+++ b/crates/control/src/primary_state_filter.rs
@@ -19,6 +19,8 @@ pub struct CycleContext {
     pub filtered_game_state: Input<Option<FilteredGameState>, "filtered_game_state?">,
     pub game_controller_state: Input<Option<GameControllerState>, "game_controller_state?">,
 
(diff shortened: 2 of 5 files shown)


oleflb:
If we want to do this, we probably want to do it properly from within hula, right? This would mean the robot could sit down if the framework crashes for example. If that's the case maybe we can close this PR?

knoellle:
Yes

PR #342, opened 2023-05-16
Behavior simulator FOV

knoellle:
## Introduced Changes

- Adds head movement to behavior simulator
  - Includes running executing the look around motion node.
- Limits ball visibility by field of view

Other fixes:
- Makes simulation playback real time 
  - Now runs at 83 Hz instead of whatever refresh rate your monitor happens to run at
- Makes simulated kicks more consistent
  - Previously force was applied to the ball for every cycle where the robot requested a kick motion
- Adds ball position ground truth visualization to behavior simulator map layer
  - Since ball visibility is now limited, the map panel would only show the ball as seen (or not) by the currently selected robot.
- Fixes kick selector cycle invocation in behavior cycler

Fixes #

## ToDo / Known Issues

- [x] Make FOV depend on config parameters

- Readd a visualization for which robots see the ball

## How to Test

In behavior simulator scenarios, the robots should now be seen doing ball search even when a ball exists on the field.

### DIFF
--- a/crates/types/src/camera_matrix.rs
+++ b/crates/types/src/camera_matrix.rs
@@ -80,7 +80,7 @@ impl CameraMatrix {
         }
     }
 
-    fn calculate_field_of_view(
+    pub fn calculate_field_of_view(
         focal_lengths: Vector2<f32>,
         image_size: Vector2<f32>,
     ) -> Vector2<f32> {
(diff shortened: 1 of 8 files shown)


MaikRe (approved):
LGTM

knoellle:
Field of view now depends on the actual camera parameters.

MaikRe (approved):
LGTM

PR #347, opened 2023-05-30
Joint Calibration Provider

### LINKED ISSUE #58: Add facility for Joint Calibration
We've noticed that multiple robots are experiencing wear n tear, including but not limited to head yaw offsets, body-roll offsets, etc.

While some could be corrected by introducing a "torso" calibration matrix, it would be nice to be able to individually add/ compensate an offset from joints.

(ie: Nao 30 after a bunch of remote games has 4~ deg of pitch in both cameras and that's the only offset. Also in some Naos, I felt more backlash (getting close to old V5 levels)

Reference from C++
https://github.com/HULKs/nao/blob/cpp/src/Motion/JointCalibrationProvider/JointCalibrationProvider.hpp

(Originally at: https://github.com/HULKs/nao/issues/4044)

tuxbotix:
As the title says, based on the C++ implementation we had. These offsets are used to correct sensor misalignments and other joint angle offsets.

## Introduced Changes

* ~Introduces a joint calibration provider which converts the parameters to radians and this output is used by sensor_data_reciever.~
* The parameters (in radians) are directly used by sensor data provider and joint command sender.
* Fixes a bug in `joints.rs` for `Joints<T>::as_vec()`

Fixes #58 

## ToDo / Known Issues
- [x] Identify the following crash: -> Happened due to the introduced calibration node running before sensor data provider.
    ```rust
    2023-05-30 04:58:30  parameters::directory  DEBUG  Reading "etc/configuration/default.json"...
    2023-05-30 04:58:30  parameters::directory  DEBUG  Reading "etc/configuration/webots_location/default.json"...
    The application panicked (crashed).
    Message:  Failed to get historic input value at given timestamp
    Location: ___/hulk/crates/framework/src/historic_input.rs:21
    ```

## How to Test

* Change values of parameter `joint_calibration_offsets_deg`, i.e. head yaw, the NAO will move the head accordingly.


crates/control/src/motion/joint_command_sender.rs
```diff
@@ -95,12 +96,20 @@ impl JointCommandSender {
...
             .hardware_interface
-            .write_to_actuators(positions, stiffnesses, *context.leds)
+            .write_to_actuators(compensated_positions, stiffnesses, *context.leds)
             .wrap_err("failed to write to actuators")?;
 
-        context.positions.fill_if_subscribed(|| positions);
+        context
+            .positions
+            .fill_if_subscribed(|| compensated_positions);
+
```
knoellle, line 112: This seems wrong to me.
I would argue that writing "compensated positions" into an additional output called "positions" is confusing.
Perhaps add a second output instead and output both the requested and compensated positions?

tuxbotix: Yes you are right, I was confused a bit and corrected the others but not this.

Compensated position is what the hardware thinks it has to reach while the position is the actual position we need to reach.

I'll add an additional output in case it helps.

tuxbotix: Fixed

knoellle (approved):
LGTM

Issue #299, opened 2023-04-29
Fix initial look-around angles


Narcha:
During the initial look-around, we look much further to the side than is necessary.

PR #349, opened 2023-05-31
Fix initial look-around angles 

MeiiMoon:
## Introduced Changes

1. Additional Output for look around mode was created so that that the mode for the head motion can be viewed on twix when subscribed.
2. The time for the pauses of the head motion was adjusted (decreased from 1 s to 0.8 s) so the robot has enough time to reach all the positions (i.e. center, half-way left, left, half-way right and right) before it exits the initial look around.
3. The signs of the Yaw angles in default.json were flipped. because the original signs were causing the opposition direction movement than intended, i.e., at the start, the robot was suppose to turn left first according to the code, but it turned right.
4.Value of the Yaw angles were also adjusted to make the look around range smaller. This fixes the problem that the robot looks outside the field. 

## How to Test

Using this code, look at how the robot head turns once it starts to move.  Also place the robot on all the different initial positions on the field, to see that the head turns far enough to see the respective corner of the penalty box.


MaikRe (approved):
LGTM

schluis:
Solves #299

PR #355, opened 2023-06-01
Error message fix in pepsi ping

okiwi6:
## Introduced Changes

This PR corrects the cryptic error message `deadline has elapsed` which was introduced with `pepsi ping`

Fixes #

## ToDo / Known Issues

## How to Test

* Trigger the timeout by `pepsi ping 31 --timeout 0.1`, the error message now says `No route to {host}` instead of `deadline has elapsed`


MaikRe (approved):
Works as advertised

PR #357, opened 2023-06-01
Reduce max allowed adjustment velocity of walk step adjustment

philipniklas-r:
## Introduced Changes

Reduce max allowed adjustment velocity of walk step adjustment. This should make the walking in extreme situations (tilting backward, tilting forward) more stable.

Fixes #

## ToDo / Known Issues

-

-

## How to Test

- Let a robot walk towards a ball -> should behave normal
- Let a robot walk towards a ball and give it small pushes from the front and back, with some time inbetween. -> robot should balance the disturbance out a lot better and be more stable than before. The feet should no longer move opposite of the "fall" direction a lot when switching the support foot.
  - For example when pushing from the front to tilt the robot backward, at the moment of a support foot switch the new swing foot should not move by a large distance forward, only by a few cm.
- Let a robot walk in place (so something like a walking speed of 0.00001 forward, 0 side and 0 turn) and give it a few pushes from the front and back
  - Robot should be more stable than before

### DIFF
--- a/etc/configuration/default.json
+++ b/etc/configuration/default.json
@@ -378,7 +378,7 @@
     },
     "max_number_of_timeouted_steps": 3,
     "max_number_of_unstable_steps": 3,
-    "max_step_adjustment": 0.004,
+    "max_step_adjustment": 0.0018,
     "maximal_step_duration": { "nanos": 0, "secs": 1 },
     "minimal_step_duration": { "nanos": 150000000, "secs": 0 },
     "number_of_stabilizing_steps": 3,


philipniklas-r:
If I see it correctly here: https://github.com/HULKs/hulk/blob/main/crates/control/src/motion/walking_engine/balancing.rs#L92
then the correct code has a huge flaw.

When the robot is walking with max speed forward (currently 0.055 -> 0.055 * 0.25 -> 0.22 -> 220 mm/s https://github.com/HULKs/hulk/blob/main/etc/configuration/default.json#L334), then in each step on average the swing foot will move 0.0026 (2.6 mm). With the current step adjustment speed per frame of 0.004 (-> 300 mm/s), the adjustment could only adjust when tilting backwards with an actual speed of 

0.004 - 0.0026 -> 0.0014 (close to what I suggest).

But when tilting forward, the swing foot will move with

0.004 + 0.0026 -> 0.0066 (540 mm/s)

Similar case when walking backwards, where the same directions add up and the opposite one reduce each other.
This is because you don't substract the "planned" step change of `` next_left_walk_request.forward - last_left_walk_request.forward`` from your ``limited_adjustment``. 

If someone with some more knowledge agrees that this is the case, then my PR is insufficient.
As an additional note for some more insights, if the robot is already walking in the direction of the fall, we subtract the planned velocity from the max allowed velocity, but clip it to still allow at least a small amount (currently 40 mm/s -> 0.00048). But this is something for another PR in the future.

philipniklas-r:
Another thing I just noticed, which was never mentioned in the paper or the talk (if I remember correctly): 

Allowing **no** adjustment speed when the swing foot measures some pressure (so holding the feet in place is allowed, further changes are not; except at the start of the walk step) also increases the stability significantly. (As a reference: https://github.com/bhuman/BHumanCodeRelease/blob/master/Src/Modules/MotionControl/WalkingEngine/WalkLibs/WalkStepAdjustment.cpp#L190)
I also tried to reproduce your version with one of our V5 robots. As the foot leveling is different, the heels got stuck in the ground faster, but with some help the swing foot also moved very far forward after a support foot switch (in the case of walking and tilting backwards).

Therefore I recommend to look into a version of that too. 

If you prefer I can close this PR and open an issue with all necessary information for you, as both problems I just noticed require some more coding than just changing a number.

schluis (approved):
Hi Philip, Thank you very much for your contribution!
We've had a look at the functions and agree, there are some serious bugs! We're currently rewriting a lot of this, but your parameter definitely improves the performance for now!
Regarding your other comment, we're looking into it. So if you want, feel free to copy the text into a new issue as you suggested 👍🏻

PR #361, opened 2023-06-07
1.70.0 Clippy Happyning

knoellle:
## Introduced Changes

Fixes/masks clippy lints that were added or changed in rust 1.70.0.
1. Added `allow(...)`s for borked lint: https://github.com/rust-lang/rust-clippy/issues/10893
2. Bumped `nix` dependency since clippy complained about a calculation in a macro from that crate. This has been fixed upstream in April, but there hasn't been a release yet, so I'm using a pinned git version.

Fixes #

Unmask redundant clone check once that is fixed upstream.

## How to Test

`pepsi clippy --workspace`

### DIFF
--- a/crates/framework/src/multiple_buffer.rs
+++ b/crates/framework/src/multiple_buffer.rs
@@ -209,6 +209,8 @@ mod tests {
             assert_eq!(*slot, 42);
         }
         {
+            // clippy bug, remove once fix is released
+            // https://github.com/rust-lang/rust-clippy/issues/10893
             #[allow(clippy::redundant_clone)]
             let reader2 = reader.clone();
             let slot = reader2.next();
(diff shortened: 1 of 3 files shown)


crates/framework/src/multiple_buffer.rs
```diff
@@ -209,6 +209,7 @@ mod tests {
             assert_eq!(*slot, 42);
         }
         {
+            #[allow(clippy::redundant_clone)]
```
h3ndrk, line 212: I would love to have a comment here that this can go away at some time with a link to the issue.

h3ndrk (changes requested):

crates/framework/src/multiple_buffer.rs
knoellle: Added comments to all three locations.

h3ndrk (approved):

PR #362, opened 2023-06-09
Gammaray follow symlinks

knoellle:
## Introduced Changes

What it says.
Yocto builds an image file with a rather verbose name. Since the name includes a timestamp, it also changes on every build.
There exists however a symlink to this file with a shorter name, only including the artifact name and version.
This change allows using this symlink for the `image-path` in `pepsi gammaray`.
Previously the symlink would simply be skipped by rsync as a "non-regular file" and the nao would reboot without flashing.

Fixes #

## How to Test

Run `pepsi gammaray --image-path ...` and point it at a symlink to an image.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -324,6 +324,7 @@ impl Nao {
     pub async fn flash_image(&self, image_path: impl AsRef<Path>) -> Result<()> {
         let status = self
             .rsync_with_nao(false)
+            .arg("--copy-links")
             .arg(image_path.as_ref().to_str().unwrap())
             .arg(format!("{}:/data/.image/", self.host))
             .status()


crates/nao/src/lib.rs
```diff
@@ -324,6 +324,7 @@ impl Nao {
     pub async fn flash_image(&self, image_path: impl AsRef<Path>) -> Result<()> {
         let status = self
             .rsync_with_nao(false)
+            .arg("--copy-links")
```
h3ndrk, line 327: Does this only allow you to give a symlink as image input file or will it also affect what will be copied to the NAO? We probably don't want to copy symlinks to the NAO because the Device Firmware Upgrade is not trustworthy :wink:

h3ndrk (changes requested):

crates/nao/src/lib.rs
knoellle: From the `rsync` manual page:
> --copy-links, -L         transform symlink into referent file/dir

And it does seem to work that way. I tested it by building an new image and flashing it that way.

h3ndrk: Ok, thanks! Could you please test what happens in the destination directory? We should be sure to not copy a symlink to the directory.

knoellle: Don't have a nao to test right now, but when testing locally, it indeed turns the link into the file its pointing to when copying. There are no extraneous symlinks at the destination.

h3ndrk: Great, fine for me. Thanks for testing. Resolving.

h3ndrk (approved):

PR #363, opened 2023-06-10
Add Webots keyboard shortcuts for head buttons

h3ndrk:
## Introduced Changes

What it says. Allows to enter and exit calibration from Webots.

## ToDo / Known Issues

None

None

## How to Test

Compile for and run in Webots.

- Hold <kbd>⌃ Control</kbd> + <kbd>⇧ Shift</kbd> + <kbd>X</kbd> (front head button + chest button) for 1 second to enter calibration mode
- Press <kbd>⌃ Control</kbd> + <kbd>⇧ Shift</kbd> + <kbd>C</kbd> (chest button) for normal primary state changes
- Hold <kbd>⌃ Control</kbd> + <kbd>⇧ Shift</kbd> + <kbd>U</kbd> (all head buttons) for 0.1 seconds to unstiff the NAO

### DIFF
--- a/crates/hulk/src/webots/keyboard_device.rs
+++ b/crates/hulk/src/webots/keyboard_device.rs
@@ -20,7 +20,6 @@ impl Default for KeyboardDevice {
 
 impl KeyboardDevice {
     pub fn get_touch_sensors(&self) -> TouchSensors {
-        let key = self.keyboard.get_key();
         let received_message = match self.receiver.get_next_packet() {
             Ok(message) => message,
             Err(error) => {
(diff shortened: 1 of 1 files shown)


knoellle (changes requested):
With this PR applied, the way <kbd>⌃ Control</kbd> + <kbd>⇧ Shift</kbd> + <kbd>C</kbd> works changes.
For example, when the robot is in initial or playing, holding down the chestbutton key combination causes the robot to go to unstiff (performing the sit down motion). Upon release, it returns to initial.
On current main, this behavior is different. Holding down the button does nothing and only upon release does it go to penalized (not unstiff).

Since the key only has to be "held" for a few cycles, this makes getting to robot to start moving initially quite inconsistent. Usually I just press the key combo for chest button thrice and it starts moving. However, with these changes, the second and third press become a lot more time sensitive.

h3ndrk:
Ah, Mist. `!= 0` is probably too general. It should be equal to the mask to be true, right?

knoellle:
Yes, I just found that too.
Pressing ctrl + shift + c was also triggering ctrl + shift + u, since they share some 1 bits.

h3ndrk:
Since Webots does not support to press multiple keys at once (e.g. <kbd>X</kbd> and <kbd>C</kbd>) the shortcut <kbd>⌃ Control</kbd> + <kbd>⇧ Shift</kbd> + <kbd>X</kbd> will now trigger the front head button and chest button at once which promotes it as calibration button.

crates/hulk/src/webots/keyboard_device.rs
```diff
@@ -29,14 +28,28 @@ impl KeyboardDevice {
...
+                const CONTROL_SHIFT_X_MASK: u32 = CONTROL_SHIFT_MASK | 'X' as u32;
+                const CONTROL_SHIFT_U_MASK: u32 = CONTROL_SHIFT_MASK | 'U' as u32;
+                (
+                    key & CONTROL_SHIFT_C_MASK == CONTROL_SHIFT_C_MASK,
+                    key & CONTROL_SHIFT_X_MASK == CONTROL_SHIFT_X_MASK,
+                    key & CONTROL_SHIFT_U_MASK == CONTROL_SHIFT_U_MASK,
+                )
+            } else {
+                (false, false, false)
+            };
```
knoellle, line 44: I'm afraid this logic is still broken. Pressing Ctrl + Shift + W for example triggers both the chest button and all three head buttons.

I would suggest simply checking for equality like before. I think that way is more sound anyways, you shouldn't just ignore modifier keys you don't require. If the user presses Mod + Ctrl + Shift + C, that probably shouldn't trigger our keybindings.
If you _do_ want to keep bitschubsing, I'd suggest using the `Keyboard::KEY` mask for removing the modifier bits before comparing to the numerical character value.

h3ndrk: Yeah, since Webots does not support pressing multiple keys at once, this whole feature does not make sense anymore. I'm now only checking whether the `key` is equal to the `*_MASK` and not ANDing it before. This should get near to where we came from.

knoellle (approved):
LGTM

PR #364, opened 2023-06-10
Allow to correct camera matrices

h3ndrk:
## Introduced Changes

What it says. The correction is currently zero so everything should work as before.

## ToDo / Known Issues

None

None

## How to Test

Compile and look through Twix to observe that projected field lines are same as before.

### DIFF
--- a/crates/types/src/camera_matrix.rs
+++ b/crates/types/src/camera_matrix.rs
@@ -1,4 +1,5 @@
-use nalgebra::{Isometry3, Matrix, Point2, Vector2};
+use approx::{AbsDiffEq, RelativeEq};
+use nalgebra::{Isometry3, Matrix, Point2, Rotation3, Vector2};
 use serde::{Deserialize, Serialize};
 use serialize_hierarchy::SerializeHierarchy;
 
@@ -12,7 +13,7 @@ pub struct CameraMatrices {
     pub bottom: CameraMatrix,
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, SerializeHierarchy)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize, SerializeHierarchy)]
 pub struct CameraMatrix {
     pub camera_to_head: Isometry3<f32>,
     pub camera_to_ground: Isometry3<f32>,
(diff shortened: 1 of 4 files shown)


crates/types/src/camera_matrix.rs
```diff
@@ -89,12 +90,118 @@ impl CameraMatrix {
...
+        let ground_to_robot = self.camera_to_robot * self.ground_to_camera;
+        let robot_to_ground = ground_to_robot.inverse();
+
+        let corrected_camera_to_head =
+            Isometry3::from_parts(Default::default(), correction_in_camera.inverse().into())
+                * camera_to_head;
+        let corrected_head_to_camera = corrected_camera_to_head.inverse();
+        let corrected_head_to_robot =
+            Isometry3::from_parts(Default::default(), correction_in_robot.inverse().into())
+                * head_to_robot;
```
knoellle, line 111: I may be wrong, but I think this multiplication are the wrong way around.
In the first calculation, when multiplying a vector from the right, it is first transformed from camera coordinates to head coordinates before being rotated "in camera" even though it is already in head coordinates.
The `corrected_head_to_robot` calculation seems correct, since you want to rotate in the destination frame of that transformation anyway.

h3ndrk: @jonathan-hellwig Could you comment on this? :joy:

crates/types/src/camera_matrix.rs
```diff
@@ -89,12 +90,118 @@ impl CameraMatrix {
...
+        let corrected_robot_to_head = corrected_head_to_robot.inverse();
+
+        CameraMatrix {
+            camera_to_head: corrected_camera_to_head,
+            camera_to_ground: robot_to_ground * corrected_head_to_robot * corrected_camera_to_head,
+            ground_to_camera: corrected_head_to_camera * corrected_robot_to_head * ground_to_robot,
+            camera_to_robot: corrected_head_to_robot * corrected_camera_to_head,
+            robot_to_camera: corrected_head_to_camera * corrected_robot_to_head,
```
h3ndrk, line 119: Reminder to self: Let's save some duplicate multiplications here.

crates/types/src/camera_matrix.rs
jonathan-hellwig: For the first one I agree with @knoellle and for the second one I would suggest a different name. I'll put it into math notation to make it clear: Let $C$ be the uncorrected camera system, $\tilde{C}$ the corrected camera system and $H$ be the head system. Then, the first transformation is given by $T_{\tilde{C}}^H = T^H_C T^C_{\tilde{C}}$. 
For the second transformation we have the uncorrected robot system $R$, the corrected robot system $\tilde{R}$ and the head system $H$. Then, the second transformation is given by $T^{\tilde{R}}_H = T^{\tilde{R}}_R T^R_H$. 
I would suggest to name the transformation `head_to_corrected_robot`, because it makes the reference frame of a point transformed by $T^{\tilde{R}}_H$ more precise. If you just use `corrected_*` everytime it is not clear whether you modified the original or the target coordinate sytem.

h3ndrk: Thanks, I renamed some variables and swapped the multiplication operands in the first calculation.

crates/types/src/camera_matrix.rs
h3ndrk: Done. Resolving.

knoellle (approved):
LGTM

PR #199, opened 2023-04-18
Drop obstacles on Unpenalized

schmidma:
## Introduced Changes

drops all obstacles in the filter, when leaving a penalty.

Based on #190 #191 #198 

## ToDo / Known Issues

none

## Ideas for Next Iterations (Not This PR)

none

## How to Test

Use your hand to generate a sonar obstacle, press the chest button to leave penalized state, the obstacle should be removed

PR #365, opened 2023-06-10
Drop all obstacle hypotheses when unpenalized (without PrimaryStateTransition)

h3ndrk:
## Introduced Changes

Drops all obstacles in the filter, when leaving a penalty. This PR reimplements #199 without `PrimaryStateTransition`. Credit goes to Schmiddy. :muscle:

## ToDo / Known Issues

none

none

## How to Test

Use your hand to generate a sonar obstacle, press the chest button to leave penalized state, the obstacle should be removed

### DIFF
--- a/crates/control/src/obstacle_filter.rs
+++ b/crates/control/src/obstacle_filter.rs
@@ -10,11 +10,12 @@ use types::{
     configuration::ObstacleFilter as ObstacleFilterConfiguration, detected_feet::DetectedFeet,
     detected_robots::DetectedRobots,
     multivariate_normal_distribution::MultivariateNormalDistribution, obstacle_filter::Hypothesis,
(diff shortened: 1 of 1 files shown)


knoellle:
The reason for implementing the original PR #199 was to avoid the robots running around phantom obstacles after a false positive obstacle detection caused by a chest button pressing hand.
Is this still relevant now that the robot looks around for five seconds before starting to move?

h3ndrk:
I don't know how long obstacles are persisted but probably it doesn't hurt to remove them as it just makes sense. I generally like the idea to start with a fresh state (at least some relevant state) when coming back from penalized.

knoellle:
Alright, fair point.

knoellle (approved):
LGTM

PR #366, opened 2023-06-11
Allow to mark crossings in Twix

h3ndrk:
## Introduced Changes

This PR only contains a new panel that allows to mark line crossings in an image. The marked crossings are not sent to the NAO yet because there is no infrastructure for accepting it. This will come in a later PR.

I'm not an expert in egui, please provide feedback and improvements. The code quality is probably not the best... :sweat_smile:

image

## ToDo / Known Issues

None

- Allow to view bottom camera
- Sent crossings to NAO and use them

## How to Test

Compile Twix and start the new panel. Crossings can be toggled via:

- <kbd>Q</kbd>: toggle left field corner at mouse hover position
- <kbd>W</kbd>: toggle left penalty field at mouse hover position
- <kbd>E</kbd>: toggle left penalty corner at mouse hover position
- <kbd>R</kbd>: toggle left goal field at mouse hover position
- <kbd>T</kbd>: toggle left goal corner at mouse hover position
- <kbd>Z</kbd>: toggle right goal corner at mouse hover position
- <kbd>U</kbd>: toggle right goal field at mouse hover position
- <kbd>I</kbd>: toggle right penalty corner at mouse hover position
- <kbd>O</kbd>: toggle right penalty field at mouse hover position
- <kbd>P</kbd>: toggle right field corner at mouse hover position


tuxbotix:
Works as advertised;
I have a few useability suggestions:
1. `ui.label` or a similar element to display the key bindings.
2. Another label to display the active binding so the user clicks it correctly.
3. Optionally (nice to have) some text at each line or colour mapping to ensure the crosses are added correctly.
    - We could add fancier validations but I think this would be overkill

tools/twix/src/main.rs
knoellle, line 1: I have some general ideas for improvements:
1. Initialize all markers with roughly the correct position such that the user only has to do some corrections.
2. Instead of spawning a marker at the cursor position on keypress, make them draggable. In the first iteration this can be achieved by teleporting the nearest marker to the cursor position during every update cycle where the left mouse button is pressed, optionally with a maximum distance.
3. Optionally, show the name of the nearest marker in a label next to the image.

Also, ff you want to stick to the current method of toggling markers by keypresses:
1. What @tuxbotix said already
2. Instead of toggling the markers on key press, do it on mouse click and automatically advance to the next marker type. Keep the key bindings to jump to specific marker types or have two to cycle forward/backwards and show the current one in a label.

tools/twix/src/panels/manual_crossing_marker.rs
```diff
@@ -0,0 +1,284 @@
...
+                point![0.0, position.y],
+                point![640.0, position.y],
+                Stroke::new(1.0, Color32::WHITE),
+            );
+            if ui.input_mut(|input| input.consume_key(Modifiers::NONE, Key::Q)) {
+                self.crossings.left_field_corner = match self.crossings.left_field_corner {
+                    Some(_) => None,
+                    None => Some(position),
+                };
+            }
```
knoellle, line 205: There is a lot of duplication here.
You could use [`Option::xor(Some(...))`](https://doc.rust-lang.org/std/option/enum.Option.html#method.xor) or better create a function that does this swap for you.
Perhaps this can also do the keypress checking?

tools/twix/src/panels/manual_crossing_marker.rs
```diff
@@ -0,0 +1,284 @@
...
+        let image_response = image.show_scaled(ui, scale);
+        let displayed_image_size = image_size * scale;
+        let image_rect = Rect::from_min_size(image_response.rect.left_top(), displayed_image_size);
+        let painter = TwixPainter::paint_at(ui, image_rect).with_camera(
+            vector![640.0, 480.0],
+            Similarity2::identity(),
+            CoordinateSystem::LeftHand,
+        );
+
+        if let Some(position) = self.crossings.left_field_corner {
```
knoellle, line 90: If `crossings` was an iterable mapping this could be a lot simpler...

tools/twix/src/panels/manual_crossing_marker.rs
```diff
@@ -0,0 +1,284 @@
...
+    left_penalty_field: Option<Point2<f32>>,
+    left_penalty_corner: Option<Point2<f32>>,
+    left_goal_field: Option<Point2<f32>>,
+    left_goal_corner: Option<Point2<f32>>,
+    right_goal_corner: Option<Point2<f32>>,
+    right_goal_field: Option<Point2<f32>>,
+    right_penalty_corner: Option<Point2<f32>>,
+    right_penalty_field: Option<Point2<f32>>,
+    right_field_corner: Option<Point2<f32>>,
+}
```
knoellle, line 31: Suggestion: Use a map from enum to Point instead.

knoellle (changes requested):

PR #367, opened 2023-06-11
Add behavior action for PrimaryState::Calibration

h3ndrk:
## Introduced Changes

What it says. Currently when entering the calibration primary state, the HULK will crash because there is no action available. To allow to use this action for manual calibration, for the moment, in primary state calibration we will stand and have an unstiff head which allows to rotate it by hand.

## ToDo / Known Issues

None

None

## How to Test

If testing in Webots, cherry pick #363. If testing on the NAO, just deploy this branch.

Then enter the calibration mode by holding the front head button in initial and holding the chest button for 1 second. The chest button should change color, as the rules require that. The robot should transition into a stand motion with an unstiff head.

### DIFF
--- a/crates/control/src/behavior/calibrate.rs
+++ b/crates/control/src/behavior/calibrate.rs
@@ -1,0 +1,11 @@
+use types::{MotionCommand, PrimaryState, WorldState};
+
+pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+    match world_state.robot.primary_state {
+        PrimaryState::Calibration => Some(MotionCommand::Stand {
+            head: types::HeadMotion::Unstiff,
+            is_energy_saving: false,
+        }),
+        _ => None,
+    }
+}
--- a/crates/control/src/motion/head_motion.rs
+++ b/crates/control/src/motion/head_motion.rs
@@ -97,21 +97,22 @@ impl HeadMotion {
     }
 
     pub fn joints_from_motion(context: &CycleContext) -> HeadJointsCommand<f32> {
+        let stiffnesses = HeadJoints::fill(0.8);
         match context.motion_command.head_motion() {
             Some(HeadMotionCommand::Center) => HeadJointsCommand {
                 positions: *context.center_head_position,
-                stiffnesses: HeadJoints::fill(0.8),
+                stiffnesses,
             },
             Some(HeadMotionCommand::LookAround | HeadMotionCommand::SearchForLostBall) => {
                 HeadJointsCommand {
                     positions: *context.look_around,
-                    stiffnesses: HeadJoints::fill(0.8),
+                    stiffnesses,
                 }
             }
             Some(HeadMotionCommand::LookAt { .. })
             | Some(HeadMotionCommand::LookLeftAndRightOf { .. }) => HeadJointsCommand {
                 positions: *context.look_at,
-                stiffnesses: HeadJoints::fill(0.8),
+                stiffnesses,
             },
             Some(HeadMotionCommand::Unstiff) => HeadJointsCommand {
                 positions: context.sensor_data.positions.head,
@@ -119,7 +120,7 @@ impl HeadMotion {
             },
             Some(HeadMotionCommand::ZeroAngles) | None => HeadJointsCommand {
                 positions: Default::default(),
-                stiffnesses: HeadJoints::fill(0.8),
+                stiffnesses,
             },
         }
     }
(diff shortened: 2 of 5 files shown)


crates/control/src/behavior/calibrate.rs
```diff
@@ -0,0 +1,11 @@
+use types::{MotionCommand, PrimaryState, WorldState};
+
+pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+    match world_state.robot.primary_state {
+        PrimaryState::Calibration => Some(MotionCommand::Stand {
+            head: types::HeadMotion::Unstiff,
+            is_energy_saving: false,
```
h3ndrk, line 7: Request for comments: Should we enable energy-saving stand? Is this already implemented?

tuxbotix:
LGTM (without testing on a NAO); As you suggested, we can add the look-at after #346 is merged ^_^

crates/control/src/behavior/calibrate.rs
knoellle: It is implemented, but disabled by default, see parameter `energy_saving_stand.enabled`.
Not sure how much of a difference that makes for calibration, but the head will be about 3 cm higher above the ground with energy saving stand enabled.

h3ndrk: Ok, let's don't care then and keep it as it is. :muscle:

crates/control/src/behavior/calibrate.rs
```diff
@@ -0,0 +1,11 @@
+use types::{MotionCommand, PrimaryState, WorldState};
+
+pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+    match world_state.robot.primary_state {
+        PrimaryState::Calibration => Some(MotionCommand::Stand {
+            head: types::HeadMotion::Unstiff,
```
knoellle, line 6: Warning, this currently does not actually unstiff the head.
The head is always driven with 0.8 stiffness as implemented [here](https://github.com/HULKs/hulk/blob/main/crates/control/src/motion/head_motion.rs#L90-L96).

This does allow one to move the head around by hand because the unstiff motion always requests the currently measured joint angles and thus is subject to induced drift. However, the head angles are still clamped, for example negative pitch values are not allowed.
If that's fine for you, I'm fine with it, but I also tried it 0 stiffness which makes it much easier to move the head by hand but still keeps its position due to the weird motor brake "feature" and also allows a larger range of head positions.

h3ndrk: Wtf, unstiff is not unstiff... :exploding_head: I will correct this.

h3ndrk: Done

crates/control/src/motion/head_motion.rs
```diff
@@ -46,67 +44,83 @@ pub struct MainOutputs {
...
 
-        let clamped_request = HeadJoints {
+        let clamped_pitch = controlled_positions.pitch.clamp(0.0, maximum_pitch);
+        let clamped_positions = HeadJoints {
             pitch: clamped_pitch,
-            yaw: controlled_request.yaw,
+            yaw: controlled_positions.yaw,
         };
 
-        self.last_request = controlled_request;
```
h3ndrk, line 89: @schmidma What was the reason to implement the `HeadMotion` open-loop with a state variable instead of using the current sensor positions? See https://github.com/HULKs/nao/commit/895151d0104a06e7535d6d681c1ffca4f5710a72#diff-e20fc1836385ae33c49171b884fa87cc2036a451dd690dae352724c517b2261eR70

crates/control/src/motion/head_motion.rs
```diff
@@ -46,67 +44,83 @@ pub struct MainOutputs {
...
 
-    pub fn joints_from_motion(context: &CycleContext) -> HeadJoints<f32> {
+    pub fn joints_from_motion(context: &CycleContext) -> HeadJointsCommand<f32> {
         match context.motion_command.head_motion() {
-            Some(HeadMotionCommand::Center) => *context.center_head_position,
-            Some(HeadMotionCommand::LookAround) | Some(HeadMotionCommand::SearchForLostBall) => {
-                *context.look_around
+            Some(HeadMotionCommand::Center) => HeadJointsCommand {
+                positions: *context.center_head_position,
+                stiffnesses: HeadJoints::fill(0.8),
```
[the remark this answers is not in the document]
h3ndrk: Done via GitHub Codespaces 😎

crates/control/src/motion/head_motion.rs
h3ndrk: Readded open-loop interpolation (via GitHub on my Smartphone in the train 😎)

knoellle (approved):
LGTM

PR #369, opened 2023-06-13
Update hulks setup documentation

PatrickGoettsch:
## Introduced Changes

Updates the hulk documentation. There the setup part is slightly outdated

### DIFF
--- a/docs/setup/main_setup.md
+++ b/docs/setup/main_setup.md
@@ -66,7 +66,7 @@ Once the compilation step is complete, open webots and load the scene at `webots
 
 ## Running Webots in external Mode
 
-To not be fordeced to reload the scene in Webots in case of an updated hulk controller, you can run in webots `webots/worlds/penalized_extern.wbt` and starting the controller with:
+To not be forced to reload the scene in Webots when rebuilding the controller, you can run in webots `webots/worlds/penalized_extern.wbt` and starting the controller with:
 
 ```sh
 ./pepsi run
--- a/docs/setup/main_setup.md
+++ b/docs/setup/main_setup.md
@@ -44,21 +44,11 @@ Use your distribution's package manager to install the following dependencies:
 
 ## Acquiring the code
 
-=== "HULKs Members"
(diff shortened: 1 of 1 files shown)


docs/setup/main_setup.md
```diff
@@ -73,3 +63,11 @@ This will first build the pepsi binary and then start the build process.
 ## Running Webots
 
 Once the compilation step is complete, open webots and load the scene at `webots/worlds/penalized.wbt` from the repository.
+
+## Running Webots in external Mode
+
+To not be fordeced to reload the scene in Webots in case of an updated hulk controller, you can run in webots `webots/worlds/penalized_extern.wbt` and starting the controller with:
```
knoellle, line 69: 1. Typo: "fordeced"
2. "in case of an updated hulk controller" sounds like an external event. Perhaps write something like "... to reload the scene when rebuilding the controller"?

docs/setup/main_setup.md
```diff
@@ -29,7 +29,7 @@ Use your distribution's package manager to install the following dependencies:
     1. Install dependencies
 
         ```sh
-        sudo apt install git git-lfs build-essential libssl-dev pkg-config libclang-dev rsync cmake libhdf5-dev python
+        sudo apt install git git-lfs build-essential libssl-dev pkg-config libclang-dev rsync cmake libhdf5-dev python3 glibc-source
```
knoellle, line 32: I don't think installing this package did anything against the problem you were facing. I just tried the whole setup process in an ubuntu docker container and it worked without installing it.
Please remove but keep the python package name change.

knoellle (changes requested):

docs/setup/main_setup.md
PatrickGoettsch: Done

docs/setup/main_setup.md
PatrickGoettsch: Done

knoellle (approved):
LGTM

PR #370, opened 2023-06-14
Twix Light Mode

okiwi6:
## Introduced Changes

Allows changing the visual theme in twix with persistence.
This should give better visibility when using a beamer.

Fixes #

## ToDo / Known Issues

* Use system theme
* Use user themes

## How to Test

* Click the new settings icon in the upper right, navigate to theme and set it to sth. different


knoellle (approved):
Lights good to me

PR #373, opened 2023-06-17
Add pepsi logs show

julianschuler:
## Introduced Changes

This PR introduces a `pepsi logs show` subcommand for outputting the logs (hulk.out and hulk.err) of one or multiple NAOs.

## ToDo / Known Issues

None.

Filtering the log output may be interesting, though this may rather be done with an external tool.

## How to Test

Execute e.g. `./pepsi logs show 21`.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -183,6 +183,35 @@ impl Nao {
         Ok(())
     }
 
+    pub async fn retrieve_file(&self, file: &str) -> Result<String> {
+        let output = self
+            .ssh_to_nao()
+            .arg("cat")
+            .arg(file)
+            .output()
+            .await
+            .wrap_err("failed to execute cat command")?;
+
+        if !output.status.success() {
+            bail!("cat ssh command exited with {}", output.status);
+        }
+
+        String::from_utf8(output.stdout).wrap_err("failed to decode UTF-8")
+    }
+
+    pub async fn retrieve_logs(&self) -> Result<String> {
+        let mut hulk_out = self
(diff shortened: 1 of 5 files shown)


crates/nao/src/lib.rs
```diff
@@ -169,6 +169,7 @@ impl Nao {
 
         let status = self
             .rsync_with_nao(true)
+            .arg("-q")
```
okiwi6, line 172: Prefer --quiet?

tools/pepsi/src/ping.rs
```diff
@@ -24,25 +19,14 @@ pub struct Arguments {
...
-                Duration::from_secs_f32(arguments.timeout),
-            )
-            .await
-            .map(|_| TaskMessage::EmptyMessage);
-            progress.finish_with(ping_state);
-        })
-        .collect::<FuturesUnordered<_>>()
-        .collect::<Vec<_>>()
-        .await;
+    ProgressIndicator::map_tasks(arguments.naos, "Pinging NAO...", |nao_address| async move {
```
okiwi6, line 22: Nice, didn't know that function

okiwi6 (changes requested):
I don't quite like that error and log are not distinct in the output. Do you think there is some way to improve this?

tools/pepsi/src/ping.rs
julianschuler: 👍🏾

crates/nao/src/lib.rs
julianschuler: You're right, done.

okiwi6 (approved):

PR #377, opened 2023-06-17
Use alternative static ball model

okiwi6:
## Introduced Changes

We noticed that the robot can shoot significantly better when the ball model does not jitter.
This PR adds a resting equivalent to each hypothesis, which has higher velocity measurement noise and therefore is slower and doesn't have as much position noise.
The switch between moving and resting ball is determined depending on the velocity of the moving ball.

Fixes #

## ToDo / Known Issues

## How to Test

* Observe the `chooses_resting_model` additional output to see when the robot switches between the models.
* The robot should make a kick decision faster when the ball is steady (given the robot is properly localized)

### DIFF
--- a/tools/twix/src/panels/map/layers/ball_filter.rs
+++ b/tools/twix/src/panels/map/layers/ball_filter.rs
@@ -42,10 +42,10 @@ impl Layer for BallFilter {
         let ball_hypotheses: Vec<Hypothesis> = self.ball_hypotheses.parse_latest()?;
 
         for hypothesis in ball_hypotheses.iter() {
-            let position =
-                robot_to_field.unwrap_or_default() * Point2::from(hypothesis.state.mean.xy());
+            let position = robot_to_field.unwrap_or_default()
+                * Point2::from(hypothesis.moving_state.mean.xy());
             let covariance = hypothesis
-                .state
+                .moving_state
                 .covariance
                 .fixed_view::<2, 2>(0, 0)
                 .into_owned();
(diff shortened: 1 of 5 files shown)


okiwi6:
@jonathan-hellwig I found the issue. To avoid teleporting balls, it was necessary to have the same position in the resting and moving ball model once the robot decides for the resting ball. This now happens at the correct position in `update_hypothesis_with_measurement` which leads to very fast kick decisions.

jonathan-hellwig:
I'll have a look at your changes tomorrow morning.

okiwi6:
@jonathan-hellwig Do you think you can review this today?

jonathan-hellwig:
Yes, my apologies.

tools/twix/src/panels/map/layers/ball_filter.rs
```diff
@@ -28,27 +28,23 @@ impl Layer for BallFilter {
...
-
-        for hypothesis in ball_hypotheses.iter() {
-            let position =
-                robot_to_field.unwrap_or_default() * Point2::from(hypothesis.state.mean.xy());
-            let covariance = hypothesis
-                .state
-                .covariance
-                .fixed_view::<2, 2>(0, 0)
-                .into_owned();
+        let best_ball_state: Option<MultivariateNormalDistribution<4>> =
```
jonathan-hellwig, line 42: I think we should add back all detected hypotheses back to twix. However, considering current priorities, I suggest creating an issue to add all hypotheses back to twix at some later point in time.

jonathan-hellwig:
I couldn't find any other issues. Great job!

okiwi6:
I rebased on main, you can approve now

jonathan-hellwig (approved):
LGTM

PR #381, opened 2023-06-17
Calibration Optimization Problem Formulation (final final)

h3ndrk:
## Introduced Changes

Introduces camera calibration optimization problem formulation.

## ToDo / Known Issues

None

None

## How to Test

LATC (Look at the code)

### DIFF
--- a/crates/calibration/src/lines.rs
+++ b/crates/calibration/src/lines.rs
@@ -1,0 +1,82 @@
+use nalgebra::Point2;
+use projection::Projection;
+use types::{CameraMatrices, CameraMatrix, Line, Line2};
+
+#[derive(Clone)]
+pub struct Lines {
+    pub top: LinesPerCamera,
+    pub bottom: LinesPerCamera,
+}
+
+impl Lines {
+    pub fn to_projected(&self, matrices: &CameraMatrices) -> Result<Self, LinesError> {
+        Ok(Self {
+            top: project_lines_and_map_error(&matrices.top, &self.top, "top")?,
+            bottom: project_lines_and_map_error(&matrices.bottom, &self.bottom, "bottom")?,
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum LinesError {
+    #[error("failed to project {which} to ground")]
+    NotProjected {
+        source: projection::Error,
+        which: String,
+    },
+}
+
+fn project_lines_and_map_error(
+    matrix: &CameraMatrix,
+    lines: &LinesPerCamera,
+    which: &str,
+) -> Result<LinesPerCamera, LinesError> {
+    Ok(LinesPerCamera {
+        border_line: project_line_and_map_error(
+            matrix,
+            lines.border_line,
+            &format!("{which} border line"),
+        )?,
+        goal_box_line: project_line_and_map_error(
+            matrix,
+            lines.goal_box_line,
+            &format!("{which} goal box line"),
+        )?,
+        connecting_line: project_line_and_map_error(
+            matrix,
+            lines.connecting_line,
+            &format!("{which} connecting line"),
+        )?,
+    })
+}
+
+fn project_line_and_map_error(
+    matrix: &CameraMatrix,
+    line: Line2,
+    which: &str,
+) -> Result<Line2, LinesError> {
+    Ok(Line(
+        project_point_and_map_error(matrix, line.0, &format!("{which} point 0"))?,
+        project_point_and_map_error(matrix, line.1, &format!("{which} point 1"))?,
+    ))
+}
+
+fn project_point_and_map_error(
+    matrix: &CameraMatrix,
+    point: Point2<f32>,
+    which: &str,
+) -> Result<Point2<f32>, LinesError> {
+    matrix
+        .pixel_to_ground(point)
+        .map_err(|source| LinesError::NotProjected {
+            source,
+            which: which.to_string(),
+        })
+}
+
+#[derive(Clone)]
+pub struct LinesPerCamera {
+    pub border_line: Line2,
+    pub goal_box_line: Line2,
+    pub connecting_line: Line2,
+}
--- a/crates/calibration/src/lines.rs
+++ b/crates/calibration/src/lines.rs
@@ -1,18 +1,24 @@
 use nalgebra::Point2;
 use projection::Projection;
-use types::{CameraMatrices, CameraMatrix, Line, Line2};
+use types::{CameraMatrix, Line, Line2};
 
 #[derive(Clone)]
 pub struct Lines {
-    pub top: LinesPerCamera,
-    pub bottom: LinesPerCamera,
+    pub border_line: Line2,
+    pub goal_box_line: Line2,
+    pub connecting_line: Line2,
 }
 
 impl Lines {
-    pub fn to_projected(&self, matrices: &CameraMatrices) -> Result<Self, LinesError> {
-        Ok(Self {
-            top: project_lines_and_map_error(&matrices.top, &self.top, "top")?,
-            bottom: project_lines_and_map_error(&matrices.bottom, &self.bottom, "bottom")?,
+    pub fn to_projected(&self, matrix: &CameraMatrix) -> Result<Self, LinesError> {
+        Ok(Lines {
+            border_line: project_line_and_map_error(matrix, self.border_line, "border line")?,
+            goal_box_line: project_line_and_map_error(matrix, self.goal_box_line, "goal box line")?,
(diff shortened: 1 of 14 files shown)


crates/calibration/src/residuals.rs
```diff
@@ -0,0 +1,107 @@
...
+                        Residuals::calculate_from(parameters, measurements, field_dimensions)
+                            .ok()?
+                            .into(),
+                    );
+                    Some(residuals)
+                }
+                None => None,
+            },
+        )
+        .map(DVector::from_vec)
```
knoellle, line 32: I like iterators too, but in this case the procedural approach is a lot more readable:
```rs
    let mut residuals = Vec::new();
    for measurement in measurements {
        let vec: Vec<f32> = Residuals::calculate_from(parameters, measurement, field_dimensions)
            .ok()?
            .into();
        residuals.extend(vec.iter());
    }

    Some(DVector::from_vec(residuals));
```

crates/types/src/line.rs
```diff
@@ -95,10 +106,14 @@ impl<const DIMENSION: usize> Line<DIMENSION> {
         self.squared_distance_to_point(point).sqrt()
     }
 
-    pub fn is_orthogonal(&self, other: &Line<DIMENSION>, epsilon: f32) -> bool {
+    pub fn angle_diff_to_orthogonal(&self, other: &Line<DIMENSION>) -> f32 {
         let self_direction = (self.1 - self.0).normalize();
         let other_direction = (other.1 - other.0).normalize();
-        (self_direction.dot(&other_direction).acos().abs() - PI / 2.0).abs() < epsilon
+        (self_direction.dot(&other_direction).acos().abs() - PI / 2.0).abs()
```
knoellle, line 112: We use the premade constants pretty much everywhere else.
```suggestion
        (self_direction.dot(&other_direction).acos().abs() - FRAC_PI_2).abs()
```

crates/calibration/src/residuals.rs
```diff
@@ -0,0 +1,107 @@
...
+        let distance_between_parallel_line_center_points = projected_lines
+            .border_line
+            .distance_to_point(projected_lines.goal_box_line.center());
+        let distance_between_parallel_line_end_points = projected_lines
+            .border_line
+            .distance_to_point(projected_lines.goal_box_line.1);
+
+        Ok(Residuals {
+            border_to_connecting_angle: border_to_connecting_angle - FRAC_PI_2,
+            connecting_to_goal_box_angle: connecting_to_goal_box_angle - FRAC_PI_2,
```
knoellle, line 80: Are you sure you know what you are doing?
Manually subtracting angles from each other is dangerous, although I couldn't find any obvious issues in this case.

knoellle (changes requested):

crates/calibration/src/residuals.rs
h3ndrk: Iirc, I'm using the new angle method that always returns a positive angle between [0,180] degrees. Subtracting 90 degrees transforms the interval into the expected one and gives us a sign for the gradient. I think this should work. 😅

crates/calibration/src/residuals.rs
jonathan-hellwig: I agree with @knoellle  here

crates/calibration/src/residuals.rs
```diff
@@ -0,0 +1,107 @@
...
+            distance_between_parallel_line_start_points: distance_between_parallel_line_start_points
+                - field_dimensions.goal_box_area_length,
+            distance_between_parallel_line_center_points:
+                distance_between_parallel_line_center_points - field_dimensions.goal_box_area_length,
+            distance_between_parallel_line_end_points: distance_between_parallel_line_end_points
+                - field_dimensions.goal_box_area_length,
```
jonathan-hellwig, line 86: The residuals here are measured in different units and ranges. You might want to normalize them or add different weights for each one of them.

crates/types/src/line.rs
```diff
@@ -12,6 +12,17 @@ pub struct Line<const DIMENSION: usize>(pub Point<f32, DIMENSION>, pub Point<f32
 pub type Line2 = Line<2>;
 
 impl Line2 {
+    pub fn signed_angle(&self, other: Self) -> f32 {
+        (self.1 - self.0)
+            .normalize()
+            .dot(&(other.1 - other.0).normalize())
+            .acos()
+    }
```
jonathan-hellwig, line 20: You are not using this function, right? I think you should remove it.

crates/types/src/line.rs
jonathan-hellwig, line 112: You can also check whether two vectors are orthogonal by calculating their dot product and checking if the result is zero. That way you do not need to calculate the angle using `acos`.

crates/calibration/src/residuals.rs
tuxbotix: What I recall from some papers about the solver (Levenberg-Marquardt) is that it is relatively immune to this issue, the ranges of each element of residual vector does not have to be  the same. As I remember the reason is the way it calculates the relative reduction of error - not as a norm or sum but it is done element-wise.

However, as you suggested, weights or some other influencing factors can be introduced to Influence which parameter gets priority, then normalizing would probably help to tune that.

h3ndrk: Ok, I will have a look once I conduct more in-depth testing with this calibration problem. I would like to keep it as is for now.

crates/calibration/src/lines.rs
```diff
@@ -0,0 +1,57 @@
...
+    point: Point2<f32>,
+    which: &str,
+) -> Result<Point2<f32>, LinesError> {
+    matrix
+        .pixel_to_ground(point)
+        .map_err(|source| LinesError::NotProjected {
+            source,
+            which: which.to_string(),
+        })
+}
```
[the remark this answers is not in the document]
h3ndrk: Done

crates/calibration/src/problem.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl CalibrationProblem {
+    pub fn new(
+        initial_corrections: Corrections,
+        measurements: &[Measurement],
+        field_dimensions: FieldDimensions,
+    ) -> Self {
+        Self {
+            parameters: initial_corrections,
+            measurements: measurements.to_vec(),
```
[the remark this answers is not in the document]
h3ndrk: Done

crates/calibration/src/residuals.rs
h3ndrk: Used and improved your suggestion.

crates/types/src/line.rs
h3ndrk: Yep, changed.

crates/types/src/line.rs
h3ndrk: Removed

crates/types/src/line.rs
h3ndrk: Now we have code duplication again but it now does not use `acos()` anymore.

knoellle: Careful! You just changed the meaning of the epsilon parameter.
If I see this correctly, you don't even use the `is_orthogonal` function in this PR. It is only used in the field border detection, where an `angle_threshold` parameter is passed for this epsilon.

Furthermore, this implementation isn't even correct.
You just check whether the dot product is less than epsilon without checking for the sign.
If the two vectors point in opposite directions, the normalized dot product would be `-1.0` and the `is_orthogonal` check would return true if we assume epsilon to be some small positive value.

I would suggest reverting this change and renaming the epsilon parameter to something more meaningful.

h3ndrk: Reverted, now we also don't have the code duplication anymore. Decided against renaming `epsilon`. Should be clear enough. The function also has been renamed to express the absolute return value.

jonathan-hellwig: I am fine with this version.

crates/types/src/line.rs
```diff
@@ -12,6 +12,10 @@ pub struct Line<const DIMENSION: usize>(pub Point<f32, DIMENSION>, pub Point<f32
 pub type Line2 = Line<2>;
 
 impl Line2 {
+    pub fn angle(&self, other: Self) -> f32 {
+        (self.1 - self.0).angle(&(other.1 - other.0))
+    }
```
h3ndrk, line 17: This needs to be reworked. Coming soon.

h3ndrk: Done

knoellle (approved):
LGTM

PR #384, opened 2023-06-18
Visual referee challenge

PatrickGoettsch:
## Introduced Changes

This PR introduces the basic framework or the Visual Referee Challenge. At the moment it generates a random gesture and sends it in case of specific state transitions from Set to play and from playing to ready or finished.

Implement gesture detection

## How to Test

test on Nao in GC. Use the GC branch julianschuler/gamecontroller3. It outputs the message in case it is heard. Test it with multiple robots and multiple transitions which require a whistle.

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -138,7 +138,8 @@ impl RoleAssignment {
         };
 
         if send_game_controller_return_message {
-            self.last_system_time_transmitted_game_controller_return_message = Some(cycle_start_time);
+            self.last_system_time_transmitted_game_controller_return_message =
+                Some(cycle_start_time);
             context
                 .hardware
                 .write_to_network(OutgoingMessage::GameController(
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -19,7 +19,7 @@ use crate::localization::generate_initial_pose;
 
 pub struct RoleAssignment {
     last_received_spl_striker_message: Option<SystemTime>,
-    last_transmitted_game_controller_return_message: Option<SystemTime>,
+    last_system_time_transmitted_game_controller_return_message: Option<SystemTime>,
     last_transmitted_spl_striker_message: Option<SystemTime>,
     role: Role,
     role_initialized: bool,
@@ -65,7 +65,7 @@ impl RoleAssignment {
     pub fn new(_context: CreationContext) -> Result<Self> {
         Ok(Self {
             last_received_spl_striker_message: None,
-            last_transmitted_game_controller_return_message: None,
+            last_system_time_transmitted_game_controller_return_message: None,
             last_transmitted_spl_striker_message: None,
             role: Role::Striker,
             role_initialized: false,
@@ -109,10 +109,10 @@ impl RoleAssignment {
         }
 
         let send_game_controller_return_message = self
-            .last_transmitted_game_controller_return_message
+            .last_system_time_transmitted_game_controller_return_message
             .is_none()
             || cycle_start_time.duration_since(
-                self.last_transmitted_game_controller_return_message
+                self.last_system_time_transmitted_game_controller_return_message
                     .unwrap(),
             )? > context.spl_network.game_controller_return_message_interval;
 
@@ -138,7 +138,7 @@ impl RoleAssignment {
         };
 
         if send_game_controller_return_message {
-            self.last_transmitted_game_controller_return_message = Some(cycle_start_time);
+            self.last_system_time_transmitted_game_controller_return_message = Some(cycle_start_time);
             context
                 .hardware
                 .write_to_network(OutgoingMessage::GameController(
--- a/crates/control/src/visual_referee_filter.rs
+++ b/crates/control/src/visual_referee_filter.rs
@@ -27,6 +27,11 @@ pub struct CycleContext {
     pub hardware: HardwareInterface,
 }
 
+
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {}
+
 impl VisualRefereeFilter {
     pub fn new(_context: CreationContext) -> Result<Self> {
         Ok(Self {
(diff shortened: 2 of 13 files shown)


crates/control/src/visual_referee_filter.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub filtered_whistle: Input<FilteredWhistle, "filtered_whistle">,
+    pub player_number: Parameter<PlayerNumber, "player_number">,
+
+    pub hardware: HardwareInterface,
+}
+
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {
+    pub last_primary_state: MainOutput<Option<PrimaryState>>,
```
h3ndrk, line 36: Remove? This node should not have any output.

crates/control/src/role_assignment.rs
```diff
@@ -109,10 +109,10 @@ impl RoleAssignment {
         }
 
         let send_game_controller_return_message = self
-            .last_transmitted_game_controller_return_message
+            .last_system_time_transmitted_game_controller_return_message
```
h3ndrk, line 112: In general there is no benefit of encoding type information into variables names (system time). We should configure our IDEs correctly.

```suggestion
            .last_transmitted_game_controller_return_message
```

crates/control/src/visual_referee_filter.rs
```diff
@@ -0,0 +1,85 @@
...
+                    PrimaryState::Finished | PrimaryState::Ready
+                )
+        );
+        self.last_primary_state = *context.primary_state;
+
+        // A random visual referee decision
+        let mut rng = thread_rng();
+        let gesture = VisualRefereeDecision::from_u32(rng.gen_range(1..=13)).unwrap();
+
+        if send_game_controller_vr_return_message {
```
h3ndrk, line 61: Please don't abbreviate:

```suggestion
        if send_game_controller_visual_referee_return_message {
```

Maybe even `send_visual_referee_message`.

crates/spl_network/src/endpoint.rs
```diff
@@ -116,6 +116,25 @@ impl Endpoint {
                     warn!("Failed to send UDP datagram via SPL socket: {error:?}")
                 }
             }
+            OutgoingMessage::VisualReferee(message) => {
```
h3ndrk, line 119: Idea for deduplication with the `OutgoingMessage::GameController`: You could keep the two match-case-arms and in each convert the message very early into `Vec<u8>`.  The rest could be refactored into a shared function to reduce code duplication.

crates/control/src/visual_referee_filter.rs
```diff
@@ -0,0 +1,85 @@
...
+    hardware::Interface, messages::OutgoingMessage, CycleTime, FilteredWhistle, PrimaryState,
+};
+
+pub struct VisualRefereeFilter {
+    last_primary_state: PrimaryState,
+}
+
+#[context]
+pub struct CreationContext {
+    pub player_number: Parameter<PlayerNumber, "player_number">,
```
h3ndrk, line 20: Remove? You are not using it in your `new()`.

h3ndrk (changes requested):
Looks good in general. I left a few comments. Thanks for your contribution 🙏

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: Okay.  I thought I must have this.

crates/control/src/role_assignment.rs
PatrickGoettsch: This is not meant as type but more changing the meaning what is in the variable as this does not contain the message itself but the timestamp of it. And even without having to use an IDE it is benefitial to have this information, IMHO

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: OK. Will change it

crates/control/src/visual_referee_filter.rs
h3ndrk: Please keep the empty `MainOutputs` struct, only remove the field from it.

crates/control/src/visual_referee_filter.rs
```diff
@@ -0,0 +1,78 @@
...
+            }
+            let message = OutgoingMessage::VisualReferee(VisualRefereeMessage {
+                player_number: *context.player_number,
+                gesture,
+                whistle_age: duration_since_last_whistle,
+            });
+            let _ = context
+                .hardware
+                .write_to_network(message)
+                .wrap_err("failed to write VisualRefereeMessage to hardware");
```
h3ndrk, line 75: Please return the error instead of consuming it into void:

```suggestion
            context
                .hardware
                .write_to_network(message)
                .wrap_err("failed to write VisualRefereeMessage to hardware")?;
```

crates/control/src/visual_referee_filter.rs
h3ndrk: Thanks but since you also used the name "game controller visual referee message" in the networking code (for the name of the deduplicated function) I would favor `send_visual_referee_message`. This is accurate and short.

crates/spl_network/src/endpoint.rs
```diff
@@ -116,8 +102,31 @@ impl Endpoint {
...
+                let message: Vec<u8> = message.into();
+                self.send_game_controller_visual_referee_message(message)
+                    .await;
+            }
         };
     }
+    async fn send_game_controller_visual_referee_message(&self, message: Vec<u8>) {
```
h3ndrk, line 112: Formatting nit-pick: Please have an empty line between functions.

h3ndrk (changes requested):

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: Done

crates/spl_network/src/endpoint.rs
PatrickGoettsch: Done

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: Done

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: Done

h3ndrk (approved):
Thanks for your contribution! 🎉

I will approve this now, but let's test this in the next test game and merge afterwards.

PR #385, opened 2023-06-19
Replace ping utility

### LINKED ISSUE #383: Pepsi upload ping bug
The command pepsi upload number is failing directly not to be able to ping the robot.

Removing line 28 in lib.rs make it work

okiwi6:
## Introduced Changes

This PR replaces the `surge_ping` crate by the use of the system binary `ping`.
This is because the `surge_ping` crate cannot create a RAW_SOCKET or an ICMP socket on some devices without root privileges.
On many linux distributions (but not all) the setting `net.ipv4.ping_group_range` includes all groups, which as far as I understand allows unprivileged ICMP traffic for all groups.

Therefore we can either
* Use the system ping library (as it has special privileges when `net.ipv4.ping_group_range` does not include all groups
* Force all users of pepsi to allow `pepsi` to  [execute pings](https://superuser.com/questions/1702163/ping-socket-operation-not-permitted-ubuntu-docker-image)

This PR goes the easy way of just using the system ping utility.

Fixes #383 

## ToDo / Known Issues

## How to Test

* So far only @PatrickGoettsch has been affected by this problem. This branch works on his machine but it is hard to test this in general.


julianschuler (changes requested):
I agree, using the system ping binary is the easiest option. I only have some small remarks left.

julianschuler (dismissed):
LGTM!

julianschuler (approved):
LGTM!

PR #386, opened 2023-06-20
Time to reach kick position

### LINKED ISSUE #297: Implement time to reach position function
what the title says.
Possible use cases:
- better striker role assignment decisions
- When unable to return to own half of the field in time after opponent scoring a goal, walk to center circle instead to prevent penalty

MaikRe:
## Introduced Changes

Calculate path to a kick position and use total length of that path as time to reach ball which replaces using the euclidean distance for striker selection.

Adds persistent state to behavior sim.

Fixes #297

## ToDo / Known Issues
~Behavior simulator breaks with this as there's currently no persistent state implemented, not sure how to fix currently.~

Behavior simulator has no ball filter, if ball is out of fov striker immediately becomes loser.

Function should also take into account robots fallen state as well as walking speed for different types of motion.

~After reframework #182 it *may* be possible to get rid of using the persistent state.~

## How to Test

Striker choosing behavior should appear different from current code. If a robot has to walk towards own goal to get behind the ball that path may now be longer than a robot that can just walk forwards in order to reach the ball, leading to more efficient striker selection.


MaikRe:
We currently have a lot more oscillation than normal. I think its because when the striker looks at a point of interest he immediately stops seeing the ball and becomes loser prompting all other robots to assume they are striker.

MaikRe:
When the striker stops looking at the ball he immediately sets time to reach to none thus opening the door for anyone to become striker. In the past since each robot had information to calculate the distance for both himself and the striker that the message came from this did not occur. It may be necessary to slowly increase time to reach after becoming striker to give a cycle or two to receive a message that the ball is still close by

crates/spl_network/src/endpoint.rs
```diff
@@ -68,7 +68,7 @@ impl Endpoint {
                 },
                 result = self.spl_socket.recv_from(&mut spl_buffer) => {
                     let (received_bytes, _address) = result.map_err(Error::ReadError)?;
-                    match spl_buffer[0..received_bytes].try_into() {
+                    match bincode::deserialize(&spl_buffer[0..received_bytes]) {
```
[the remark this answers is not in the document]
MaikRe: @Narcha ?

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -0,0 +1,121 @@
...
+            .as_ref()
+            .map(|path| {
+                path.iter()
+                    .map(|segment: &PathSegment| segment.length())
+                    .sum()
+            })
+            .map(Duration::from_secs_f32);
+
+        *context.time_to_reach_kick_position =
+            time_to_reach_kick_position.unwrap_or(Duration::from_secs(1900));
```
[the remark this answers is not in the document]
MaikRe: Done

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub parameters: Parameter<Behavior, "behavior">,
+    pub path_obstacles: AdditionalOutput<Vec<PathObstacle>, "time_to_reach_obstacles">,
+}
+
+#[context]
+pub struct CreationContext {}
+
+#[context]
+pub struct MainOutputs {
+    pub time_to_reach_kick_position: MainOutput<Option<Duration>>,
```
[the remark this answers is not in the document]
MaikRe: No, removed, may need to be added as additional output if desired

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -0,0 +1,121 @@
...
+        let is_near_ball = matches!(
+            world_state.ball,
+            Some(ball) if ball.ball_in_ground.coords.norm() < parameters.ignore_robot_when_near_ball_radius,
+        );
+        let obstacles = if is_near_ball {
+            &[]
+        } else {
+            world_state.obstacles.as_slice()
+        };
+
```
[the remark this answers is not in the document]
MaikRe: Done

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -0,0 +1,121 @@
...
+        world_state: &WorldState,
+        configuration: &PathPlanning,
+        parameters: &Dribbling,
+        path_obstacles_output: &mut AdditionalOutput<Vec<PathObstacle>>,
+        field_dimensions: &FieldDimensions,
+    ) -> Option<Vec<PathSegment>> {
+        let walk_path_planner =
+            WalkPathPlanner::new(field_dimensions, &world_state.obstacles, configuration);
+        let kick_decisions = world_state.kick_decisions.as_ref()?;
+        let best_kick_decision = match kick_decisions.first() {
```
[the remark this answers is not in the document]
MaikRe: Dribble path planning has been moved to separate node and is ready for review!

crates/control/src/dribble_path_plan.rs
```diff
@@ -0,0 +1,98 @@
+use color_eyre::Result;
```
okiwi6, line 1: maybe prefer `dribble_path_planner.rs`?

crates/control/src/dribble_path_plan.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::f32::consts::PI;
+use types::{
+    configuration::Behavior, FieldDimensions, GameControllerState, PathObstacle, PathSegment,
+    WorldState,
+};
+
+use crate::behavior::walk_to_pose::WalkPathPlanner;
+
+#[context]
+
```
okiwi6, line 15: remove newline here

crates/control/src/dribble_path_plan.rs
```diff
@@ -0,0 +1,98 @@
...
+pub struct CreationContext {}
+
+#[context]
+pub struct CycleContext {
+    pub world_state: Input<WorldState, "world_state">,
+    pub field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
+    pub configuration: Parameter<Behavior, "behavior">,
```
okiwi6, line 22: Please sort the members within a group alphabetically

crates/control/src/dribble_path_plan.rs
```diff
@@ -0,0 +1,98 @@
...
+        Ok(Self {})
+    }
+
+    pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
+        let world_state = context.world_state;
+        let field_dimensions = context.field_dimensions;
+        let configuration = &context.configuration.path_planning;
+        let parameters = &context.configuration.dribbling;
+        let path_obstacles_output = &mut context.path_obstacles;
+        let walk_path_planner =
```
okiwi6, line 44: maybe add some newlines to visually group together reading parameters, creating path planner, doing sth. with kickdecisions, etc.

crates/control/src/role_assignment.rs
```diff
@@ -254,6 +257,9 @@ impl RoleAssignment {
                                     robot_to_field,
                                     cycle_start_time,
                                 ),
```
okiwi6, line 259: If i see this correctly here, only the ball_position is different in the if-else branch. If that's the case I'd prefer to only compute the ball_position in the if-else and do the network message after that.

crates/spl_network/src/endpoint.rs
```diff
@@ -90,7 +90,7 @@ impl Endpoint {
                     .await;
             }
             OutgoingMessage::Spl(message) => {
-                let message: Vec<u8> = message.into();
+                let message: Vec<u8> = bincode::serialize(&message).unwrap();
```
okiwi6, line 93: When does this unwrap fail? Is it better to use an expect here with an error message?

okiwi6 (changes requested):
Separating the dribble path planning is much better than before, just some more comments :)

crates/spl_network/src/endpoint.rs
MaikRe: @Narcha

crates/control/src/dribble_path_plan.rs
MaikRe: Yes, has been renamed

crates/control/src/dribble_path_plan.rs
MaikRe: done

crates/control/src/dribble_path_plan.rs
MaikRe: done

crates/control/src/dribble_path_plan.rs
MaikRe: Done!

crates/control/src/role_assignment.rs
MaikRe: Yes that's true, I've rearranged it like suggested

crates/spl_network/src/endpoint.rs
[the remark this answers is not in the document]
MaikRe: replaced with serde json

crates/spl_network/src/endpoint.rs
MaikRe: also replaced with serde json

crates/control/src/role_assignment.rs
okiwi6: Looks much better, thanks

crates/spl_network/src/endpoint.rs
okiwi6: We decided now to use `bincode` instead of `serde_json` because we are only allowed to use 128 bytes. Please convert the `unwrap` to an `except` with an appropriate error message. Also it would be a nice to write a test that asserts the size of the struct to be less or equal to 128 Bytes to prevent future additions of fields to break the rules.

MaikRe: @Narcha  and I have changed this back to bincode and used a match statement to properly expect the error per @h3ndrk suggestions. Please feel free to comment if it should be different

okiwi6 (approved):
Nice!

PR #388, opened 2023-06-21
Remote compilation

knoellle:
## Introduced Changes

Yes, its a hack, but pending a better distributed/remote compilation solution this has the potential to decrease build times by an order of magnitude.

Fixes #

## ToDo / Known Issues

- [x] The git shenanigans currently unstage staged but uncommitted changes. ~~Not sure if there is a way around this.~~
  - [x] ~~Use `git diff` / `git apply` and send a patch instead of gwip / gunwip~~
  - [x] ... or just rsync all uncommited changes
- [x] Move return channel from script to pepsi.
  - This would allow more selectively rsyncing compile artifacts because we know target and profile
  - Also prevents rsyncing artifacts when running commands that don't produce any directly via the script
- [x] Move checkout operation from post-update hook on the remote to script

## How to Test

clone bare repo to `path/to/repo.git` on the remote host. 
```sh
git remote add compiler user@host:path/to/repo.git
./pepsi upload <number> --remote
```

### DIFF
--- a/scripts/remote
+++ b/scripts/remote
@@ -67,23 +67,23 @@ remotePath="$(echo $remoteGitPath | sed 's/\.git/\//')"
 branch=remoteCompile
 git push $remote HEAD:$branch --force --no-verify
 # check out pushed branch on the remote
-ssh $address "sh -c \"\
+ssh $address "sh -c \" \
     cd "$remotePath" \
     && git checkout --force \\\$(git rev-parse $branch) \
     && git clean -d --force \
 \""
 # send all changes that aren't uncommitted yet
-git status -s | cut -c 4- | rsync -a --info=progress --delete-missing-args --files-from=- . $remoteURL
+git status -s | cut -c 4- | rsync -a --info=progress --delete-missing-args --files-from=- . "$remoteURL"
 
 # invoke compile script remotely
 ssh $address "sh -c \" \
     . '$HOME/.profile' \
-    && cd "$remotePath" \
-    && $@\" \
-"
+    && cd \"$remotePath\" \
+    && $@
+\""
 
 # fetch results
 echo Returning files
 printf '  %s\n' "${files[@]}"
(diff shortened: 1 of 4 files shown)


tools/pepsi/src/cargo.rs
```diff
@@ -17,8 +18,11 @@ pub struct Arguments {
     /// Pass through arguments to cargo ... -- PASSTHROUGH_ARGUMENTS
     #[arg(last = true, value_parser)]
     pub passthrough_arguments: Vec<String>,
+    #[arg(long)]
```
julianschuler, line 21: Please add documentation comments like for the other flags.

tools/pepsi/src/cargo.rs
```diff
@@ -27,6 +31,54 @@ pub enum Command {
 }
 
 pub async fn cargo(arguments: Arguments, repository: &Repository, command: Command) -> Result<()> {
+    if arguments.remote {
```
julianschuler, line 34: Please consider factoring this part out to a separate function.

julianschuler (changes requested):
I really like this feature!

tools/pepsi/src/cargo.rs
knoellle: Done

tools/pepsi/src/cargo.rs
knoellle: Done

scripts/remote
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
```
julianschuler, line 2: I think this script should exit on error. Consider adding the line
```
set -e
```
to enable this.

julianschuler (changes requested):
I noticed some funky behavior when testing with an incorrect remote, IMO making the script exit on error should be enabled.

PR #398, opened 2023-06-28
Prevent interception step in penalty shootout/ kick

JonasHolz:
## Introduced Changes

Checks for penalty shootout and penalty kick in intercept ball 

## ToDo / Known Issues

None

None

## How to Test

Try forcing intercepting steps in penalty shootout and penalty kicks

### DIFF
--- a/crates/control/src/behavior/intercept_ball.rs
+++ b/crates/control/src/behavior/intercept_ball.rs
@@ -1,19 +1,19 @@
 use nalgebra::{point, Point2, UnitComplex};
 use spl_network_messages::{GamePhase, SubState};
 use types::{
-    configuration::InterceptBall, GameControllerState, HeadMotion, LineSegment, MotionCommand,
-    OrientationMode, PathSegment, PrimaryState, WorldState,
+    configuration::InterceptBall, FilteredGameState, GameControllerState, HeadMotion, LineSegment,
+    MotionCommand, OrientationMode, PathSegment, WorldState,
 };
 
 pub fn execute(world_state: &WorldState, parameters: InterceptBall) -> Option<MotionCommand> {
     match (
-        world_state.robot.primary_state,
+        world_state.filtered_game_state,
         world_state.ball,
         world_state.robot.robot_to_field,
         world_state.game_controller_state,
     ) {
         (
-            PrimaryState::Playing,
+            Some(FilteredGameState::Playing { .. }),
             _,
             _,
             Some(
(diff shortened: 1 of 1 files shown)


crates/control/src/behavior/intercept_ball.rs
```diff
@@ -1,16 +1,39 @@
...
+                },
+            ),
+        ) => None,
+        (
+            Some(FilteredGameState::Playing {
+                ball_is_free: false,
+            }),
+            ..,
+        ) => None,
+        (Some(FilteredGameState::Playing { .. }), Some(ball), Some(robot_to_field), _) => {
```
julianschuler, line 36: You can remove this case by specifying `Playing { ball_is_free: true }` in the following case instead:
```suggestion
        (Some(FilteredGameState::Playing { ball_is_free: true }), Some(ball), Some(robot_to_field), _) => {
```

crates/control/src/behavior/intercept_ball.rs
```diff
@@ -1,16 +1,39 @@
...
     match (
-        world_state.robot.primary_state,
+        world_state.filtered_game_state,
         world_state.ball,
         world_state.robot.robot_to_field,
+        world_state.game_controller_state,
     ) {
-        (PrimaryState::Playing, Some(ball), Some(robot_to_field)) => {
+        (
+            Some(FilteredGameState::Playing { .. }),
```
julianschuler, line 16: I think if we are in any of the two game phases we do not want interception behavior independent of the `FilteredGameState`. If so, you don't have to match on `Playing` here:
```suggestion
            _,
```

crates/control/src/behavior/intercept_ball.rs
```diff
@@ -1,16 +1,39 @@
...
+            Some(
+                GameControllerState {
+                    game_phase: GamePhase::PenaltyShootout { .. },
+                    ..
+                }
+                | GameControllerState {
+                    sub_state: Some(SubState::PenaltyKick),
+                    ..
+                },
+            ),
```
julianschuler, line 28: Since the `GameControllerState` is only used here, I would argue that this condition should be checked before the match statement separately and returning None if fulfilled. This keeps the remaining match statement smaller and more readable while allowing the first check to be expanded more easily in the future if needed.

julianschuler (changes requested):
Thanks for the important addition! However, I think the added code can be simplified quite a bit.

crates/control/src/behavior/intercept_ball.rs
MaikRe: Done

crates/control/src/behavior/intercept_ball.rs
MaikRe: Good catch, changed

crates/control/src/behavior/intercept_ball.rs
MaikRe: Changed

julianschuler (approved):
LGTM!

PR #416, opened 2023-07-03
Implement traits to convert YCbCr422Image to RGB

tuxbotix:
## Introduced Changes

The functionality to perform the conversion was already there but was not exposed in a reusable way.
I also fixed a logic error in the height calculation of `YCbCr422Image -> RgbImage` conversion.

*note*: ~`impl EncodeJpeg for YCbCr422Image` does not use the `YCbCr422Image::into()` as it requires a clone while `EncodeJpeg` only need a view. If the cloning overhead is considered negligible we can use the `into()` function.~ **Fixed**

Fixes #

Add unit tests for these functions!
We already found a logic error (of height calculation) during this PR.

### DIFF
--- a/crates/types/src/ycbcr422_image.rs
+++ b/crates/types/src/ycbcr422_image.rs
@@ -25,6 +25,25 @@ pub struct YCbCr422Image {
     buffer: Arc<Vec<YCbCr422>>,
 }
 
+impl From<RgbImage> for YCbCr422Image {
+    fn from(rgb_image: RgbImage) -> Self {
+        let buffer = Arc::new(buffer_422_from_rgb_image(rgb_image));
+        let width_422 = rgb_image.width() / 2;
+        let height = rgb_image.height() / 2;
+        Self {
+            width_422,
+            height,
+            buffer,
+        }
+    }
+}
+
+impl Into<RgbImage> for YCbCr422Image {
+    fn into(self) -> RgbImage {
+        rgb_image_from_buffer_422(self.width_422, self.height, &self.buffer)
+    }
+}
+
 impl EncodeJpeg for YCbCr422Image {
     const DEFAULT_QUALITY: u8 = 40;
     type Error = ImageError;
--- a/crates/types/src/ycbcr422_image.rs
+++ b/crates/types/src/ycbcr422_image.rs
@@ -27,23 +27,78 @@ pub struct YCbCr422Image {
 
 impl From<RgbImage> for YCbCr422Image {
     fn from(rgb_image: RgbImage) -> Self {
+        let width_422 = rgb_image.width() / 2;
+        let height = rgb_image.height();
+        let data = rgb_image
+            .into_vec()
+            .chunks(6)
+            .map(|pixel| {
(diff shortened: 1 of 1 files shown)


knoellle:
@h3ndrk what is the state here?

crates/types/src/ycbcr422_image.rs
```diff
@@ -25,12 +25,34 @@ pub struct YCbCr422Image {
...
+        Self {
+            width_422: rgb_image.width() / 2,
+            height: rgb_image.height(),
+            buffer: Arc::new(buffer_422_from_rgb_image(rgb_image)),
+        }
+    }
+}
+
+impl From<&YCbCr422Image> for RgbImage {
+    fn from(val: &YCbCr422Image) -> Self {
```
h3ndrk, line 39: Value? Vallejo (Amtrak station code; Vallejo, CA)? Valley? Valve? Valerie? Valid? Validation? Validate? Valine (Amino Acid)? Valedictorian? Véhicule Automatique Léger (French driverless metro system)? Viking Air Limited (Canada)? Vatican City Lira (defunct currency; ISO designation)? Value Added Logistics (supply chain)? Voice Activated Lightstand (photography)? Variable Angle Launcher? Vertical Alarm Limit? Value-Oriented Algorithmic Language?

Please don't abbreviate. Also in the method below.

What is it? An image, right? Why not name it like that? :wink:

crates/types/src/ycbcr422_image.rs
```diff
@@ -25,12 +25,34 @@ pub struct YCbCr422Image {
     buffer: Arc<Vec<YCbCr422>>,
 }
 
+impl From<RgbImage> for YCbCr422Image {
+    fn from(rgb_image: RgbImage) -> Self {
```
h3ndrk, line 29: What do you think about changing all uses of the `buffer_422_from_rgb_image()` and `rgb_image_from_buffer_422()` to uses of this trait? Maybe we could even inline the code of these functions into this trait to enforce usage of it? Currently there is some code duplication in the `impl YCbCr422Image` and this trait.

h3ndrk (changes requested):

crates/types/src/ycbcr422_image.rs
tuxbotix: Quite true :D but the list is epic xD I'll fix it.

crates/types/src/ycbcr422_image.rs
tuxbotix: Yeah I noted that too. I think i made some improvements in the edge detection branch, will check and patch this.

tuxbotix: Done: After looking at the buffer from/ to functions, I also noticed it is better to directly use the trait instead. Inlined those functions into the trait.

crates/types/src/ycbcr422_image.rs
tuxbotix: Fixed.

h3ndrk (approved):

PR #427, opened 2023-07-04
Too early replacment keeper switch

### LINKED ISSUE #395: ReplacementKeeper leaves too early
When Keeper enters from being penalized the replacementkeeper immediately leaves the goal, which creates a several seconds long opening directly to the goal.

JonasHolz:
## Introduced Changes

Prevents to early Keeper switches
Improves replacmentkeeper_test.lua file

Fixes #395 

## ToDo / Known Issues

None

None

## How to Test

Penalize the keeper and unpenalize him afterwards and check if the switch is delayed
The upgraded lua file can be used for this testing


tests/behavior/replacementkeeper_test.lua
```diff
@@ -67,6 +67,25 @@ function on_cycle()
...
+            velocity = { 0.0, 0.0 },
+        }
+    end
+
+    if state.cycle_count == 3000 then
+        state.ball = {
+            position = { -4.0, 0.0 },
+            velocity = { 0.0, 0.0 },
+        }
+    end
```
MaikRe, line 87: Why did you add these ball position teleports?

etc/parameters/default.json
```diff
@@ -328,7 +328,8 @@
     "goal_post_obstacle_radius": 0.2
   },
   "role_assignment": {
-    "forced_role": null
+    "forced_role": null,
+    "keeper_replacementkeeper_switch_time": { "nanos": 0, "secs":  8}
```
MaikRe, line 332: Based on the behavior sim file this should at least be 12, I recommend changing it to that and then doing a real life test.

MaikRe (changes requested):

tests/behavior/replacementkeeper_test.lua
JonasHolz: In order to check if the replacmentkeeper can still become striker and is not forced to stay replacmentkeeper until the keeper is back in the goal.

etc/parameters/default.json
JonasHolz: Done

tests/behavior/replacementkeeper_test.lua
MaikRe: What do you think about 
```lua

    if state.cycle_count == 2150 then
        unpenalize(1);
        state.game_controller_state.penalties.one = nil;
    end

    if state.cycle_count == 3550 then
        state.ball = {
            position = { 0.0, 0.0 },
            velocity = { 0.0, 0.0 },
        }
    end

    if state.cycle_count == 3600 then
        penalize(1);
        state.game_controller_state.penalties.one = {
            Manual = {
                remaining = {
                    nanos = 0,
                    secs = 5
                },
            }
        };
        set_robot_pose(1, { -3.2, 3 }, -1.5707963267948966);
    end

    if state.cycle_count == 3900 then
        unpenalize(1);
        state.game_controller_state.penalties.one = nil;
    end

    if state.cycle_count == 4650 then
        state.ball = {
            position = { -4.0, 0.0 },
            velocity = { 0.0, 0.0 },
        }
    end
```

As this would test both situations within the same file by penalizing the striker twice?

JonasHolz: This indeed looks like a better version, therefore I changed the file to this.

MaikRe (approved):
LGTM

PR #440, opened 2023-07-05
Add SPL_HULKs support

Narcha:
## Introduced Changes

Bump OS version to 5.8.1 and add support for the SPL_HULKs access point.

## How to Test

`./pepsi gammaray $NAO_NUMBER && ./pepsi wireless set SPL_HULKs $NAO_NUMBER`

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -301,6 +301,7 @@ impl Nao {
             Network::SplD,
             Network::SplE,
             Network::SplF,
+            Network::SplHulks,
         ]
         .into_iter()
         .map(|possible_network| {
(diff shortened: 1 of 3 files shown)


crates/nao/src/lib.rs
```diff
@@ -373,6 +373,7 @@ pub enum Network {
     SplD,
     SplE,
     SplF,
+    SplHulks,
```
julianschuler, line 376: Please add this also to command_string in set_network.

julianschuler (changes requested):
One minor change, the rest LGTM.

crates/nao/src/lib.rs
Narcha: done

julianschuler (approved):

PR #445, opened 2023-07-05
Twix panel for Image Segmenter and Field Colour calibration

tuxbotix:
## Introduced Changes

Introduces a rather hacky panel similar to the Manual (extrinsic) calibration with sliders to speed up Image Segmenter and Field Colour tuning.

Fixes #

## ToDo / Known Issues

- [x] Verify the ranges set for the sliders are valid.
- [x] Remove degree symbol in the label showing ranges 
- [x] Remove fields that show up in the diff (f32 -> f64 -> f32 reasons) which aren't actually modified in this panel

## How to Test

Choose `Segmenter Calibration` in the Twix panel. It should appear like below.

image


knoellle (approved):
LGTM

PR #446, opened 2023-07-05
Implement interpolated parameters

h3ndrk:
## Introduced Changes

Allows to interpolate parameters in the field color based on the X-component and the orientation angle.

## ToDo / Known Issues

None

- Fancier interpolation

## How to Test

Configure different parameters and walk over the field to observe slow changes in the behavior of the robot.


crates/types/src/interpolated.rs
```diff
@@ -0,0 +1,68 @@
...
+use serde::{Deserialize, Serialize};
+use serialize_hierarchy::SerializeHierarchy;
+
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize, SerializeHierarchy)]
+pub struct Interpolated {
+    pub first_half_own_half_towards_own_goal: f32,
+    pub first_half_own_half_away_own_goal: f32,
+    pub first_half_opponent_half_towards_own_goal: f32,
+    pub first_half_opponent_half_away_own_goal: f32,
+}
```
knoellle, line 11: I don't like the name `Interpolated`, but let's kloppen about that after the game.

knoellle (changes requested):

crates/types/src/interpolated.rs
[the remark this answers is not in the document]
h3ndrk: Sure

crates/types/src/interpolated.rs
h3ndrk: Sure, which year?

crates/spl_network_messages/src/game_controller_state_message.rs
```diff
@@ -80,6 +81,7 @@ impl TryFrom<RoboCupGameControlData> for GameControllerStateMessage {
                     message.teams[1]
                 ),
             };
+        dbg!((hulks_team_index, opponent_team_index));
```
schmidma, line 84: Here is a print

h3ndrk: Should be removed

crates/types/src/interpolated.rs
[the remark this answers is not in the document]
h3ndrk: Test cases added

crates/types/src/field_color.rs
```diff
@@ -21,7 +21,7 @@ impl FieldColor {
         if red_chromaticity > self.red_chromaticity_threshold
             || blue_chromaticity > self.blue_chromaticity_threshold
             || green_chromaticity < self.lower_green_chromaticity_threshold
-            || rgb.g < self.green_luminance_threshold
+            || (rgb.g as f32) < self.green_luminance_threshold
```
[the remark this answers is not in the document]
h3ndrk: Because `self.green_luminance_threshold` is now a `f32` and not an `u8` anymore. Therefore I thought we could compare in `f32`. We could also compare in `u8` but does that matter?

crates/control/src/game_controller_filter.rs
```diff
@@ -57,6 +57,7 @@ impl GameControllerFilter {
             self.game_controller_state = Some(GameControllerState {
                 game_state: game_controller_state_message.game_state,
                 game_phase: game_controller_state_message.game_phase,
+                half: game_controller_state_message.half,
```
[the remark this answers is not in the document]
h3ndrk: This code is outdated and so is this comment.

crates/vision/src/image_segmenter.rs
```diff
@@ -48,10 +58,20 @@ pub struct MainOutputs {
...
+        })
     }
 
     pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
+        let robot_to_field_of_home_after_coin_toss_before_second_half = context
+            .robot_to_field_of_home_after_coin_toss_before_second_half
+            .copied()
+            .unwrap_or(self.fallback_robot_to_field_of_home_after_coin_toss_before_second_half);
+        self.fallback_robot_to_field_of_home_after_coin_toss_before_second_half =
+            robot_to_field_of_home_after_coin_toss_before_second_half;
```
knoellle, line 73:
```rs
if let Some(x) = context.x {
    self.x = x;
}
```
might be easier to understand. Then just use `self.x` below, I don't see a need to have a separate local variable.

etc/parameters/field_b/default.json
```diff
@@ -1 +1,20 @@
...
+    "vision_bottom": {
+      "vertical_edge_threshold": {
+        "first_half_own_half_towards_own_goal": 32.0,
+        "first_half_own_half_away_own_goal": 32.0,
+        "first_half_opponent_half_towards_own_goal": 32.0,
+        "first_half_opponent_half_away_own_goal": 32.0
+      }
+    }
+  }
+}
```
knoellle, line 20: Isn't this the same as in the default.json?

etc/parameters/nao_location
```diff
@@ -1 +1 @@
-robocup/
\ No newline at end of file
+/home/h3ndrk/Documents/hulks/hulk2/etc/parameters/field_d
```
knoellle: #459

etc/parameters/field_b/default.json
h3ndrk: Yes, but this way, changes in the top-level `default.json` will not change the already tuned parameters in this location.

crates/types/src/field_color.rs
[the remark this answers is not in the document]
knoellle: Ah, I see. I was wondering why you changed the `self.green_luminance_threshold` type but I already found the reason. Resolving.

crates/vision/src/image_segmenter.rs
h3ndrk: Done

PR #447, opened 2023-07-05
Vrc2

PatrickGoettsch:
## Introduced Changes

Should send message first after 5 sec and only on state changes for kick off, goal or finish

Do any kind of detection

## How to Test

Test with julians gamecontroller3 branch and observe if only messages appear at kick-off, goal and finish and only after 8 seconds


crates/control/src/visual_referee_filter.rs
```diff
@@ -34,25 +38,38 @@ impl VisualRefereeFilter {
...
         self.last_primary_state = *context.primary_state;
 
-        // Initially a random visual referee decision
-        let mut rng = thread_rng();
-        let gesture = VisualRefereeDecision::from_u32(rng.gen_range(1..=13)).unwrap();
-
-        if send_visual_referee_message {
+        if self
+            .time_of_last_visual_referee_related_state_change
+            .map_or(false, |time_of_last_state_change| {
```
knoellle, line 63: You can use `is_some_and` here.

crates/control/src/visual_referee_filter.rs
```diff
@@ -65,8 +82,13 @@ impl VisualRefereeFilter {
                 })
                 .unwrap_or(Duration::from_secs(15));
             if duration_since_last_whistle.as_secs_f32() < 1.0 {
-                duration_since_last_whistle = Duration::from_secs(5)
+                duration_since_last_whistle = Duration::from_secs(8)
             }
```
knoellle, line 86: I'm not sure what this is supposed to do.
Can you find better names or leave a comment to explain why whistle detections younger than one second are replaced by ones that are eight seconds old, or why we send a whistle age of 15 seconds if we didn't hear any.

knoellle (changes requested):

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: I hope this is now solved ;-)

crates/control/src/visual_referee_filter.rs
PatrickGoettsch: Will be replaced by new numbers and states in a future PR, elaborating on this due to rules changes 2024.

PR #450, opened 2023-07-06
Add `check_neighboring_segments` parameter

rmburg:
## Introduced Changes

What the title says.
When enabled, this will discard line segments whose neighboring segments definitely do not have field color.

## ToDo / Known Issues

ToDo: Test

## How to Test

twix


oleflb:
what is the state here @rmburg?

PR #464, opened 2023-07-07
Adds new ear signs

PatrickGoettsch:
## Introduced Changes

This PR shows in the first place the whistle (as usual) and adds the charging status and in case there are no game controller message for 5 seconds, when a first message has arrived. It also shows the maximum temperature on the ears in percent.

Fixes #

## How to Test

Run this with game controller. Let it send message, stop it see if it blinks. See if the whistle works. Let the robot walk and see if the temperature increases.

### DIFF
--- a/crates/control/src/led_status.rs
+++ b/crates/control/src/led_status.rs
@@ -225,9 +225,10 @@ impl LedStatus {
         }
 
         let minimum_temperature = 30.0;
+        let maximum_temperature = 105.0;
 
         let relative_temperature = (current_maximum_temperature - minimum_temperature)
-            / (105.0 - minimum_temperature).floor();
+            / (maximum_temperature - minimum_temperature).floor();
 
         Ear::percentage_ears(1.0, relative_temperature)
     }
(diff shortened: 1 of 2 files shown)


etc/parameters/default.json
knoellle, line 1: Why does this PR change walking parameters?

crates/control/src/led_status.rs
```diff
@@ -171,6 +200,38 @@ impl LedStatus {
...
+                return Ear::full_ears(1.0);
+            } else {
+                return Ear::full_ears(0.0);
+            }
+        }
+
+        let minimum_temperature = 30.0;
+
+        let relative_temperature = (current_maximum_temperature - minimum_temperature)
+            / (105.0 - minimum_temperature).floor();
```
knoellle, line 230: You named the minimum temperature constant but not the maximum. Please do so.

knoellle (changes requested):
You seem to have a large squash commit in your branch that includes a lot of changes from other people. Please remove it.

etc/parameters/default.json
PatrickGoettsch: Done

crates/control/src/led_status.rs
PatrickGoettsch: Done

knoellle (approved):
LGTM

PR #465, opened 2023-07-07
Centercircle kickoff violation

MaikRe:
## Introduced Changes

Our rule obstacle composer had an impossible scenario in which the center circle obstacle was never created. This fixes it.

Fixes #

## ToDo / Known Issues

This may not  be the actual cause for the violation, but hopefully is a good fix.

## How to Test

Behavior simulator file is included, or in game.


julianschuler (approved):
LGTM, even though I don't know whether this fixes the center circle kickoff violation.

PR #468, opened 2023-07-07
Record localization

h3ndrk:
## Introduced Changes

Records localization inputs and outputs in log file on the robot. Will automatically downloaded by Pepsi.

## ToDo / Known Issues

None

- Configure path in `hardware.json`

## How to Test

Deploy on a NAO, activate it in the parameters, run over the field, download the logs and observe that there is a file written. We currently don't have a viewer/reader for the file.

### DIFF
--- a/crates/control/src/localization_recorder.rs
+++ b/crates/control/src/localization_recorder.rs
@@ -25,6 +25,8 @@ pub struct CreationContext {
 #[context]
 pub struct CycleContext {
     pub enable: Parameter<bool, "localization_recorder.enable">,
+    pub only_record_during_active_localization:
+        Parameter<bool, "localization_recorder.only_record_during_active_localization">,
 
     pub current_odometry_to_last_odometry:
         HistoricInput<Option<Isometry2<f32>>, "current_odometry_to_last_odometry?">,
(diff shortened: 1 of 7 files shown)


schmidma:
I think, in the long run, we should integrate such a recording into our framework.
But I agree, that in the meantime, we can easily provide a quick solution with a recording node.
Still, I think this is a general topic and we should extend this node to support recording any dependency, not only localization. In the end this may boil down to only change naming from `localization*` to `*`.
What do you think?

crates/control/src/localization_recorder.rs
```diff
@@ -0,0 +1,157 @@
...
+use bincode::serialize;
+use color_eyre::{eyre::Context, Result};
+use context_attribute::context;
+use framework::{HistoricInput, PerceptionInput};
+use nalgebra::Isometry2;
+use serde::{Deserialize, Serialize};
+use types::{GameControllerState, LineData, PrimaryState};
+
+pub struct LocalizationRecorder {
+    recording: Option<File>,
```
schmidma, line 17: We should use a buffered writer here. e.g. `std::io::BufWriter`

h3ndrk: Done

h3ndrk:
> I think, in the long run, we should integrate such a recording into our framework. But I agree, that in the meantime, we can easily provide a quick solution with a recording node.

Exactly that was my reasoning. We had a lot of problems with localization on the challanging field and I don't know why, therefore the recording. And since the framework does not support that currently, it is implemented in the robotics domain.

> Still, I think this is a general topic and we should extend this node to support recording any dependency, not only localization. In the end this may boil down to only change naming from `localization*` to `*`. What do you think?

I hope, if we need to record more data on RoboCup 2023, we can extend this node by changing the code manually. After RoboCup, we should implement that in the framework properly. We should have a design session before that to align the requirements before implementation.

I didn't get your naming suggestion... :thinking: :sweat_smile:

knoellle (approved):
LGTM

PR #480, opened 2023-07-07
Add Opponent Gamestate

okiwi6:
## Introduced Changes

Adds the GameState of the opponents, assuming both teams hear the whistle.
This will be used in order to examine if the ball is free or not for the opponent team which
allows us to do shots with higher precision.

Fixes #

## ToDo / Known Issues

## How to Test

* Start the game controller and examine the opponent game state, make sure the ball is free changes only for the kickoff team when a whistle is heard

### DIFF
--- a/crates/control/src/game_state_filter.rs
+++ b/crates/control/src/game_state_filter.rs
@@ -12,6 +12,7 @@ use types::{
 
 pub struct GameStateFilter {
     state: State,
+    opponent_state: State,
 }
 
 #[context]
@@ -42,12 +43,14 @@ pub struct CycleContext {
 #[derive(Default)]
 pub struct MainOutputs {
     pub filtered_game_state: MainOutput<Option<FilteredGameState>>,
+    pub filtered_opponent_game_state: MainOutput<Option<FilteredGameState>>,
 }
 
 impl GameStateFilter {
     pub fn new(_context: CreationContext) -> Result<Self> {
         Ok(Self {
             state: State::Initial,
+            opponent_state: State::Initial,
         })
     }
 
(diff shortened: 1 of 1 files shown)


crates/control/src/game_state_filter.rs
```diff
@@ -229,18 +250,55 @@ impl State {
...
+    fn construct_opponent_filtered_game_state(
+        &self,
+        game_controller_state: &GameControllerState,
+        cycle_start_time: SystemTime,
+        ball_detected_far_from_kick_off_point: bool,
+        config: &GameStateFilterParameters,
+    ) -> FilteredGameState {
+        let hulks_is_kicking_team = matches!(
+            game_controller_state.kicking_team,
+            Team::Hulks | Team::Uncertain
```
MaikRe, line 282: Shouldn't only one of these cases include the Team::Uncertain?

knoellle: This value comes from the game controller state message where it is set to either `Hulks or `Opponent` during parsing.
AFAIK the `Uncertain` state is only used in the filtered game state when a whistle is heard during play.
Since we have no reliable way to determine which team will be the kicking team, it is set to `Uncertain` until we receive the true value from the game controller a few seconds later.

It would probably be more correct to assert this here instead of coercing `Uncertain` to the opponent team having kickoff.

okiwi6: If I understand you correctly @knoellle, you mean the same thing as @MaikRe? Now replaced by only `Team::Hulks`

knoellle (approved):
LGTM

PR #486, opened 2023-07-07
keep ball obstacle radius if ball is between robot and goal

### LINKED ISSUE #479: Walk around ball: don't reduce ball obstacle radius when standing on the wrong side of the ball
In today's game vs. B-Human, our number 6 almost scored an own goal at 7:30 because it decided to do side steps toward the ball in front of our goal. @MaikRe suggested this was because it reduced the ball obstacle radius based on the angle to the ball, regardless of the relative position.

MaikRe:
## Introduced Changes

If the ball is between the robot and the goal based on field x coordinates the ball obstacle radius factor is kept at 1 in order to avoid own goals when doing side steps.

Fixes #479 

## ToDo / Known Issues

## How to Test

Must be tested in a scenario in which the robot chooses to use side steps to approach a ball that is on the own goal side of the robot. Ask me how to recreate this situation if needed.


julianschuler (approved):
LGTM!

PR #467, opened 2023-07-07
Network and joint temperature via aliveness

julianschuler:
## Introduced Changes

This PR exposes the current network and joint temperatures via aliveness.

Fixes #453 and fixes #470.

## ToDo / Known Issues

- [x] Update documentation

## Ideas for Next Iterations (Not This PR)

Maybe adding player number assignments when using pepsi aliveness

## How to Test

In HULKs repo:
```
. naosdk/5.8.0/environment-setup-corei7-64-aldebaran-linux
cd tools/aliveness
cargo build && scp target/x86_64-aldebaran-linux-gnu/debug/aliveness-service 10.1.24.36:
cd ../hula
cargo build && scp target/x86_64-aldebaran-linux-gnu/debug/hula 10.1.24.36:
pepsi shell 36
systemctl stop aliveness hula
./hula &
./aliveness-service
```

PR #490, opened 2023-07-08
Change the order of aliveness output

julianschuler:
## Introduced Changes

The new information available via aliveness introduced in #467 cause the output of `./pepsi aliveness` to be cluttered. This PR improves this by showing the network and temperature which are always displayed first and aligning their columns.

## ToDo / Known Issues

None.

A proper alignment for all columns in the future is desirable.

## How to Test

```
./pepsi aliveness
```


Narcha (approved):
LGTM. Let's implement proper alignment after Robocup.

PR #492, opened 2023-07-08
Penalty keeper tuning and extra parameters

### LINKED ISSUE #477: Penalty shootout behaviour is buggy/ weird
We did a quick test of the keeper; 
- the transition to "arms up" pose takes some time (since command from GC).
    - The head wobbles like in a very fast lookaround and do something weird.
- The motion is triggered *okay* under some conditions 
    - might be due to incorrect vision params for the field
- ~The side jumps are rather harsh and almost sounds like bones cracking, Nao Devils got us some tips~ -> #482 
- The jump is triggered just by a static ball , etc. It is way too sensntive.

TODOs:
- [x] check if we can tune the trigger happiness like the old C++ keeper
    - Old implementation for reference: https://github.com/HULKs/nao/blob/cpp/src/Brain/Behavior/PenaltyKeeperActionProvider.cpp
    - ^ Turns out that we already have something similar, but without checking any velocity stuff
- [ ] The current implementation relies on the position of the ball. this means it will jump for a static ball as well. Decide if we should use a minimum velocity to cause a trigger.
- ~see if we can find a way to improve the fanciness of kicking~ -> #483 
- [ ] add a way to disable jumping motions for testing (just do a look at)

tuxbotix:
## Introduced Changes

I tried to define some tuning possibilities with the observations in #477.

Also introduced a `side_jump_threshold` in order to not jump to the wrong direction by `ball_position.y >= 0.0` check alone (which caused a jump to the wrong direction in RC2023 penalty shootout against Runswift).

We used this branch (without the above-mentioned side jump threshold) and worked as expected -> less trigger-happy. 
- Discuss how we can do a test mode, extra motion (i.e. look at) or some other way like LEDs or audio.
- These parameters and the behaviour could be further tuned by more testing, as the keeper jumped too late in RC2023 (maybe the parameters or some other place is the reason).

## How to Test

The parameters reduce the trigger happiness by increasing the distance from the penalty marker the ball has to move.

### DIFF
--- a/crates/control/src/penalty_shot_direction_estimation.rs
+++ b/crates/control/src/penalty_shot_direction_estimation.rs
@@ -15,13 +15,21 @@ pub struct PenaltyShotDirectionEstimation {
 }
 
 #[context]
-pub struct CreationContext {}
+pub struct CreationContext {
+    pub field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
+    pub moving_distance_threshold:
+        Parameter<f32, "penalty_shot_direction_estimation.moving_distance_threshold">,
+    pub minimum_velocity_threshold:
+        Parameter<f32, "penalty_shot_direction_estimation.minimum_velocity_threshold">,
+}
 
 #[context]
 pub struct CycleContext {
     field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
     moving_distance_threshold:
         Parameter<f32, "penalty_shot_direction_estimation.moving_distance_threshold">,
+    pub minimum_velocity_threshold:
+        Parameter<f32, "penalty_shot_direction_estimation.minimum_velocity_threshold">,
 
     ball_position: RequiredInput<Option<BallPosition>, "ball_position?">,
     game_controller_state: RequiredInput<Option<GameControllerState>, "game_controller_state?">,
@@ -56,6 +64,8 @@ impl PenaltyShotDirectionEstimation {
                         - context.field_dimensions.penalty_marker_distance)
                         .abs()
                         > *context.moving_distance_threshold
+                        && context.ball_position.velocity.norm()
+                            > *context.minimum_velocity_threshold
                     {
                         if context.ball_position.position.y >= 0.0 {
                             self.last_shot_direction = PenaltyShotDirection::Left;
--- a/crates/control/src/penalty_shot_direction_estimation.rs
+++ b/crates/control/src/penalty_shot_direction_estimation.rs
@@ -62,7 +62,7 @@ impl PenaltyShotDirectionEstimation {
             }
             (PrimaryState::Playing, GamePhase::PenaltyShootout { .. }) => {
                 if let PenaltyShotDirection::NotMoving = self.last_shot_direction {
-                    let is_ball_position_exceeed_moving_distance_threshold =
+                    let is_ball_position_exceeding_moving_distance_threshold =
                         (context.ball_position.position.x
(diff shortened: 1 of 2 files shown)


crates/control/src/penalty_shot_direction_estimation.rs
```diff
@@ -52,10 +58,21 @@ impl PenaltyShotDirectionEstimation {
             }
             (PrimaryState::Playing, GamePhase::PenaltyShootout { .. }) => {
                 if let PenaltyShotDirection::NotMoving = self.last_shot_direction {
-                    if (context.ball_position.position.x
-                        - context.field_dimensions.penalty_marker_distance)
-                        .abs()
-                        > *context.moving_distance_threshold
+                    let is_ball_position_exceeed_moving_distance_threshold =
```
h3ndrk, line 61: No neeeeeed 😉

```suggestion
                    let is_ball_position_exceeding_moving_distance_threshold =
```

crates/control/src/penalty_shot_direction_estimation.rs
h3ndrk, line 1: Are we still reacting fast enough if you reduced trigger happiness?

h3ndrk (changes requested):

crates/control/src/penalty_shot_direction_estimation.rs
tuxbotix: Will fix it :P 
I split to parameters as there were multiple `&&` parts making it a bit hard to read.

crates/control/src/penalty_shot_direction_estimation.rs
tuxbotix: Maybe-Maybe not - depends on which field, how fast the ball moves, etc; The changes in this PR help to tune better and avoid jumping in the wrong direction. But the fundamental question you asked is multi factored:

### Ball detection at faster velocities

I tried before the 3rd place match without any penalty shootout stuff just to see if the Nao can see fast-moving balls, it struggled in this at faster velocities (judging from eye LEDs).

The fundamental problem (which we had even in 2017, 2018) is that fast (possibly blurred) balls are not detected well enough or they are detected too late (to jump and catch). Sometimes, the lighting caused high camera gain which in turn gave really bad images, etc. Some other times, it was harsh lighting conditions with outdoor fields.

The tuning we always had to do in each of these years was to balance the trigger-happiness on a knife edge to react fast enough but not falsely. The end goal could be to avoid this or minimise the need to tune so much.

### Ball model and filtering

We also need to investigate if ball model switching has a hand in this or not. If I recall correctly, at one point (i think 2018), we had parameters to skip fancier ball filtering for penalty keeper to give it the edge in reacting (remove the lag by filters).

Since we have a dual-ball model system in this season, we also need to verify if the lag is worsened by this (or if the dynamic model's output is always available for cases like this).

I'll check and update the issues we've opened on this topic with these observations.

h3ndrk: Ok, but adding more parameters to aid tuning it on-site is a good thing. Resolving.

crates/control/src/penalty_shot_direction_estimation.rs
h3ndrk: Please do not blindly apply suggestions on GitHub. This variable has now a different name but all remaining code uses still the old name... The CI is red. You also violated our policy that the author of a comment resolves it but you resolved this automatically by applying the suggestion. This comment needs more manual work. :wink:

MaikRe:
Superseded by #1088

PR #495, opened 2023-07-08
Fix path oscillations

### LINKED ISSUE #264: Prevent oscillations in path when walk target behind obstacle.
When there is an obstacle right between the walk target and the robot, the planned path can oscillate between passing the obstacle on either side.
This causes the robot to jockel in place.

julianschuler:
## Introduced Changes

This PR introduces a `rotation_penalty_factor` to the path planner to reduce flipping between two similar paths, which can cause jockeling.
It also adds a scenario file for testing different values of `rotation_penalty_factor`.

Fixes #264.

## ToDo / Known Issues

None.

None.

## How to Test

This is really hard to test outside of the behavior simulator, so I would suggest to watch closely within a game.

### DIFF
--- a/crates/control/src/behavior/node.rs
+++ b/crates/control/src/behavior/node.rs
@@ -39,8 +39,8 @@ pub struct CreationContext {
 #[context]
 pub struct CycleContext {
     pub path_obstacles: AdditionalOutput<Vec<PathObstacle>, "path_obstacles">,
+    pub dribble_path_obstacles: AdditionalOutput<Vec<PathObstacle>, "dribble_path_obstacles">,
     pub active_action: AdditionalOutput<Action, "active_action">,
-    pub time_to_reach_obstacles: AdditionalOutput<Vec<PathObstacle>, "time_to_reach_obstacles">,
 
     pub has_ground_contact: Input<bool, "has_ground_contact">,
     pub world_state: Input<WorldState, "world_state">,
@@ -181,12 +181,23 @@ impl Behavior {
             &walk_and_stand,
             &look_action,
         );
+
+        let mut dribble_path_obstacles = None;
+        let mut dribble_path_obstacles_output = AdditionalOutput::new(
+            context.path_obstacles.is_subscribed()
+                || context.dribble_path_obstacles.is_subscribed(),
+            &mut dribble_path_obstacles,
+        );
+
         let dribble_path = dribble_path_planner::plan(
             &walk_path_planner,
             world_state,
             &context.parameters.dribbling,
-            &mut context.time_to_reach_obstacles,
+            &mut dribble_path_obstacles_output,
         );
+        context
+            .dribble_path_obstacles
+            .fill_if_subscribed(|| dribble_path_obstacles.clone().unwrap_or_default());
 
         let (action, motion_command) = actions
             .iter()
@@ -316,6 +327,12 @@ impl Behavior {
         context.active_action.fill_if_subscribed(|| *action);
 
         self.last_motion_command = motion_command.clone();
+
+        if matches!(action, Action::Dribble) {
+            context
+                .path_obstacles
+                .fill_if_subscribed(|| dribble_path_obstacles.unwrap_or_default())
+        }
 
         Ok(MainOutputs {
             motion_command: motion_command.into(),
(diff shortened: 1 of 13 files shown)


tests/behavior/oscillating_obstacle.lua
```diff
@@ -0,0 +1,36 @@
...
+function on_cycle()
+  if state.cycle_count == 100 then
+    state.ball = {
+      position = { 1.0, 0.0 },
+      velocity = { 0.0, 0.0 },
+    }
+    set_robot_pose(2, { -1.0, 0 }, 0)
+
+    create_obstacle(2, { 0.1, 0.0 }, 0.3)
+  end
```
MaikRe, line 22: I suggest doing this in frame 1 to prevent unnecessary robot teleport after start.

MaikRe (changes requested):

MaikRe:
image
Do you have any idea why this happens? It only appears for one from in the lua file but seems high suspicious to me.

julianschuler:
> image Do you have any idea why this happens? It only appears for 9 frames (913-921) in the lua file but seems highly suspicious to me.

The wierd target position is from the fact that there is no ball filter in the behavior simulator and the robot directly forgets about the ball after it leaves its field of view. Due to its rotation the longer path is preferred. Thus the `rotation_penalty_factor` probably has to be reduced.

tests/behavior/oscillating_obstacle.lua
julianschuler: Thanks, done.

crates/control/src/behavior/node.rs
```diff
@@ -38,11 +40,11 @@ pub struct CreationContext {
 pub struct CycleContext {
     pub path_obstacles: AdditionalOutput<Vec<PathObstacle>, "path_obstacles">,
     pub active_action: AdditionalOutput<Action, "active_action">,
+    pub time_to_reach_obstacles: AdditionalOutput<Vec<PathObstacle>, "time_to_reach_obstacles">,
```
knoellle, line 43: What is the point of this additional output?

knoellle (changes requested):

crates/control/src/behavior/node.rs
julianschuler: This was moved out of `dribble_path_planner` since it is not a node anymore. @MaikRe why did you need this?

crates/control/src/dribble_path_planner.rs
```diff
@@ -1,110 +1,70 @@
...
+pub fn plan(
+    walk_path_planner: &WalkPathPlanner,
+    world_state: &WorldState,
+    dribbling_parameters: &Dribbling,
+    path_obstacles_output: &mut AdditionalOutput<Vec<PathObstacle>>,
+) -> Option<Vec<PathSegment>> {
+    let Some(kick_decisions) = world_state.kick_decisions.as_ref() else { return None };
+    let Some(best_kick_decision) = kick_decisions.first() else { return None };
+    let Some(ball) = world_state.ball else { return None };
+    let Some(robot_to_field) = world_state.robot.robot_to_field else { return None };
```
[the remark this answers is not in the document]
julianschuler: Thanks for the hint, done.

crates/control/src/path_planner.rs
```diff
@@ -34,9 +34,50 @@ pub struct PathPlanner {
...
     pub obstacles: Vec<PathObstacle>,
+    pub last_path_direction: Option<UnitComplex<f32>>,
+    pub rotation_penalty_factor: f32,
 }
 
 impl PathPlanner {
+    pub fn from_last_motion(
+        last_motion_command: &MotionCommand,
+        rotation_penalty_factor: f32,
+    ) -> Self {
```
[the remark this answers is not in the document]
julianschuler: Thanks, done.

crates/control/src/behavior/node.rs
knoellle: More specifically, the `path_obstacles` additional output is currently empty when the robot dribbles.
The dribble action used to fill this output but now that the dribble path is precalculated, it only fills the `time_to_reach_obstacles` (which might not be the best name anyways).
Are there cases where we'd want to have the dribble path obstacles separately even though the robot does something else?
If not, I'd say we should merge the two, otherwise, the dribble obstacles should be copied to the path obstacles when the robot is executing the dribble action.

julianschuler: Good point. I renamed the `time_to_reach_obstacles` to `dribble_path_obstacles` and filled the `path_obstacles` accordingly.

MaikRe (approved):

knoellle (approved):
LGTM

PR #498, opened 2023-07-08
Localization Replay Viewer

knoellle:
## Introduced Changes

Adds a new tool at `tools/localizer` which can load and replay log files generated by #468.
It can be controlled the same way as the behavior simulator but only provides the information stored in the log file augmented with line correspondences.
A map layer to show line correspondences is also added to twix.
Tool code is hacky and optimized to reduce code changes elsewhere.

Fixes #

Proper replay & viewer

## How to Test

```
cd tools/localizer
cargo run <path-to-localization-log.bincode>
```
- Open twix, map panel, behavior simulator panel, connect to localhost, turn on line correspondences.
- Gaze upon the robots path and be delighted.


tools/twix/src/panels/map/layers/line_correspondences.rs
```diff
@@ -0,0 +1,51 @@
...
+    nao::Nao, panels::map::layer::Layer, twix_painter::TwixPainter, value_buffer::ValueBuffer,
+};
+
+pub struct LineCorrespondences {
+    lines_in_robot_bottom: ValueBuffer,
+    lines_in_robot_top: ValueBuffer,
+}
+
+impl Layer for LineCorrespondences {
+    const NAME: &'static str = "Line Associations";
```
h3ndrk, line 19:
```suggestion
    const NAME: &'static str = "Line Correspondences";
```

tools/localizer/src/main.rs
h3ndrk, line 1: Where can I find a localization log file to test this with?

h3ndrk (changes requested):

h3ndrk (dismissed):
Wörks

tools/twix/src/panels/map/layers/line_correspondences.rs
knoellle: Done

h3ndrk (approved):

PR #509, opened 2023-07-09
Penalty shootout jump

Vivituhh:
## Introduced Changes

Reduce stiffness in the jump of the penalty shootout after 0.1s.
Talked to Nao Devils about the timings they have 0.1-0.2s in their motion and it looked good on a robot, but may be tested more. 
Also add stiffnesses in motion file to achieve separat stiffnesses per angle and per motion pose.

Fixes #

## ToDo / Known Issues

## How to Test

### DIFF
--- a/crates/control/src/motion/jump_left.rs
+++ b/crates/control/src/motion/jump_left.rs
@@ -8,7 +8,6 @@ use types::{
     cycle_time::CycleTime,
     joints::JointsCommand,
     motion_selection::{MotionSafeExits, MotionSelection, MotionType},
-    sensor_data::SensorData,
 };
 
 pub struct JumpLeft {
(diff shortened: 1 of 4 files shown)


crates/types/src/joints.rs
```diff
@@ -720,6 +720,60 @@ pub struct JointsCommand<T> {
...
+impl Mul<f32> for JointsCommand<f32> {
+    type Output = JointsCommand<f32>;
+
+    fn mul(self, right: f32) -> Self::Output {
+        Self::Output {
+            positions: Joints::mul(self.positions, right),
+            stiffnesses: self.stiffnesses * right,
+        }
+    }
+}
```
knoellle, line 743: I'm not sure of this implementation.
While it makes sense in the context of interpolating between keyframes, it could be confusing when used in other places.
Without knowing the what this implementation does exactly, I might expect that multiplying a joints command by 0.5 would only change the joint angles.

Vivituhh: I tried a different approach, where I created a new type `JointsWithStiffnesses` to implement the interpolation. To avoid to not interpolate `JointsCommand` at an other position. Maybe you could have a look. 
I didn't looked over the motion again, because I think the motion can be changed in a different PR.

crates/control/src/motion/jump_left.rs
```diff
@@ -6,26 +6,29 @@ use motionfile::{MotionFile, MotionInterpolator};
...
-    motion_safe_exits: PersistentState<MotionSafeExits, "motion_safe_exits">,
+    pub motion_safe_exits: PersistentState<MotionSafeExits, "motion_safe_exits">,
 
-    condition_input: Input<ConditionInput, "condition_input">,
-    cycle_time: Input<CycleTime, "cycle_time">,
-    motion_selection: Input<MotionSelection, "motion_selection">,
+    pub condition_input: Input<ConditionInput, "condition_input">,
+    pub cycle_time: Input<CycleTime, "cycle_time">,
+    pub motion_selection: Input<MotionSelection, "motion_selection">,
+    pub sensor_data: Input<SensorData, "sensor_data">,
```
schmidma, line 31: These `pub`s are legacy. You can remove them

Vivituhh: Done

schmidma (approved):
LGTM

PR #511, opened 2023-07-14
Behavior simulator ball model

knoellle:
## Introduced Changes

On main, the behavior simulator is in a soft-broken state since robots keep losing the ball when they become striker.
This is because they look too far left and right of the ball without having a ball model. That means they immediately lose the ball once it leaves their field of view.

This PR both fixes the simulated glance angle as well as introducing a simplified ball model.
As opposed to the real world, the robot's ball model keeps getting updates with the true position of the ball for some time after the ball leaves the field of view of the robot.
In the real world, noisy percepts are filtered and a physical model is applied to simulate movement of the ball when it is not seen.
In this implementation, a robot would be able to detect a change of the ball's velocity, e.g. due to a kick, without the ball currently being inside the field of view.
This saves us from having to keep separate ball models for each robot that would be updated by simulated percepts and while also simulating the ball physics.
I think this is a reasonable simplification. 

Fixes #

## How to Test

Run any scene in the behavior simulator, e.g. golden goal.
The robots should be able to score goals again without requiring half a dozen striker switches.

### DIFF
--- a/tools/behavior_simulator/src/state.rs
+++ b/tools/behavior_simulator/src/state.rs
@@ -220,44 +220,38 @@ impl State {
 
             robot.database.main_outputs.cycle_time.start_time = now;
 
-            robot.last_seen_ball_in_field = self
-                .ball
-                .clone()
-                .filter(|ball| {
-                    let head_rotation = UnitComplex::from_angle(
-                        robot.database.main_outputs.sensor_data.positions.head.yaw,
-                    );
-                    let ball_in_head = head_rotation.inverse() * ball.position.coords;
-                    let field_of_view = robot.field_of_view();
-                    let angle_to_ball = ball_in_head.angle(&Vector2::x_axis());
+            let ball_visible = self.ball.as_ref().is_some_and(|ball| {
+                let head_rotation = UnitComplex::from_angle(
+                    robot.database.main_outputs.sensor_data.positions.head.yaw,
+                );
+                let ball_in_head = head_rotation.inverse() * ball.position.coords;
+                let field_of_view = robot.field_of_view();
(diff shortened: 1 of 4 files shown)


tools/behavior_simulator/src/state.rs
```diff
@@ -220,25 +214,38 @@ impl State {
...
+                let angle_to_ball = ball_in_head.coords.angle(&Vector2::x_axis());
+
+                angle_to_ball.abs() < field_of_view / 2.0 && ball_in_head.coords.norm() < 3.0
+            });
+            if ball_visible {
+                robot.ball_last_seen = Some(now);
+            }
+            robot.database.main_outputs.ball_position =
+                if robot.ball_last_seen.is_some_and(|last_seen| {
+                    now.duration_since(last_seen).expect("Time ran backwards")
```
h3ndrk, line 238: Non capitalized error messages

h3ndrk (changes requested):

tools/behavior_simulator/src/state.rs
```diff
@@ -183,7 +183,8 @@ impl State {
                 }
                 HeadMotion::LookAt { target, .. } => target.coords.angle(&Vector2::x_axis()),
                 HeadMotion::LookLeftAndRightOf { target } => {
-                    target.coords.angle(&Vector2::x_axis()) + f
+                    target.coords.angle(&Vector2::x_axis())
+                        + f * robot.parameters.look_at.glance_angle
```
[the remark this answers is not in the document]
knoellle: Done

tools/behavior_simulator/src/state.rs
knoellle: Done

h3ndrk (approved):

PR #517, opened 2023-07-16
Recording of robotics nodes

h3ndrk:
## Introduced Changes

- `PersistentState` is renamed to `CyclerState` because it stores state in a cycler
- Node states, setup main outputs, and cross-inputs (from other cyclers) are recorded by serializing to Bincode and writing it to disk
- `hardware.json` is split into `framework.json` and `hardware.json`
- Recording can be enabled by changing the `framework.json`

## ToDo / Known Issues

- Naming of the structs?
- Fix Pepsi parameter setting

- Delete localization recording
- Change pixel storage
- Reuce frequency
- Same seconds in `launchHULK` log files and recording file names
- Reduce recording interleavement
- Add channel directly from robotics domain to "runtime" framework

## How to Test

Deploy to a NAO, check that the cyclers are enabled for recording. When the NAO is in playing primary state, the NAO records to `logs` directory in Bincode format.

### DIFF
--- a/crates/audio/src/whistle_detection.rs
+++ b/crates/audio/src/whistle_detection.rs
@@ -3,12 +3,13 @@ use std::{f32::consts::PI, sync::Arc};
 use color_eyre::Result;
 use context_attribute::context;
 use filtering::statistics::{mean, standard_deviation};
-use framework::{AdditionalOutput, MainOutput};
+use framework::{deserialize_not_implemented, AdditionalOutput, MainOutput};
 use rustfft::{
     num_complex::{Complex32, ComplexFloat},
     num_traits::Zero,
     Fft, FftPlanner,
 };
+use serde::{Deserialize, Serialize};
 use types::{
     parameters::WhistleDetectionParameters,
     samples::Samples,
@@ -20,8 +21,11 @@ pub const NUMBER_OF_AUDIO_CHANNELS: usize = 4;
 pub const NUMBER_OF_AUDIO_SAMPLES: usize = 2048;
 const NUMBER_OF_FREQUENCY_SAMPLES: usize = NUMBER_OF_AUDIO_SAMPLES / 2;
 
+#[derive(Deserialize, Serialize)]
 pub struct WhistleDetection {
+    #[serde(skip, default = "deserialize_not_implemented")]
     fft: Arc<dyn Fft<f32>>,
+    #[serde(skip)]
     scratch: Vec<Complex32>,
 }
 
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -376,9 +376,8 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
         .cycle_nodes
         .iter()
         .map(|node| generate_node_execution(node, cycler, RecordingGeneration::Skip));
-    let required_cross_inputs = get_required_cross_inputs(cycler);
-    let required_cross_input_recordings =
-        generate_required_cross_inputs_recording(cycler, required_cross_inputs);
+    let cross_inputs = get_cross_inputs(cycler);
+    let cross_input_recordings = generate_cross_inputs_recording(cycler, cross_inputs);
 
     let post_setup = match cycler.kind {
         CyclerKind::Perception => quote! {
@@ -460,7 +459,7 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                     let own_subscribed_outputs = self.own_subscribed_outputs_reader.next();
                     let parameters = self.parameters_reader.next();
                     #lock_readers
-                    #required_cross_input_recordings
+                    #cross_input_recordings
                     #(#cycle_node_executions)*
                 }
 
@@ -478,7 +477,7 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
     }
 }
 
-fn get_required_cross_inputs(cycler: &Cycler) -> HashSet<Field> {
+fn get_cross_inputs(cycler: &Cycler) -> HashSet<Field> {
     cycler
         .setup_nodes
         .iter()
@@ -507,11 +506,11 @@ fn get_required_cross_inputs(cycler: &Cycler) -> HashSet<Field> {
         .collect()
 }
 
-fn generate_required_cross_inputs_recording(
+fn generate_cross_inputs_recording(
     cycler: &Cycler,
-    required_inputs: impl IntoIterator<Item = Field>,
+    cross_inputs: impl IntoIterator<Item = Field>,
 ) -> TokenStream {
-    let recordings = required_inputs.into_iter().map(|field| {
+    let recordings = cross_inputs.into_iter().map(|field| {
         let error_message = match &field {
             Field::CyclerState { name, .. } => format!("failed to record cycler state {name}"),
             Field::Input { cycler_instance: Some(_), name, .. } => format!("failed to record input {name}"),
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -371,11 +371,11 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
     let setup_node_executions = cycler
         .setup_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler, true));
+        .map(|node| generate_node_execution(node, cycler, RecordingGeneration::Generate));
     let cycle_node_executions = cycler
         .cycle_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler, false));
+        .map(|node| generate_node_execution(node, cycler, RecordingGeneration::Skip));
     let required_inputs = get_required_inputs(cycler);
     let required_input_recordings = generate_required_inputs_recording(cycler, required_inputs);
 
@@ -392,9 +392,7 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                     #perception_cycler_updates
                 });
                 if enable_recording {
-                    let begin = std::time::Instant::now();
                     bincode::serialize_into(&mut recording_frame, &now).wrap_err("failed to record time")?;
-                    recording_duration += begin.elapsed();
                 }
             }
         }
@@ -447,7 +445,6 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                 };
 
                 let enable_recording = self.enable_recording && <HardwareInterface as hardware::RecordingInterface>::get_recording(&*self.hardware_interface);
-                let mut recording_duration = std::time::Duration::ZERO;
                 let mut recording_frame = Vec::new(); // TODO: possible optimization: cache capacity
 
                 {
@@ -456,7 +453,6 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                     #(#setup_node_executions)*
                 }
 
-                let cycle_begin = std::time::Instant::now();
                 #post_setup
 
                 {
@@ -469,9 +465,7 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
 
                 #after_remaining_nodes
 
-                let cycle_end = cycle_begin.elapsed();
                 if enable_recording {
-                    println!("{instance:?}\t{cycle_end:?}\t{recording_duration:?}");
                     self.recording_sender.try_send(match instance {
                         #(#recording_variants)*
                     }).wrap_err("failed to send recording frame")?;
@@ -611,9 +605,7 @@ fn generate_required_inputs_recording(cycler: &Cycler, required_inputs: Vec<Fiel
     });
     quote! {
         if enable_recording {
-            let begin = std::time::Instant::now();
             #(#recordings)*
-            recording_duration += begin.elapsed();
         }
     }
 }
@@ -631,7 +623,11 @@ fn generate_perception_cycler_updates(cyclers: &Cyclers) -> TokenStream {
         .collect()
 }
 
-fn generate_node_execution(node: &Node, cycler: &Cycler, generate_recording: bool) -> TokenStream {
+fn generate_node_execution(
+    node: &Node,
+    cycler: &Cycler,
+    recording_generation: RecordingGeneration,
+) -> TokenStream {
     let are_required_inputs_some = generate_required_input_condition(node, cycler);
     let node_name = &node.name;
     let node_module = &node.module;
@@ -639,14 +635,12 @@ fn generate_node_execution(node: &Node, cycler: &Cycler, generate_recording: boo
     let context_initializers = generate_context_initializers(node, cycler);
     let recording_error_message = format!("failed to record `{}`", node.name);
     let cycle_error_message = format!("failed to execute cycle of `{}`", node.name);
-    let database_updates = generate_database_updates(node, generate_recording);
+    let database_updates = generate_database_updates(node, recording_generation);
     let database_updates_from_defaults = generate_database_updates_from_defaults(node);
     quote! {
         {
             if enable_recording {
-                let begin = std::time::Instant::now();
                 bincode::serialize_into(&mut recording_frame, &self.#node_member).wrap_err(#recording_error_message)?;
-                recording_duration += begin.elapsed();
             }
             #[allow(clippy::needless_else)]
             if #are_required_inputs_some {
@@ -666,6 +660,11 @@ fn generate_node_execution(node: &Node, cycler: &Cycler, generate_recording: boo
             }
         }
     }
+}
+
+enum RecordingGeneration {
+    Generate,
+    Skip,
 }
 
 fn generate_required_input_condition(node: &Node, cycler: &Cycler) -> TokenStream {
@@ -890,20 +889,24 @@ fn generate_context_initializers(node: &Node, cycler: &Cycler) -> TokenStream {
     }
 }
 
-fn generate_database_updates(node: &Node, generate_recording: bool) -> TokenStream {
+fn generate_database_updates(
+    node: &Node,
+    recording_generation: RecordingGeneration,
+) -> TokenStream {
     node.contexts
         .main_outputs
         .iter()
         .filter_map(|field| match field {
             Field::MainOutput { name, .. } => {
                 let error_message = format!("failed to record {name}");
-                let recording_serialization = generate_recording.then(|| quote! {
-                    if enable_recording {
-                        let begin = std::time::Instant::now();
-                        bincode::serialize_into(&mut recording_frame, &main_outputs.#name.value).wrap_err(#error_message)?;
-                        recording_duration += begin.elapsed();
-                    }
-                });
+                let recording_serialization = match recording_generation {
+                    RecordingGeneration::Generate => quote! {
+                        if enable_recording {
+                            bincode::serialize_into(&mut recording_frame, &main_outputs.#name.value).wrap_err(#error_message)?;
+                        }
+                    },
+                    RecordingGeneration::Skip => Default::default(),
+                };
                 let setter = quote! {
                     #recording_serialization
                     own_database_reference.main_outputs.#name = main_outputs.#name.value;
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -445,7 +445,7 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                     own_database.deref_mut()
                 };
 
-                let enable_recording = self.enable_recording && self.hardware_interface.get_recording();
+                let enable_recording = self.enable_recording && self.hardware_interface.should_record();
                 let mut recording_frame = Vec::new(); // TODO: possible optimization: cache capacity
 
                 {
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -94,6 +94,7 @@ fn generate_struct(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
             #realtime_inputs
             #input_output_fields
             #node_fields
+            recording_file: std::io::BufWriter<std::fs::File>,
         }
     }
 }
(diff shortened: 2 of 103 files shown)


crates/context_attribute/src/lib.rs
```diff
@@ -91,13 +94,13 @@ pub fn context(_attributes: TokenStream, input: TokenStream) -> TokenStream {
...
                     },
-                    "Parameter" | "PersistentState" => match &mut first_segment.arguments {
+                    "CyclerState" | "Parameter" => match &mut first_segment.arguments {
                         PathArguments::AngleBracketed(arguments) if arguments.args.len() == 2 => {
                             pop_string_argument(arguments);
                             let data_type = get_data_type(arguments);
                             into_reference_with_lifetime(
                                 data_type,
-                                (first_segment.ident == "PersistentState").then(Mut::default),
+                                (first_segment.ident != "Parameter").then(Mut::default),
```
knoellle, line 103: Why did you change this comparison?
Previously it said that PersistentState requires a mutable reference. Now it says that parameters don't.
But from just reading the code, no statement is made about whether the Persistent-/CyclerState require it, that information is implicit.

knoellle (changes requested):

knoellle (commented):
Looks good so far. I have left some comments.
From what I can see, the main things left to implement are:
 - serializing node state
 - writing frames to a file

schmidma:
Did I miss something? I can't find the `NodeState` anywhere?

knoellle:
There is no `NodeState` identifier.
All member field of the node structs are part of the node state.
However, fields that are skipped during (de)serialization must be considered ephemeral and should only be used for caching.

schmidma:
As you are implementing `Serialize` and `Deserialize` for nearly all nodes, is it possible to expose these to communication now, such that they can be subscribed via twix?

knoellle:
Not without further changes, no.
I played around with it for a bit and I don't see a way to construct an AdditionalOutput with a type path that is valid both in the module itself as well as in the generated structs module for the database types.
The only way I see for this to work is to move the state to a separate crate (as we do with `types`), but that's no different from the state on the main branch.

schmidma:
Sorry, my previous statement was phrased in a confusing manner. I was expecting further necessary changes. The question was meant in general. We introduce a new (very strong) requirement to nodes: implementing `Serialize` and `Deserialize`. We previously decided against such requirements. But introducing these implementations, a node can directly serialized and thus also exposed via communication.

To realize such a thing, I imagine the generated framework to expose not only databases containing the main and additional outputs, but also the list of nodes to communication. I think they do not necessarily need to exist in a separate crate for that, do they?

Maybe we should discuss further details after the upcoming dev meeting or at another occasion.

knoellle:
Yes, introducing this trait bound certainly opens up the possibility of exposing all nodes state via communication.
Although we also have to consider `SerializeHierarchy`.

crates/code_generation/src/cyclers.rs
```diff
@@ -438,6 +447,139 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
+                        ..
+                    }
+            ) && !required_inputs.contains(field)
+            {
+                required_inputs.push(field.clone());
+            }
+        }
+    }
+    required_inputs
+}
```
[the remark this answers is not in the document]
h3ndrk: It's a naming thing. I named it like that because these are all needed inputs for executing the cycle in replay.

crates/code_generation/src/cyclers.rs
```diff
@@ -415,6 +420,9 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
                     own_database.deref_mut()
                 };
 
+                let enable_recording = <HardwareInterface as hardware::RecordingInterface>::get_recording(&*self.hardware_interface);
+                let mut recording_frame = Vec::new(); // TODO: possible optimization: cache capacity
```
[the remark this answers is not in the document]
h3ndrk: It seems, I forgot to serialize the nodes. The idea was to do this at the beginning. This would be the information needed for replay. Currently, I would try a simple implementation by serializing every frame. Let's see, if we need to optimize.

tools/behavior_simulator/src/interfake.rs
```diff
@@ -23,6 +23,14 @@ impl NetworkInterface for Interfake {
     }
 }
 
+impl RecordingInterface for Interfake {
+    fn get_recording(&self) -> bool {
+        false
+    }
+
+    fn set_recording(&self, _enable: bool) {}
+}
```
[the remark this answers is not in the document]
h3ndrk: We need to discuss these interesting ideas. For the current maturity level, my goal is a feasibility/performance test. Depending on the result we can think about the next steps.

h3ndrk:
> Did I miss something? I can't find the `NodeState` anywhere?

Sorry, please ignore outdated the description... 😅

h3ndrk:
> As you are implementing `Serialize` and `Deserialize` for nearly all nodes, is it possible to expose these to communication now, such that they can be subscribed via twix?

In theory yes but out-of-scope for this PR.

h3ndrk:
> Not without further changes, no.
> 
> I played around with it for a bit and I don't see a way to construct an AdditionalOutput with a type path that is valid both in the module itself as well as in the generated structs module for the database types.
> 
> The only way I see for this to work is to move the state to a separate crate (as we do with `types`), but that's no different from the state on the main branch.

If I would implement it, I would build a separate Communication infrastructure and don't rely on e.g. additional outputs. But this needs more design work.

h3ndrk:
> Yes, introducing this trait bound certainly opens up the possibility of exposing all nodes state via communication.
> 
> Although we also have to consider `SerializeHierarchy`.

Exactly 100% what Schmiddy wrote above. But I think we don't need a trait, Code generation should be enough. But it needs more thinking.

crates/code_generation/src/cyclers.rs
```diff
@@ -438,6 +447,139 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
+        };
+        quote! {
+            if enable_recording {
+                bincode::serialize_into(&mut recording_frame, #value_to_be_recorded).wrap_err(#error_message)?;
+            }
+        }
+    });
+    quote! {
+        #(#recordings)*
+    }
```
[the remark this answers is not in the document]
h3ndrk: Done

h3ndrk:
The current approach has been tested on a real NAO. Here are the findings and conclusions:

# Size of recorded data

1 - 3 GiB data is written per minute. 10 minutes of one half time would be 10 - 30 GiB. The NAO has 32 GiB internal memory. USB sticks can be used for external storage, our usual sticks have 32 GiB storage. That means we might run into storage size problems during games.

One possibility is to optimize the amount of data further but it is unlikely that this will reduce data usage by orders of magnitude. Only some cyclers accumulate a lot of data (vision) and the serialization only has 30 % overhead compared to the raw data. Compression is unlikely to help because this is mostly image data which has high entropy and is hard to compress losslessy. Another possibility is to reduce the frequency of recorded frames, maybe even for some cyclers while recording other cyclers at full frequency.

# Execution time impact

The biggest impact has vision recording with 20 ms Bincode serialization and writing to `BufWriter<File>`. This leaves 13 ms headroom for the actual calculation which is nearly the time our current vision pipeline needs for calculation. Control cycler recording takes 0.3 ms and the whole cycle takes 2 ms (10 ms headroom). The other cyclers are even faster. Again, vision is the critical path and needs special care. If we would reduce the frequency, only every other frame would take longer. Maybe we would skip images but we control that via the frequency. One idea is to change the internal buffer format of `YCbCr422Image` to `Vec<u8>` to decrease one layer of serialization hopefully resulting in simple generated code. All accesses must be changed to use the different buffer format. Maybe we need to introduce some `unsafe` stuff to efficiently cast pointers around.

h3ndrk:
Some ideas to support different frequencies and dynamic recording.

Dynamic recording may be triggered by different parties. For example a trigger might be the current game state (robotics domain) or someone requesting a recording from Twix (debug tooling, framework domain via communication) or it has been requested from the deployed configuration (framework domain via communication).

Communication and the hardware interface will both provide a quadruple state: Force to not record, force to record, intend to not record, intend to record. This is also the order of descending precendence. Communication will provide a state for each cycler enabling fine-grained control of the recording. The hardware interface may just provide a single state for the moment, but this can be extended later to one for each cycler.

If the cycler derives from all states that a recording of the current frame should happen, it records everything and sends the data to its instance of a channel to communication. Communication will receive the data and store it in the file system. We might think about, reducing computation overhead in the cyclers even more by moving the serialization into communication.

Communication will also provide recording frequency information, for example whether every frame should be recorded, none at all or at which frequency. One approach is to have this fixed over the whole process runtime by storing this information in e.g. the `hardware.json` configuration and loading it once at startup. Another approach is to also provide this information via communication and be able to control it from a connected debug client at runtime. Dynamic frequencies have the benefit of more fine-grained analysis during testing without restarting the HULK process.

Communication will serve an additional API endpoint for recording s.t. clients are able to control the recording state and maybe also the frequency.

crates/code_generation/src/cyclers.rs
```diff
@@ -353,11 +371,13 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
         .setup_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler));
+        .map(|node| generate_node_execution(node, cycler, true));
     let cycle_node_executions = cycler
         .cycle_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler));
+        .map(|node| generate_node_execution(node, cycler, false));
+    let required_inputs = get_required_inputs(cycler);
```
h3ndrk, line 379: Unused?

crates/code_generation/src/cyclers.rs
```diff
@@ -415,29 +446,178 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
                     #lock_readers
+                    #required_input_recordings
                     #(#cycle_node_executions)*
                 }
 
                 #after_remaining_nodes
+
+                let cycle_end = cycle_begin.elapsed();
+                if enable_recording {
+                    println!("{instance:?}\t{cycle_end:?}\t{recording_duration:?}");
```
h3ndrk, line 474: Remove debug prints

crates/code_generation/src/cyclers.rs
```diff
@@ -451,17 +631,23 @@ fn generate_perception_cycler_updates(cyclers: &Cyclers) -> TokenStream {
...
+    let recording_error_message = format!("failed to record `{}`", node.name);
+    let cycle_error_message = format!("failed to execute cycle of `{}`", node.name);
+    let database_updates = generate_database_updates(node, generate_recording);
     let database_updates_from_defaults = generate_database_updates_from_defaults(node);
     quote! {
         {
+            if enable_recording {
+                let begin = std::time::Instant::now();
+                bincode::serialize_into(&mut recording_frame, &self.#node_member).wrap_err(#recording_error_message)?;
+                recording_duration += begin.elapsed();
```
h3ndrk, line 649: Remove execution time measurements

crates/code_generation/src/cyclers.rs
```diff
@@ -698,20 +884,28 @@ fn generate_context_initializers(node: &Node, cycler: &Cycler) -> TokenStream {
...
-            })
-    ;
+            });
     quote! {
         #(#initializers,)*
     }
 }
 
-fn generate_database_updates(node: &Node) -> TokenStream {
+fn generate_database_updates(node: &Node, generate_recording: bool) -> TokenStream {
```
h3ndrk, line 893: Enum

crates/control/src/localization_recorder.rs
```diff
@@ -15,7 +15,9 @@ use types::{
     game_controller_state::GameControllerState, line_data::LineData, primary_state::PrimaryState,
 };
 
+#[derive(Deserialize, Serialize)]
 pub struct LocalizationRecorder {
```
h3ndrk, line 19: Remove or track in ideas for next iteration

crates/code_generation/src/cyclers.rs
h3ndrk: No

crates/audio/src/whistle_detection.rs
```diff
@@ -20,8 +21,11 @@ pub const NUMBER_OF_AUDIO_CHANNELS: usize = 4;
 pub const NUMBER_OF_AUDIO_SAMPLES: usize = 2048;
 const NUMBER_OF_FREQUENCY_SAMPLES: usize = NUMBER_OF_AUDIO_SAMPLES / 2;
 
+#[derive(Deserialize, Serialize)]
 pub struct WhistleDetection {
+    #[serde(skip, default = "deserialize_not_implemented")]
```
schmidma, line 26: what do you think of https://doc.rust-lang.org/std/macro.unimplemented.html instead of `deserialize_not_implemented`?

crates/control/src/localization_recorder.rs
schmidma, line 19: Can we remove this node now?

crates/hardware/src/lib.rs
```diff
@@ -43,6 +44,11 @@ pub trait PathsInterface {
     fn get_paths(&self) -> Paths;
 }
 
+pub trait RecordingInterface {
+    fn get_recording(&self) -> bool;
+    fn set_recording(&self, enable: bool);
```
schmidma, line 49: As discussed. Please reiterate on the naming here ;)

crates/code_generation/src/run.rs
```diff
@@ -147,6 +199,8 @@ fn generate_cycler_constructors(cyclers: &Cyclers) -> TokenStream {
                 communication_server.get_parameters_reader(),
                 #own_producer_identifier
                 #(#other_cycler_inputs,)*
+                recording_sender.clone(),
+                cycler_instances_to_be_recorded.contains(#cycler_instance_name),
```
schmidma, line 203: Please extract that variable to an earlier position like e.g. `enable_recording`

schmidma (changes requested):
only minor remarks. This looks good to me in general. Let's see a test game before merging this?

crates/code_generation/src/cyclers.rs
```diff
@@ -698,20 +888,32 @@ fn generate_context_initializers(node: &Node, cycler: &Cycler) -> TokenStream {
...
     node.contexts
         .main_outputs
         .iter()
         .filter_map(|field| match field {
             Field::MainOutput { name, .. } => {
+                let error_message = format!("failed to record {name}");
+                let recording_serialization = match recording_generation {
+                    RecordingGeneration::Generate => quote! {
+                        if enable_recording {
+                            bincode::serialize_into(&mut recording_frame, &main_outputs.#name.value).wrap_err(#error_message)?;
```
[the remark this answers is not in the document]
h3ndrk: I generally agree with optimizing. Do we want to do that in a later iteration? This latency is currently in the range 0.1 milliseconds in the control cycler (the only recorded cycler at the moment). In the future we can further improve the point of recording, e.g. recording only cycler states at the beginning and the rest in the end. But we also need to be careful to finish cycles early but also keep holding locks for the values to record s.t. we record the values of the last cycle...

crates/code_generation/src/run.rs
```diff
@@ -97,6 +113,42 @@ fn generate_future_queues(cyclers: &Cyclers) -> TokenStream {
...
+        let error_message = format!("failed to write into recording file for {instance}");
+        quote! {
+            crate::cyclers::RecordingFrame::#instance_name { data } => #instance_name_snake_case.write_all(data.as_slice()).wrap_err(#error_message)?,
+        }
+    });
+
+    quote! {
+        std::thread::Builder::new()
+            .name("Recording".to_string())
+            .spawn(move || -> color_eyre::Result<()> {
```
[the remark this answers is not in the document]
h3ndrk: Yes

```
error[E0282]: type annotations needed
    --> target/incremental/build/behavior_simulator-b25b3ec13813f75b/out/generated_code.rs:1178:17
     |
1178 |                 Ok(())
     |                 ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
     |
help: consider specifying the generic arguments
     |
1178 |                 Ok::<(), E>(())
     |                   +++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `behavior_simulator` (lib test) due to previous error
warning: build failed, waiting for other jobs to finish...
```

crates/code_generation/src/run.rs
```diff
@@ -97,6 +113,42 @@ fn generate_future_queues(cyclers: &Cyclers) -> TokenStream {
...
+                use std::io::Write;
+                let seconds = std::time::SystemTime::now().duration_since(std::time::SystemTime::UNIX_EPOCH).unwrap().as_secs();
```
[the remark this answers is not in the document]
h3ndrk: Added to Ideas for Next Iteration

crates/audio/src/whistle_detection.rs
h3ndrk:
```
error[E0423]: expected function, found macro `unimplemented`
  --> crates/audio/src/whistle_detection.rs:26:29
   |
26 |     #[serde(skip, default = "unimplemented")]
   |                             ^^^^^^^^^^^^^^^ not a function
   |
help: use `!` to invoke the macro
   |
26 |     #[serde(skip, default = "unimplemented"!)]
   |                                            +
```

crates/control/src/localization_recorder.rs
h3ndrk: Yesterday I added this to Ideas for Next Iteration because the localization recording is a fully working feature (with replay in Twix) but recording alone is not feature-complete since replay is still missing. Once a basic version of replay is working, we can delete it.

crates/code_generation/src/cyclers.rs
```diff
@@ -353,11 +371,13 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
         .setup_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler));
+        .map(|node| generate_node_execution(node, cycler, RecordingGeneration::Generate));
     let cycle_node_executions = cycler
         .cycle_nodes
         .iter()
-        .map(|node| generate_node_execution(node, cycler));
+        .map(|node| generate_node_execution(node, cycler, RecordingGeneration::Skip));
+    let required_inputs = get_required_inputs(cycler);
```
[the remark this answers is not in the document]
h3ndrk: "required inputs" is now "required cross inputs". I hope this is sufficient now, otherwise please make a proposal :wink:

crates/code_generation/src/cyclers.rs
```diff
@@ -427,17 +459,162 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers) -> TokenStream {
...
+                }
             }
             self.own_changed.notify_one();
             Ok(())
         }
     }
 }
 
+fn get_required_inputs(cycler: &Cycler) -> Vec<Field> {
+    let mut required_inputs = Vec::new();
```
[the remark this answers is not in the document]
h3ndrk: HashSettified now

crates/code_generation/src/run.rs
```diff
@@ -97,6 +113,42 @@ fn generate_future_queues(cyclers: &Cyclers) -> TokenStream {
         .collect()
 }
 
+fn generate_recording_thread(cyclers: &Cyclers) -> TokenStream {
+    let file_creations = cyclers.instances().map(|(_cycler, instance)| {
+        let instance_name_snake_case = format_ident!("{}", instance.to_case(Case::Snake));
+        let instance_name = format!("logs/{instance}.{{seconds}}.bincode");
```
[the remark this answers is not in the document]
h3ndrk: It is now "recording file name"

crates/code_generation/src/run.rs
```diff
@@ -97,6 +113,42 @@ fn generate_future_queues(cyclers: &Cyclers) -> TokenStream {
...
+        let error_message = format!("failed to create recording file for {instance}");
+        quote! {
+            let mut #instance_name_snake_case = std::io::BufWriter::new(std::fs::File::create(format!(#instance_name)).wrap_err(#error_message)?);
```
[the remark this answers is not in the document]
h3ndrk: Added a TODO comment

crates/hardware/src/lib.rs
h3ndrk: Renamed to

- "should record"
- "set whether to record"

Feel free to propose something :wink:

crates/code_generation/src/run.rs
h3ndrk: Extracted to variable

crates/audio/src/whistle_detection.rs
h3ndrk: Same with `panic!()`. I thought of something like [`panic_any()`](https://doc.rust-lang.org/std/panic/fn.panic_any.html) function but I need a function `fn() -> T`.

crates/code_generation/src/cyclers.rs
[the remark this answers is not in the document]
schmidma: I think, the "required" needs context. Required for what?

crates/code_generation/src/cyclers.rs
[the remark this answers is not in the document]
schmidma: Then at least, we should not forget about it? I'm generally concerned on doing that interleaved, but 0.1ms might actually be negligible.

crates/audio/src/whistle_detection.rs
schmidma: Then at least use it in your custom function?

crates/code_generation/src/cyclers.rs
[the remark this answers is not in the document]
h3ndrk: Let's make it longer :joy: How about "cross input required for recording"?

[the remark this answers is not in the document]
h3ndrk: Kloppt euch gern :joy:

crates/audio/src/whistle_detection.rs
h3ndrk: Using it now. I thought you wanted to use `unreachable!()` which I would argue against. But `unimplemented!()` is fine for me.

crates/code_generation/src/cyclers.rs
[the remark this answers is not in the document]
h3ndrk: It is now `cross_inputs`, I hope this is fine now.

crates/hardware/src/lib.rs
h3ndrk: To explain my motivation in keeping it for now and trying to shortcut the discussion: I think we should keep in mind the intended maturity level. The long-term plan is (as written in a comment earlier) that the `framework.json`, Communication, and `HardwareInterface` will provide whether they want to record, in which frequency, and/or whether they want to record a burst for a given time duration. This will all be some state that is written once in a cycle or in a lower frequency. The simple boolean that currently exists will be replaced by that. Therefore, I think we should not overfit to it now. In addition, I would vote against making the robotics domain longer due to calling two functions instead of one. The idea of "record this cycle" changes the API and requires additional state and code to reset the state, while we are not yet sure whether this additional code can survive under the planned ideas.

schmidma (approved):

PR #528, opened 2023-07-29
Hardcode cache in workflow

### LINKED ISSUE #527: The new CI is slower than the previous iteration
With #512 we introduced a next iteration of CI. The build time changed from 2-3 minutes to 7-8 minutes. The following things have been changed:

- each job spins up a Docker container instead of one long running container
- we have more runners (from 5 to 10 now), reducing cache benefits because caches are runner-local
- we have deployed runners to the DELL servers, which seem to be slower than Rechenknecht

My guess is that the first point is not that relevant, but the other two points might cause this.

h3ndrk:
## Introduced Changes

It seems that environment variables configured in the runners `.env` are not forwarded to the Docker container, this PR adds the variables to our workflow. This is a little smelly, because the execution environment leaks into our workflow, but since we also configure the user to run the job steps with, it should be fine.

Fixes #527

## ToDo / Known Issues

None

None

## How to Test

Check that everything compiles and the cache is picked up on later jobs.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - gh-readonly-queue/main/**
   pull_request:
+env:
+  CARGO_HOME: /__w/hulk/cargo
+  CARGO_TARGET_DIR: /__w/hulk/target
 jobs:
   required_checks:
     name: Require all CI jobs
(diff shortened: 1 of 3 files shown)


h3ndrk:
Now we are down to around 2 minutes. The caches aren't even hot yet. Looks way better. Ready for review.

knoellle (approved):
Significant improvement once the cache is warmed up, thanks!

PR #530, opened 2023-08-08
Remove unused dependencies

knoellle:
## Introduced Changes

Used [`cargo-machete`](https://github.com/bnjbvr/cargo-machete) to find and remove unused dependencies.
There were some false positives due to the generated code not being considered when searching for uses. These have been removed.

While looking over the changes, I also found a false negative: `hulks_webots` depended on the `audio` crate even though it is never used. The false positive probably stems from [this line](https://github.com/HULKs/hulk/blob/main/crates/hulk_webots/src/hardware_interface.rs#L26) which, taken in isolation looks like it uses the audio crate.
There may be more false negatives.

From my limited testing (mostly building `hulk_nao` and `hulks_webots`) there seems to be no significant gain in the time it takes to compile.

Fixes #

Would be nice to have something like this in our CI but I'm not sure that's a good idea due to the large number of false positives.
`cargo-machete` provides a way to ignore certain dependencies, but that's one more list of dependencies we have to maintain.
We may want to discuss this in a future dev-meeting.

## How to Test

Check if everything still builds/checks/runs correctly.
Due to the nature of these changes, I don't see how they might break anything at runtime while compiling successfully. But I tested this on a NAO nonetheless.


okiwi6 (approved):
Nice work, since the integration into CI does not seem straight forward, I opened an issue (#533) for this to keep track of that so we can merge this PR.

PR #536, opened 2023-08-24
1.72 delinting

knoellle:
## Introduced Changes

Fixes lints newly added to clippy in [Rust 1.72.0](https://blog.rust-lang.org/2023/08/24/Rust-1.72.0.html).

Fixes #

## ToDo / Known Issues

Not sure about the change to the code generation, hence the extra commit. I'm in favor of ignoring the lint instead of adding more logic to the code generation but I'm unsure where to place the `#[allow(...)]` exactly.
For now I have placed it directly on the if statement in question, which is the one wrapping the cycle call for each node.
In the true case, the node is executed and in the else case the respective main outputs in the database are set to default.
However, there are some nodes that don't have main outputs, leaving the else branch empty.
@h3ndrk Opinions?

## How to Test

Try to build/clippy with latest Rust release. There should be no more warnings.

### DIFF
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -462,6 +462,7 @@ fn generate_node_execution(node: &Node, cycler: &Cycler) -> TokenStream {
     let database_updates_from_defaults = generate_database_updates_from_defaults(node);
     quote! {
         {
+            #[allow(clippy::needless_else)]
             if #are_required_inputs_some {
                 let main_outputs = {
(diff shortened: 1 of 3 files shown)


h3ndrk (approved):
Fine for me. I assume the compiler will optimize the predicate for empty else cases away (maybe check the assembly output). Therefore this warning silencing is only a syntactical thing and that is IMHO fine for generated code. And the locality of this allow-statement to the statement that generates the warning is also a good thing, documentation-wise.

PR #555, opened 2023-09-06
Players iterator traits

### LINKED ISSUE #415: Players Container Iterators
Our [`Players`](https://github.com/HULKs/hulk/blob/main/crates/types/src/players.rs) container struct currently does not implement more fancy iterator traits such as `DoubleEndedIterator` (for `.rev()`) or `ExactSizeIterator`, see [here](https://doc.rust-lang.org/std/iter/index.html#traits).

knoellle:
## Introduced Changes

Implements [`DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) and [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) for the `Players` container type.
Note that neither trait is currently used outside of tests.

Plus some cleanup by indexing into the Player struct more.

Fixes #415

## How to Test

There should be no functional change.

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -117,15 +117,7 @@ impl RoleAssignment {
                     player_roles[striker] = Role::Striker;
                 }
             }
-            role = match context.player_number {
-                PlayerNumber::One => player_roles.one,
-                PlayerNumber::Two => player_roles.two,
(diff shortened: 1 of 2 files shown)


okiwi6 (approved):
LGTM

h3ndrk:
What is the benefit of this change?

PR #559, opened 2023-09-25
Cihaviorsimulator

### LINKED ISSUE #551: Integrate Behavior Simulator into CI
We want the behavior simulator in our CI pipeline so we can check for golden goals, defending behavior, ...
This includes adding return values to the behavior simulator and adding it to the CI.

Estimated time: 1 week

knoellle:
## Introduced Changes

This PR adds tools to the lua API with which behavior scenarios can implement pass/fail conditions.
These conditions can then be checked in a CI job to prevent behavior regressions.

The behavior simulator's CLI interface was changed, it now takes a command: `run` or `serve`, both of which still take the path to a scenario file.
`run` exits after the simulation finishes, `serve` replicates the old behavior.

Fixes #551

## ToDo / Known Issues

- [x] Add way to run simulator without starting a blocking communication server
- [x] Allow generating errors from lua
- [x] Run scenarios in CI
  - [x] Get inspect.lua into CI environment
  - [x] Figure out how to organize running multiple scenarios
- [x] Don't terminate program on error in `serve` mode

## How to Test

Run scenario like so:
```
cargo run --manifest-path tools/behavior_simulator/Cargo.toml --release run tests/behavior/golden_goal.lua
```
This should run without errors and exit with code 0.

Now break it in some way, for example by either never setting the goal scored flag in the scenario file or never assigning a striker.

The above command should then show an error message and exit with code 1.


.github/workflows/pull-request.yml
```diff
@@ -117,3 +117,25 @@ jobs:
...
+      image: 134.28.57.223:5000/hulk
+      options: --user=1000:1000
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          lfs: true
+      - name: Build simulator
+        run: cargo build --manifest-path tools/behavior_simulator/Cargo.toml --release
+      - name: Run scenario
+        run: cargo run --manifest-path tools/behavior_simulator/Cargo.toml --release run ${{ matrix.scenario }}
```
knoellle, line 141: # Discussion question: How do we want to organize these test?

I mainly see three options here

### Using the matrix feature

This would result in `n` *jobs*, each of which is shown separately in the GitHub UI.
However, this also incurs a job dispatch overhead for each scenario file and up to `n` runner have to build the simulator.

### Using a single job/step that runs all scenarios

This means no per-scenario granularity in the UI but is likely much faster. Only one runner has to build the simulator.
Sadly there is no matrix for *steps*, so we can't dynamically generate steps for each scenario.

### Using a single job with one step per scenario

A bit more granularity as each scenario run would be shown in a collapsible section in the action log, even though there's only a single pass/fail state shown in the UI and only the first failing scenario is detected.
However, this would mean duplicating the shell command for every scenario and hardcoding the path.
I don't think this is a particularly elegant solution even though it probably has the best balance between performance and convenience.

## Build / run times

On my PC, rebuilding the behavior simulator in release mode after touching a file in the `control` crate takes about 9.3 seconds.
Running the golden goal scenario takes about 0.2 seconds.

From looking at past actions job runs, the overhead for container setup/teardown, repo checkout steps seems to be somewhere between [8](https://github.com/HULKs/hulk/actions/runs/6305634866/job/17119390831) and [20](https://github.com/HULKs/hulk/actions/runs/6305634866/job/17119391274) seconds.
This does not include time spent "waiting for a runner to pick up the job".

knoellle: Decided to use rust integration tests now.
One test per scenario but all in the same file to reduce building/linking overhead.

schluis (approved):
LGTM! (After some disucssion)
Very nice! :)

PR #562, opened 2023-09-27
Only repaint Twix when new data is received

knoellle:
## Introduced Changes

Instead of repainting twix with the monitor's refresh rate, only request a repaint when new data is received by the communication client.

This is implemented by introducing a general notification feature to the communication client which sends an empty mpsc message when new data arrives (from outputs or parameter changes).
The `Nao` type in twix then provides a way to register a callback executed on these channel messages.
The twix app then registers such a callback which requests a repaint with the egui context.

Fixes #

## How to Test

Compared to main, twix should use much less cpu/gpu when not receiving updates.
The greatest effect should be seen on devices like @h3ndrk's Gurke.

### DIFF
--- a/crates/communication/src/client/output_subscription_manager.rs
+++ b/crates/communication/src/client/output_subscription_manager.rs
@@ -11,6 +11,7 @@ use uuid::Uuid;
 use crate::{
     client::{
         id_tracker::{self, get_message_id},
+        notify::notify_all,
         responder, Output, SubscriberMessage,
     },
     messages::{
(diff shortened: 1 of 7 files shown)


tools/twix/src/nao.rs
```diff
@@ -88,6 +91,21 @@ impl Nao {
         self.runtime
             .block_on(self.communication.update_parameter_value(path, value));
     }
+
+    pub fn on_update<F>(&self, f: F)
```
okiwi6, line 95: more descriptive name, e.g. `callback`?

crates/communication/src/client/output_subscription_manager.rs
```diff
@@ -201,6 +205,11 @@ pub async fn output_subscription_manager(
                         }
                     }
                 }
+                for sender in &notification_senders {
```
okiwi6, line 208: Also add to other `Update...` message types

okiwi6 (changes requested):
Only minor remarks

crates/communication/src/client/output_subscription_manager.rs
knoellle: Done

tools/twix/src/nao.rs
knoellle: Renamed

crates/communication/src/client/parameter_subscription_manager.rs
```diff
@@ -159,6 +164,11 @@ pub async fn parameter_subscription_manager(
                         error!("{error}");
                     }
                 }
+                for sender in &notification_senders {
```
okiwi6, line 167: Maybe also extract this code to a function?

okiwi6 (changes requested):

crates/communication/src/client/parameter_subscription_manager.rs
knoellle: Döne

okiwi6 (approved):
Nice!

PR #564, opened 2023-09-28
Allow for parallel camera access

okiwi6:
## Introduced Changes

Makes the camera access non-mutable so that it can be used from multiple cyclers.
This will be required for other cyclers that want to operate on camera images.

Fixes #

## ToDo / Known Issues
* [x] Test on NAO

## How to Test
* upload check if everything still works
* check that images in twix are still displayed correctly


crates/hulk_nao/src/camera.rs
```diff
@@ -82,9 +101,8 @@ impl Camera {
         bail!("too many unsuccessful waiting retries");
     }
 
-    fn reset(&mut self) -> Result<()> {
+    fn reset(&self) -> Result<()> {
         let _lock = self.i2c_head_mutex.lock();
-        self.camera.take();
```
schmidma, line 87: For the duration of the camera reset, the camera is not locked. What happens to other receivers requesting access to the camera? Should this `reset()` lock the camera for the duration of the reset?

crates/vision/src/image_receiver.rs
```diff
@@ -1,20 +1,27 @@
...
-pub struct CreationContext {}
+pub struct CreationContext {
+    hardware_interface: HardwareInterface,
+}
 
 #[context]
 pub struct CycleContext {
     hardware_interface: HardwareInterface,
     camera_position: Parameter<CameraPosition, "image_receiver.$cycler_instance.camera_position">,
+    cycle_time: AdditionalOutput<Duration, "cycle_time">,
```
schmidma, line 24: naming... what do you think of `last_cycle_duration`

schmidma (changes requested):
only minor things, looks good in general

crates/vision/src/image_receiver.rs
okiwi6: Done

crates/hulk_nao/src/camera.rs
okiwi6: `reset()` is private

schmidma (approved):

PR #567, opened 2023-09-29
Check if location is valid

### LINKED ISSUE #561: Check if location exists for pregame command
When executing the pregame command, it is not checked if the location is valid

CallMeElle:
## Introduced Changes

Pepsi didn't check if location exists. Now there is a check in set_location if parameter location is an existing location.

Fixes #561

## ToDo / Known Issues

implement something similar as for networks: suggestion of location which matches closely with input if the input location does not exist

## How to Test
./pepsi (should run normally)
./pepsi pregame uzt SPL_A 34:1  (wrong location name -> should fail and print all locations that exist)
./pepsi pregame smd SPL_A 34:1 (should run normally)

### DIFF
--- a/crates/repository/src/lib.rs
+++ b/crates/repository/src/lib.rs
@@ -2,7 +2,7 @@ use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
     env::current_dir,
     ffi::OsStr,
-    fmt::Display,
+    fmt::{Display, Write},
     fs::Permissions,
     io::{self, ErrorKind},
     os::unix::prelude::PermissionsExt,
@@ -364,6 +364,15 @@ impl Repository {
     pub async fn set_location(&self, target: &str, location: &str) -> Result<()> {
         let target_location = self.parameters_root().join(format!("{target}_location"));
         let new_location = Path::new(location);
+        let new_location_path = self.parameters_root().join(location);
+        if !new_location_path.exists(){
+            let mut available_locations = String::new();
+            let location_set = self.list_available_locations().await?;
+            for location in location_set{
+                writeln!(&mut available_locations, "  -{location}")?;
+            }
+            bail!("available locations are:\n{available_locations}but you wrote: {location}");
+        }
         let _ = remove_file(&target_location).await;
         symlink(&new_location, &target_location)
             .await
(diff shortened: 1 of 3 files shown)


crates/repository/src/lib.rs
```diff
@@ -364,6 +364,15 @@ impl Repository {
     pub async fn set_location(&self, target: &str, location: &str) -> Result<()> {
         let target_location = self.parameters_root().join(format!("{target}_location"));
         let new_location = Path::new(location);
+        let new_location_path = self.parameters_root().join(location);
+        if !new_location_path.exists() {
```
schmidma, line 368: Use a https://docs.rs/tokio/latest/tokio/fs/fn.try_exists.html instead. Your code is executed in an async context, we can thus use async functions to perform io.

crates/repository/src/lib.rs
```diff
@@ -364,6 +364,15 @@ impl Repository {
...
+            let mut available_locations = String::new();
+            let location_set = self.list_available_locations().await?;
+            for location in location_set {
+                writeln!(&mut available_locations, "  -{location}")?;
+            }
+            bail!("available locations are:\n{available_locations}but you wrote: {location}");
```
schmidma, line 374: IMHO, the error should be first, and then the hints and suggestions to the user. Thus, swap `available_locations` with the missing `location`

crates/repository/src/lib.rs
```diff
@@ -364,6 +364,15 @@ impl Repository {
     pub async fn set_location(&self, target: &str, location: &str) -> Result<()> {
         let target_location = self.parameters_root().join(format!("{target}_location"));
         let new_location = Path::new(location);
+        let new_location_path = self.parameters_root().join(location);
+        if !new_location_path.exists() {
+            let mut available_locations = String::new();
+            let location_set = self.list_available_locations().await?;
+            for location in location_set {
+                writeln!(&mut available_locations, "  -{location}")?;
+            }
```
schmidma, line 373: You can use `.iter()` and `.collect()` instead. If you want to use your " - ", you can try https://doc.rust-lang.org/std/iter/struct.Intersperse.html

schmidma (changes requested):
looks good in general, didn't test yet, only some minor remarks

crates/repository/src/lib.rs
CallMeElle: Done

crates/repository/src/lib.rs
CallMeElle: Done

crates/repository/src/lib.rs
CallMeElle: Done

schmidma (approved):

PR #569, opened 2023-09-30
Twix slider UX

### LINKED ISSUE #534: Slow slider in Twix
The slider as used in the camera calibration panel or field color calibration panel is massive pain to use.
They are lagging behind the mouse position and need a lot of time to reach their final position.
Probably this is because for every slider change, a parameter change request is send to the robot.
Maybe apply rate limiting or send it only when the sliderbutton is released.

knoellle:
## Introduced Changes

Previously, a changed parameter slider value would have to be sent to the server and the updated slider value sent back before the new value was reflected in the slider position.
This is especially an issue if there is more than one update "in flight" as it would cause the slider to "replay" the movement over some period of time.

This PR changes the behavior by moving the parameter setting feature to the value buffers, where the new value is applied immediately. For each change request sent to the nao, one update is ignored to prevent overwriting newly requested values with old responses.

Fixes #534 

## How to Test

- Open twix
- Connect to nao
- For added effect, open image panel
- Open a panel with sliders (manual calibration, vision tuner)
- Drag a slider around

Compared to current master, the slider should not jump around anymore.

### DIFF
--- a/tools/twix/src/panels/manual_camera_calibration.rs
+++ b/tools/twix/src/panels/manual_camera_calibration.rs
@@ -17,7 +17,7 @@ type SubscribedType = Vector3<f64>;
 struct CameraParameterSubscriptions<DeserializedValueType> {
     human_friendly_label: String,
     path: String,
-    value_buffer: Option<ValueBuffer>,
+    value_buffer: ValueBuffer,
     value: DeserializedValueType,
     update_notify_receiver: mpsc::Receiver<()>,
 }
@@ -39,7 +39,7 @@ impl Panel for ManualCalibrationPanel {
             let path = CAMERA_KEY_BASE.to_owned() + name.to_lowercase().as_str() + ROTATIONS;
 
             let (update_notify_sender, update_notify_receiver) = mpsc::channel(1);
-            let value_buffer = subscribe(nao.clone(), &path, update_notify_sender);
+            let value_buffer = subscribe(nao.clone(), &path, update_notify_sender).unwrap();
 
             info!("Subscribing to path {}", path);
 
@@ -66,7 +66,7 @@ fn add_extrinsic_calibration_ui_components(
     repository_parameters: &RepositoryParameters,
     extrinsic_rotations_subscription: &mut CameraParameterSubscriptions<Option<SubscribedType>>,
 ) {
-    let extrinsic_rotations_buffer_option = &extrinsic_rotations_subscription.value_buffer;
+    let extrinsic_rotations_buffer = &extrinsic_rotations_subscription.value_buffer;
     let mut extrinsic_rotations_option = &mut extrinsic_rotations_subscription.value;
     let label = &extrinsic_rotations_subscription.human_friendly_label;
     let extrinsic_rotations_subscription_path = &extrinsic_rotations_subscription.path;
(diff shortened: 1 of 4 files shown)


tools/twix/src/panels/manual_camera_calibration.rs
```diff
@@ -39,7 +39,7 @@ impl Panel for ManualCalibrationPanel {
             let path = CAMERA_KEY_BASE.to_owned() + name.to_lowercase().as_str() + ROTATIONS;
 
             let (update_notify_sender, update_notify_receiver) = mpsc::channel(1);
-            let value_buffer = subscribe(nao.clone(), &path, update_notify_sender);
+            let value_buffer = subscribe(nao.clone(), &path, update_notify_sender).unwrap();
```
okiwi6, line 42: I was wondering if this could fail if there is no connection to the nao yet. While testing I moved some sliders and reconnected some times using the connected check. This crashed consistently not at this line but here 
`thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: SendError { .. }', hulk/crates/communication/src/client/communication.rs:146:14`

okiwi6: But just checked, this problem also exists on `main` :sweat_smile:

tools/twix/src/value_buffer.rs
```diff
@@ -141,28 +152,30 @@ impl ValueBuffer {
...
-                                    },
-                                    _ => {
-                                        let mut new_buffer = VecDeque::with_capacity(buffer_capacity);
-                                        new_buffer.push_back(new_value);
-                                        values = Some(Ok(new_buffer));
-                                    },
+                            SubscriberMessage::Update{value} => {
+                                if skip_updates > 0 {
+                                    skip_updates -= 1;
+                                    println!("Skipping update");
```
okiwi6, line 172: Forgot to remove the print?

tools/twix/src/value_buffer.rs
```diff
@@ -141,28 +152,30 @@ impl ValueBuffer {
...
+                                    continue;
+                                }
+                                if parameter_path.is_some() {
+                                    println!("Updating");
```
okiwi6, line 176: Forgot to remove the print?

tools/twix/src/value_buffer.rs
```diff
@@ -217,10 +230,39 @@ async fn value_buffer(
                         Message::ListenToUpdates{response_sender} => {
                             update_listeners.push(response_sender)
                         },
+                        Message::UpdateParameterValue{value} => {
+                            skip_updates += 1;
+                            println!("Sending update");
```
okiwi6, line 235: Forgot to remove the print?

tools/twix/src/value_buffer.rs
```diff
@@ -217,10 +230,39 @@ async fn value_buffer(
...
+                            add_element(&mut values, buffer_capacity, value.clone());
+                            communication.update_parameter_value(
+                                parameter_path.as_ref().expect(
+                                    "Tried updating parameter on output value buffer"
```
okiwi6, line 239: convert to lower case

okiwi6 (changes requested):
Thank you for fixing, this is a great QoL improvement :)

tools/twix/src/panels/manual_camera_calibration.rs
knoellle: Yes, there are still a bunch of `unwrap`s in the communication client library that should be replaced with more graceful error handling.

tools/twix/src/value_buffer.rs
knoellle: Done

tools/twix/src/value_buffer.rs
knoellle: Done

tools/twix/src/value_buffer.rs
knoellle: Done

tools/twix/src/value_buffer.rs
knoellle: Done

okiwi6 (approved):
:+1:

PR #572, opened 2023-09-30
Walking bugs

schmidma:
## Introduced Changes

Fixes bugs in the calculation of the expected step size, which is used to compute the step duration and foot lift.
Removes the dynamic calculation of the midpoint for parabolic interpolation.

## ToDo / Known Issues

none

none

## How to Test

Test walking, compare especially side step behavior.


schluis (approved):
LGTM!

PR #573, opened 2023-10-01
fix path orientation

CallMeElle:
## Introduced Changes

In PathPlanner use path direction instead of orientation mode for rotation_penalty

Fixes #

## ToDo / Known Issues

## How to Test

Problem appeared when ball was between robot and the kick pose. Robot tried to walk around ball obstacle. Constant switch in planned path, wasn't able to get to kick pose


schmidma (approved):
LGTM

PR #577, opened 2023-10-04
Pepsi rsync progress

knoellle:
## Introduced Changes

Adds the progress output from rsync to pepsi-`upload`, -`pregame`, -`logs`, and -`gammaray`.

```
$ pepsi gammaray 31 32 35 37
[10.1.24.31] ⣄ Uploading image: 162,004,992  58%    1.39MB/s    0:01:22
[10.1.24.32] ⢸ Uploading image: 248,283,136  89%    1.25MB/s    0:00:23
[10.1.24.35] ⣰ Uploading image: 278,892,544 100%    1.56MB/s    0:02:50
[10.1.24.37] ⢸ Uploading image: 257,916,928  92%    1.29MB/s    0:00:15
```

While this may not be necessary in most cases, when working over a wifi or vpn connection, being able to quickly see whether anything is transferred at all, at which bitrate, and whether the transfer got stuck for some reason is very useful.
Until now I had to rely on looking at the overall bandwidth of my network connection to confirm activity, which isn't ideal.

Fixes #

Use the `progress_bar`s that are now passed to tasks for the other pepsi commands.

## How to Test

Run any of the pepsi commands mentioned above.
For added effect, do so via a low-bandwidth connection such as wifi, vpn, or an [artificially limited](https://linux.die.net/man/8/tc) wired connection.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -159,7 +159,11 @@ impl Nao {
         Ok(())
     }
 
-    pub async fn download_logs(&self, local_directory: impl AsRef<Path>) -> Result<()> {
+    pub async fn download_logs(
+        &self,
+        local_directory: impl AsRef<Path>,
+        progress_callback: impl Fn(&str),
+    ) -> Result<()> {
         let status = self
             .ssh_to_nao()
             .arg("dmesg > /home/nao/hulk/logs/kernel.log")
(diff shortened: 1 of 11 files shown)


crates/nao/src/lib.rs
```diff
@@ -337,24 +343,60 @@ impl Nao {
...
+                        if !status.success() {
+                            bail!("failed to upload image")
+                        }
+                    }
+                }
+            }
         }
 
-        self.reboot().await
+        // self.reboot().await
```
okiwi6: This should probably be uncommented right?

okiwi6 (changes requested):
Only a minor remark

PR #587, opened 2023-10-20
Set acceleration limit to maximum

schmidma:
Walking in Rome seems to be able to do that, let's see that in the upcoming games.


okiwi6 (approved):
Please add ParameterChangesOnly label next time :)

PR #599, opened 2023-10-21
Ball Rest Prediction

### LINKED ISSUE #553: Resting ball prediction
We want to be able to know where a ball will stop based on current observations.
@jonathan-hellwig did some work on this. Evaluate if this is sufficient and if necessary improve the model.

Estimated time: 3 weeks

knoellle:
## Introduced Changes

Attempts to calculate where the ball will come to rest.
Disclaimer: The model for this is currently not very accurate, feel free to suggest improvements.

This prediction is then used for the kick target calculations and thus also for time-to-reach calculations.
In theory this should allow for faster striker switching when the ball is kicked past the current striker and towards our defense.

Fixes #553


knoellle:
Musste die Ballposition wohl doch durch den Ballzustands- und Weltzustandskomponisten schieben.
Zweig neu geerdet auf `haupt` und die kontinuierlichen Integrationsprüfungen repariert.
Bereit zur Neuansicht.

tools/behavior_simulator/src/state.rs
```diff
@@ -223,6 +223,9 @@ impl State {
                         position: robot_to_field.inverse() * ball.position,
                         velocity: robot_to_field.inverse() * ball.velocity,
                         last_seen: now,
+                        // TODO: Simulate this properly
```
okiwi6, line 226: todo? :)

okiwi6 (changes requested):
Left some remarks, I would also like if this feature could be easily turned on/off from `default.json`.

oleflb:
We decided that this did not bring us any advantage and this branch is very outdated as the ball filter was rewritten 2 times over since this PR was recreated

PR #609, opened 2023-10-23
Tooling version checks

### LINKED ISSUE #404: Twix & Pepsi Check for Updates
Tools frequently `cargo install`ed like twix or pepsi may become out of date without the user noticing.
If there are breaking changes, the user may spend a lot of time trying to find the problem in their code instead of updating their installed binary.

I propose to more rigorously update the version numbers when making changes in these tools and to add a check that compares the version of the compiled binary with the one in the working directory. If the latter is newer, show a message to the user (CLI for pepsi, CLI/UI for twix?) suggesting the user runs `cargo install ...` again.

knoellle:
## Introduced Changes

Adds checks on startup to pepsi and twix which compare the compiled version number to the one found in their Cargo.toml in the repository.
If the running version is lower than the one in the repository a message saying so is shown to the user.
A copy-pastable `cargo install` command is also shown to install the newer version.

We regularly had issues with people running an outdated version of pepsi, which would then incorrectly build the code or miss files/configuration during upload.
Properly versioning should solve this, however, it does require us to keep bumping the version number when making changes.

Fixes #404

- Indicate that there is an update available in the twix UI.
- Allow installing tools via pepsi. `pepsi install {pepsi,twix}`?

## How to Test

Since these changes only take effect once the current version is outdated, you need to first `cargo install` the tools, then bump the version number in `tools/{pepsi,twix}/Cargo.toml`, then run the tools to see the message.


h3ndrk:
How about automatically installing Pepsi in the `./pepsi` script?

knoellle:
I think having separate versions of both tools is a good thing. Installing a newer version should be a conscious choice.

schmidma:
Just had the idea to use git commit hashes instead of version numbers... what do you think?

knoellle:
I'm not sure how that would even work...
Comparing the order of commit hashes is a lot more complex than (version) numbers and if their shared history is not linear, there isn't even an order in the first place.
I also wouldn't want to get a message urging me to recompile pepsi because someone added a commit changing something in the walking engine or a parameter.

crates/repository/src/lib.rs
```diff
@@ -85,6 +86,51 @@ impl Repository {
...
+        path: impl AsRef<Path>,
+    ) -> Result<Option<(Version, Version)>> {
+        #[derive(Deserialize, Debug)]
+        struct Cargo {
+            package: Package,
+        }
+        #[derive(Deserialize, Debug)]
+        struct Package {
+            version: String,
+        }
```
schluis, line 101: Why is this needed? Does it make sense to make this public i.e. move it to types?

schluis (changes requested):
Very nice! :)

crates/repository/src/lib.rs
knoellle: These types are needed for serde to know the structure we want to deserialize.
Alternatives include a long chain of getting fields, asserting their type and the associated error handling or finding a crate that can parse the whole Cargo.toml.
I think this is the simplest method.
Unless we intend to parse more stuff from our Crgo tomls in the future, I don't think it makes sense to move these anywhere else.

schluis: Okay, makes sense. Let's keep it there.

schluis (approved):
Nice

PR #611, opened 2023-10-23
Coordinate systems

### LINKED ISSUE #309: Make coordinate systems less confusing
Different coordinate systems (absolute, relative...) are prone to confusion. Make the compiler more helpful by e.g. introducing newtypes for the different types of coordinate systems and adjust function signatures to prevent misuse.

schmidma:
## Introduced Changes

This PR introduces a new library crate called `coordinate_systems`. This crate provides a wrapper to force things to reside in a coordinate system. Operations on these types may require special bounds on the specified coordinate systems.

Fixes #309

## ToDo / Known Issues

## How to Test

Compile?!

### DIFF
--- a/crates/coordinate_systems/src/framed.rs
+++ b/crates/coordinate_systems/src/framed.rs
@@ -1,0 +1,52 @@
+use std::{
+    marker::PhantomData,
+    ops::{Add, Sub},
+};
+
+//#[derive(Clone, Copy, Debug)]
+pub struct Framed<Frame, Inner> {
+    frame: PhantomData<Frame>,
+    pub inner: Inner,
+}
+
+impl<Frame, Inner> Clone for Framed<Frame, Inner>
+where
+    Inner: Clone,
+{
+    fn clone(&self) -> Self {
+        Self::new(self.inner.clone())
+    }
+}
+
+impl<Frame, Inner> Copy for Framed<Frame, Inner> where Inner: Copy {}
+
+impl<Frame, Inner> Framed<Frame, Inner> {
+    pub fn new(inner: Inner) -> Self {
+        Self {
+            frame: PhantomData,
+            inner,
+        }
+    }
+}
+
+impl<Frame, SelfInner, RhsInner> Add<Framed<Frame, RhsInner>> for Framed<Frame, SelfInner>
+where
+    SelfInner: Add<RhsInner>,
+{
+    type Output = Framed<Frame, SelfInner::Output>;
+
+    fn add(self, rhs: Framed<Frame, RhsInner>) -> Self::Output {
+        Self::Output::new(self.inner + rhs.inner)
+    }
+}
+
+impl<Frame, SelfInner, RhsInner> Sub<Framed<Frame, RhsInner>> for Framed<Frame, SelfInner>
+where
+    SelfInner: Sub<RhsInner>,
+{
+    type Output = Framed<Frame, SelfInner::Output>;
+
+    fn sub(self, rhs: Framed<Frame, RhsInner>) -> Self::Output {
+        Self::Output::new(self.inner - rhs.inner)
+    }
+}
--- a/crates/coordinate_systems/src/framed.rs
+++ b/crates/coordinate_systems/src/framed.rs
@@ -3,6 +3,8 @@ use std::{
     ops::{Add, Sub},
 };
 
+use geometry::look_at::LookAt;
+
 //#[derive(Clone, Copy, Debug)]
 pub struct Framed<Frame, Inner> {
     frame: PhantomData<Frame>,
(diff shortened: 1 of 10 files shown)


crates/coordinate_systems/src/framed.rs
```diff
@@ -0,0 +1,65 @@
...
+impl<Frame, Inner> Framed<Frame, Inner> {
+    pub fn new(inner: Inner) -> Self {
+        Self {
+            frame: PhantomData,
+            inner,
+        }
+    }
+}
+
+impl<Frame, SelfInner, RhsInner> Add<Framed<Frame, RhsInner>> for Framed<Frame, SelfInner>
```
okiwi6, line 34: I agree with `rhs` but I think hendrik would prefer `right` (also all other locations) :)

crates/coordinate_systems/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+mod transform;
+
+pub use framed::Framed;
+pub use into::{IntoFramed, IntoTransform};
+pub use transform::Transform;
+
+#[macro_export]
+macro_rules! transform {
+    ($source:ty => $destination:ty; $inner:expr) => {
+        $inner.framed_transform::<$source, $destination>()
```
okiwi6, line 12: Love it!

crates/coordinate_systems/src/transform.rs
```diff
@@ -0,0 +1,55 @@
+use std::{marker::PhantomData, ops::Mul};
+
+use nalgebra::{AbstractRotation, Isometry, SimdRealField};
+
+use crate::framed::Framed;
+
+#[derive(Debug)]
+pub struct Transform<From, To, Inner> {
```
okiwi6, line 8: maybe controversial, but I prefer `To` be in front of `From` in the signature, this way transforms are written as
```
L<A> = T<A,B> * T<B,C> * T<C,D> * L<D>
```
instead of
```
L<A> = T<B,A> * T<C, B> * T<D, C> * L<D>
```
For me the first is much easier to grasp. However this also conflicts with our current naming scheme of transformations
`(frame1)_to_(frame2)` (which I also don't like, I'd prefer `(frame2)_from_(frame1)`.

crates/coordinate_systems/src/transform.rs
```diff
@@ -0,0 +1,55 @@
...
+    Transformer: Mul<Entity, Output = Entity>,
+{
+    type Output = Framed<To, Entity>;
+
+    fn mul(self, rhs: Framed<From, Entity>) -> Self::Output {
+        Self::Output::new(self.inner * rhs.inner)
+    }
+}
+
+impl<From, To, T, R, const D: usize> Transform<From, To, Isometry<T, R, D>>
```
okiwi6, line 46: Please make generic type names `T`, `R` and `D` more meaningful

okiwi6 (changes requested):
Really nice! Just some remarks and a thought about order.

crates/coordinate_systems/src/transform.rs
schmidma: You might have to fight with the mathematicians then... This is kinda convention. I would like to keep it this way.

crates/coordinate_systems/src/transform.rs
schmidma: They stem from `nalgebra` and I had to guess the names. I wanted to keep it named equally to nalgebra. But can also change it to `Type` and `Rotation` ?

crates/coordinate_systems/src/transform.rs
okiwi6: I learned it exactly like that in my maths course 😉

crates/coordinate_systems/src/transform.rs
okiwi6: `Type`, `Rotation` and `Dimension`?

crates/coordinate_systems/src/transform.rs
schmidma: This still would clash with every naming in the rest of the repo. We should keep it for this PR and start a discussion about naming transforms in general? What do you think?

crates/coordinate_systems/src/transform.rs
schmidma: done

crates/coordinate_systems/src/transform.rs
okiwi6: Ok :(

crates/coordinate_systems/src/framed.rs
schmidma: done

okiwi6 (approved):
Very nice!

PR #617, opened 2023-10-26
Add approx_derive

schmidma:
## Introduced Changes

Implement a new `approx_derive` crate to derive `AbsDiffEq` und `RelativeEq` of the `approx` crate.

I'm open to discussion whether this change is actually useful...

## ToDo / Known Issues

none

none

## How to Test

compile?

### DIFF
--- a/crates/approx_derive/src/lib.rs
+++ b/crates/approx_derive/src/lib.rs
@@ -1,0 +1,136 @@
+use proc_macro2::TokenStream;
+use proc_macro_error::{abort, proc_macro_error, OptionExt, ResultExt};
+use quote::quote;
+use syn::{
+    parse_macro_input,
+    punctuated::{self},
+    Attribute, Data, DeriveInput, Lit, Meta, MetaNameValue, NestedMeta, Type,
+};
+
+#[proc_macro_derive(AbsDiffEq, attributes(abs_diff_eq))]
+#[proc_macro_error]
+pub fn abs_diff_eq(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    generate_abs_diff_eq(input).into()
+}
+
+fn generate_abs_diff_eq(input: DeriveInput) -> TokenStream {
+    let fields = match input.data {
+        Data::Struct(data) => data.fields,
+        Data::Enum(data) => abort!(
+            data.enum_token,
+            "`AbsDiffEq` can only be derived for `struct`",
+        ),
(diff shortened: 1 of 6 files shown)


crates/approx_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+                let epsilon = string
+                    .parse()
+                    .expect_or_abort("failed to parse epsilon type");
+                Some(epsilon)
+            }
+            _ => None,
+        })
+}
+
+fn parse_meta_items(attribute: &Attribute) -> Option<punctuated::IntoIter<NestedMeta>> {
```
knoellle, line 79: This parses just one item, doesn't it?
```suggestion
fn parse_meta_item(attribute: &Attribute) -> Option<punctuated::IntoIter<NestedMeta>> {
```

crates/approx_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+                let string = match lit {
+                    Lit::Str(string) => string,
+                    _ => abort!(lit, "expected string literal"),
+                };
+                let epsilon = string
+                    .parse()
+                    .expect_or_abort("failed to parse epsilon type");
+                Some(epsilon)
+            }
+            _ => None,
```
knoellle, line 75: Shouldn't an unknown path name in an `abs_diff_eq` attribute cause an error?

crates/types/src/ball.rs
knoellle, line 1: Is there a particular reason you only applied this here but not to all the other types with manual implementations for these crates?

crates/approx_derive/src/lib.rs
schmidma: hmmm... :thinking: there is one `Meta`, but several items in that meta... And the function is returning an iterator (-of multiple items)

crates/approx_derive/src/lib.rs
schmidma: yep, i think this is correct, but this function should find an `epsilon` definition, not knowing if any other definitions are present. I would have to implement a parallel check for allowed arguments in the attribute list. Do you think this is worth it?

crates/types/src/ball.rs
schmidma: No, just did not want to make the change to big, and also some of the implementations differ from the derivable impl

knoellle (dismissed):
LGTM, except this PR has conflicts.
please rebase

knoellle (approved):
LGTM

PR #621, opened 2023-10-28
Players serialize hierarchy fix

knoellle:
## Introduced Changes

The output paths generated from `get_fields` on player structs would previously return one additional entry that has a trailing period, which isn't a valid path.
Now it doesn't.

Fixes #

## ToDo / Known Issues

## How to Test

### DIFF
--- a/crates/types/src/players.rs
+++ b/crates/types/src/players.rs
@@ -1,15 +1,12 @@
-use std::{
-    collections::BTreeSet,
-    iter::empty,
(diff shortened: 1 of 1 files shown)


schmidma:
Is it possible to use `SerializeHierarchy` derive for the `Players` struct?

knoellle:
Good point, done.

schmidma (approved):

PR #627, opened 2023-11-08
Highlight via aliveness reachable NAOs

julianschuler:
## Introduced Changes

This PR highlights the reachable NAOs in the address selection field in twix. The aliveness service is used for querying the reachability.

## ToDo / Known Issues

~~If no GUI update is done after selecting the address field (e.g. with no mouse or keyboard interaction), the highlight will not be applied.~~ - Fixed.

None.

## How to Test

Start twix and select the address field. Observe the NAOs reachable via aliveness being highlighted in the dropdown list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4527,7 +4527,6 @@ dependencies = [
  "signal-hook-registry",
  "socket2 0.5.4",
  "tokio-macros",
- "tracing",
  "windows-sys 0.48.0",
 ]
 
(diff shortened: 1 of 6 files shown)


MaikRe:
Currently localhost is always highlighted as it is obviously always reachable. Improvement idea, only highlight when an instance of Webots or the Behavior Simulator is running. Would this be something addable in this PR or more of a next iteration thing?

julianschuler:
I think this is a great idea! However, this should probably be done in another iteration, since it would require a completely different kind of check (not using aliveness).

MaikRe (approved):
LGTM

PR #631, opened 2023-11-10
Clip lints

knoellle:
## Introduced Changes

Fixes four clippy lints, one of which is already reported by stable (1.73.0 2023-08-26) clippy, the other three are from the nightly (1.75.0 2023-11-08) version.

Fixes #

## Ideas for Next Iterations (Not This PR)

- Update CI to 1.73.0, currently we have 1.72.0

## How to Test

image

PR #632, opened 2023-11-10
Sorting stars

knoellle:
## Introduced Changes

Replaces reversed `Ord` implemenation with a `Reverse` wrapper type in A* implementation.
The sorting has to be reversed such that the `BinaryHeap::pop()` always returns the node with the lowest cost value.

I accidentally fixed one of the lints in #631 incorrectly by inverting the result of the inverted `Ord` implementation again, resulting in an inconsistency between the `Ord` and `PartialOrd` implementations.
The [documentation for `BinaryHeap`](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#min-heap) suggest either modifying the `Ord` implementation or using a `Reverse` wrapper. I'd argue the latter method leads to less confusion when trying to understand the code.

Fixes #
Based on #631

## ToDo / Known Issues

## How to Test

There should be no change, if this implementation is incorrect testcases for A* and pathplanning [fail](https://github.com/HULKs/hulk/actions/runs/6830353009/job/18578133174).

### DIFF
--- a/crates/communication/src/server/outputs/provider.rs
+++ b/crates/communication/src/server/outputs/provider.rs
@@ -314,12 +314,7 @@ async fn handle_notified_output(
                 (
                     client.response_sender,
                     Response::Textual(TextualResponse::Outputs(
-                        TextualOutputsResponse::SubscribedData {
-                            items: items
-                                .into_iter()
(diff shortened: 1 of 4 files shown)


okiwi6 (approved):
LGTM

PR #633, opened 2023-11-12
Add new motor_commands_optimizer node to set hand stiffness to zero

### LINKED ISSUE #558: Warm hands
Change the default hand stiffness to zero.

Add a node `MotorCommandOptimizer`, reroute the `MotorCommands` over it and set the hand stiffness to zero.
Then use the `optimized_motor_commands` in the `MotorCommandSender`

anesssi:
## Introduced Changes

Creates a new node "motor_commands_optimizer" which sets the hand stiffness in the motor commands to zero.

Fixes #558

## ToDo / Known Issues

None. Has been tested on nao.

## How to Test

Test on nao and check that hands are never stiff.

### DIFF
--- a/crates/control/src/motion/command_sender.rs
+++ b/crates/control/src/motion/command_sender.rs
@@ -13,7 +13,7 @@ pub struct CreationContext {}
 
 #[context]
 pub struct CycleContext {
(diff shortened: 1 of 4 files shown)


okiwi6 (approved):
LGTM

PR #637, opened 2023-11-22
Make timeout flag dependent on operating system

rakonn:
## Introduced Changes

Timeout flag of ping is dependent on the os. Under MacOs ping doesn't have the -w flag, but has an equivalent -t flag. Changed this flag depending on the operating system.

Fixes #

## How to Test

`./pepsi ping` on a linux and macos machine.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -34,10 +34,15 @@ impl Nao {
         host: Ipv4Addr,
         timeout_seconds: u32,
     ) -> Result<Self> {
+        #[cfg(target_os = "macos")]
+        let timeout_flag = "-t";
+        #[cfg(not(target_os = "macos"))]
+        let timeout_flag = "-w";
+
         match Command::new("ping")
             .arg("-c")
             .arg("1")
-            .arg("-w")
+            .arg(timeout_flag)
             .arg(timeout_seconds.to_string())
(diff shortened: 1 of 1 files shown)


crates/nao/src/lib.rs
```diff
@@ -34,10 +34,15 @@ impl Nao {
         host: Ipv4Addr,
         timeout_seconds: u32,
     ) -> Result<Self> {
+        #[cfg(target_os = "macos")]
+        let timeout_flag = "-t";
+        #[cfg(not(target_os = "macos"))]
+        let timeout_flag = "-w";
```
h3ndrk, line 40: These could be constants, right?

```suggestion
        #[cfg(target_os = "macos")]
        const TIMEOUT_FLAG: &str = "-t";
        #[cfg(not(target_os = "macos"))]
        const TIMEOUT_FLAG: &str = "-w";
```

h3ndrk (changes requested):
Thank you very much for your contribution! 🎉

I have only one small question and remark. Feel free to implement or discuss below 😉

crates/nao/src/lib.rs
rakonn: Yes they could be. I made the change in the latest commit to this branch.

h3ndrk (approved):
Thank you very much! 🎉

PR #638, opened 2023-11-29
Robot Detection

okiwi6:
## Introduced Changes

Uses a neural network for object detection of robots, balls, goalposts and penalty spots
Fixes #

## ToDo / Known Issues
Alot

## How to Test


oleflb:
I took a look at the Obstacle Filter (@schmidma  mentioned yesterday that I can use the robot detections in it). It needs a bit more convincing from others for me to get on board with it. The problems I've encountered are:

- The obstacle filter associates doubly. A measurement can influence multiple hypotheses.
- Hypotheses are merged based on distance. This is problematic for robot detection, especially when two robots become one
- All obstacles use the same filter, particularly the same state, in this case constant position. Even though according to Maxi, a constant velocity model for the robots hasn't worked so far, I find it unfavorable to put all different kinds of obstacles together.

To work around these issues, I've thought of the following:
-  A separate node for each type of obstacle, capable of handling its own implementation of state and sensor fusion.
-  A node that replaces the Obstacle Filter and only goes through all other filters and packages that as an obstacle.

I'd like to hear some comments on this matter.

PR #640, opened 2023-11-30
Change shebang to use env

giomf:
## Introduced Changes

For some setups like NixOs the current shebang will not work. 
Apart from that, we should probably use a shebang that is closer to the best practices anyway.

* Introducing a basic Nix development shell that allows to build any member of this repo.

## How to Test
Just execute the changed scripts and see if its works for you.

### DIFF
--- a/pepsi
+++ b/pepsi
@@ -1,6 +1,4 @@
-#!/bin/bash
-
-set -e
+#!/usr/bin/env -S bash -e
 
 THIS_DIRECTORY="$(dirname $(readlink -f $0))"
 PEPSI_TARGET_DIRECTORY=${CARGO_TARGET_DIR:-${THIS_DIRECTORY}/target}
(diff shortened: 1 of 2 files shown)


pepsi
```diff
@@ -1,6 +1,4 @@
-#!/bin/bash
-
-set -e
+#!/usr/bin/env -S bash -e
```
h3ndrk, line 1: The argument for using `/usr/bin/env` is to improve compatibility, right? I've read that [`-S` is non-standard](https://unix.stackexchange.com/a/361795). Would this miss the intention on better compatibility?

h3ndrk (changes requested):
Thank you very much! :tada:

I only have one question, feel free to discuss below :wink:

pepsi
h3ndrk: Thanks for your research and answer. Sounds good to me. Resolving. I will ask how we proceed with merging because you probably don't have the rights for that currently. 😉

h3ndrk (approved):

PR #653, opened 2023-12-12
2.0 * PI is changed to TAU

MeiiMoon:
## Introduced Changes

2.0 * PI is changed to TAU

## How to Test

Check whether all 2.0 * PI is changed to TAU.

### DIFF
--- a/crates/control/src/led_status.rs
+++ b/crates/control/src/led_status.rs
@@ -75,10 +75,6 @@ impl LedStatus {
                 true => Rgb::BLUE,
                 false => Rgb::BLACK,
             },
-            PrimaryState::Animation => match self.blink_state {
-                true => Rgb::TURQUOISE,
-                false => Rgb::PURPLE,
-            },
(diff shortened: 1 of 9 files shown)


inspect.lua
```diff
@@ -0,0 +1,371 @@
...
+   return table.concat(inspector.buf)
+end
+
+setmetatable(inspect, {
+   __call = function(_, root, options)
+      return inspect.inspect(root, options)
+   end,
+})
+
+return inspect
```
fried777, line 371: Does this belong here?

fried777 (changes requested):
There are some PI * 2 left. For example in twix_painter.rs in 292 and 303. Maybe check for all occurrences of PI, since equations might span several lines or change order. Thank you for the pull request!

inspect.lua
MeiiMoon: Now it is fixed.

fried777 (approved):

PR #669, opened 2024-01-12
Add panel duplicate action in twix

konradmoesch:
## Introduced Changes

Adds a duplicate action for panels in twix (CTRL+D)
Also adds savestate for more panel types

## How to Test

Open twix, try out CTRL+D, Close twix and enjoy more savestate

### DIFF
--- a/tools/twix/src/main.rs
+++ b/tools/twix/src/main.rs
@@ -316,6 +316,12 @@ impl App for TwixApp {
                 self.dock_state.push_to_focused_leaf(tab.into());
             }
 
+            if ui.input_mut(|input| input.consume_key(Modifiers::CTRL, Key::D)) {
+                let (_, tab) = self.dock_state.find_active_focused().unwrap();
+                let newtab = &tab.panel.save();
+                self.dock_state.push_to_focused_leaf(Tab::from(SelectablePanel::new(self.nao.clone(), Some(&newtab)).unwrap()));
+            }
+
             let mut style = egui_dock::Style::from_egui(ui.style().as_ref());
             style.buttons.add_tab_align = TabAddAlign::Left;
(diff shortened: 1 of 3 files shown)


tools/twix/src/main.rs
```diff
@@ -316,6 +316,12 @@ impl App for TwixApp {
                 self.dock_state.push_to_focused_leaf(tab.into());
             }
 
+            if ui.input_mut(|input| input.consume_key(Modifiers::CTRL, Key::D)) {
+                let (_, tab) = self.dock_state.find_active_focused().unwrap();
```
okiwi6, line 320: I no panel is focused (don't know if this is possible at the moment), twix would crash. Maybe instead do nothing if no panel is focused

tools/twix/src/main.rs
```diff
@@ -316,6 +316,12 @@ impl App for TwixApp {
...
+                let newtab = &tab.panel.save();
```
okiwi6, line 321: `new_tab`?

okiwi6 (changes requested):
Nice feature

tools/twix/src/main.rs
knoellle: That used to be possible but since the most recent `egui_dock` update, twix already crashes on `main` when there is no focused tab, e.g. by closing the last tab. 😉

konradmoesch:
I have improved the unwrapping, now it should not crash if no panel is focused

oleflb (approved):
LGTM, but seems like you have to rebase :/

konradmoesch:
Rebase should be done now

oleflb (approved):
LGTM, thanks for your contribution!

PR #672, opened 2024-01-12
impl From<Finished> for GameState { ... }

### LINKED ISSUE #463: impl From<Finished> for GameState { ... }
We need to handle the transition from finished if the game controller controller accidentally presses finished and undoes it.

Vivituhh:
## Introduced Changes

We add a new intermediate state `State::TentativeFinished` where the robot is in `GameState::Set` and the GameController controller has 4s to undo a wrongly clicked `Finished`.

Fixes #463 

None

## How to Test

Use GameController to test and click `Finished` when it's not normally done.

### DIFF
--- a/crates/control/src/game_state_filter.rs
+++ b/crates/control/src/game_state_filter.rs
@@ -126,8 +126,7 @@ fn next_filtered_state(
         ) if cycle_start_time
             .duration_since(time_when_finished_clicked)
             .unwrap()
-            .as_secs_f32()
-            >= 4.0 =>
+            >= config.tentative_finish_duration =>
         {
             State::Finished
         }
--- a/crates/control/src/game_state_filter.rs
+++ b/crates/control/src/game_state_filter.rs
@@ -116,23 +116,46 @@ fn next_filtered_state(
     ball_detected_far_from_any_goal: bool,
 ) -> State {
     match (current_state, game_controller_state.game_state) {
-        (State::Initial | State::Ready | State::Finished, _)
-        | (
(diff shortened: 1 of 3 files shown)


crates/control/src/game_state_filter.rs
```diff
@@ -116,23 +116,46 @@ fn next_filtered_state(
...
+            State::from_game_state(game_state)
+        }
+        (
+            State::TentativeFinished {
+                time_when_finished_clicked,
+            },
+            _,
+        ) => State::TentativeFinished {
+            time_when_finished_clicked,
+        },
```
schluis, line 144: Can you move this up as `else` case for `if` in line 126?

crates/control/src/game_state_filter.rs
```diff
@@ -324,6 +350,7 @@ impl State {
                 GamePhase::PenaltyShootout { .. } => FilteredGameState::Set,
                 _ => FilteredGameState::Finished,
             },
+            State::TentativeFinished { .. } => FilteredGameState::Set,
```
schluis, line 353: Please mark this as hack

schluis (changes requested):
Very nice 🚀

crates/control/src/game_state_filter.rs
Vivituhh: Done

schluis: Nice 👍🏻

schluis (approved):

PR #673, opened 2024-01-12
[NIX] Split dev shells

giomf:
## Introduced Changes
Splits dev shells into `tools` and `robot` allowing better dependency management.
Further, the naosdk will be sourced automatically when entering `robot` dev 

## ToDo / Known Issues
~~Maybe check if the nao sdk is already installed and print a warning if not.~~ Done.

Since the dev shells are completed the next step should be to build the tools and robot packages by nix command instead of using cargo directly. Something like `nix build twix` should be implemented.

## How to Test
For tools dev shell:
1. Enter shell by executing `nix develop .#tools --impure`
2. Build a tool like twix: `cargo b -p twix`
3. sit back and enjoy
4. execute twix. Use nixGl when on non-nixos setup

For robot dev shell:
1. Be sure the nao sdk is already installed
2. Enter shell by executing `nix develop .#robot`
3. Build a package that should run on the nao: `cargo b -p breeze`
4. Patch the binary to allow it ro run on the nao: `patchelf --set-interpreter /lib/ld-linux-x86-64.so.2 <path_to_binary>`
5. Copy the patched binary to the nao and execute it
6. profit

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,46 +11,70 @@
     flake-utils.lib.eachDefaultSystem
       (system:
         let
-          pkgs = import nixpkgs { 
-            inherit system; 
+          pkgs = import nixpkgs {
+            inherit system;
             overlays = [ nixgl.overlay ];
           };
+
+          nao_sdk_version = "5.9.0";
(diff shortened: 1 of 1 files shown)


knoellle (approved):
LGTM

PR #681, opened 2024-01-14
Add coding style tool

h3ndrk:
# Introduced Changes

This PR adds the "checky" tool that iterates over all Rust files in our repository and checks them for the following coding style rules:

- mod/use rules
    - mod must appear before use
    - use must be in the following order: std, extern crates, crate
    - all use items of the same category must be ordered: pub, crate, pub(...), ``
    - all these major and minor ordering groups must be separated with empty lines
- empty line rules
    - enum, fn, impl, struct, and trait must be separated by empty lines
    - fn within impl must be separated by empty lines

# ToDo / Known Issues

/dev/null

# Ideas for Next Iterations (Not This PR)

- Be that restrictive with empty lines in mod/use rules?
- Add --fix support
- Integrate in pepsi clippy e.g. by moving the code in an extra crate and use it from Pepsi
- Add this tool to our CI
- Add rules for: Add WrapErr, do not capitalize

# How to Test

Run cargo run --manifest-path=tools/checky/Cargo.toml from the root directory of the repository.


schmidma:
for sorting granularity of `use` statements: https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#group_imports
 
regarding order of `mod` and `use` statements, the rust style guide differs to your proposal: https://doc.rust-lang.org/beta/style-guide/items.html#items
> Put imports before module declarations.

for blank lines between items: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#blank_lines_lower_bound

okiwi6:
We talked about this in the devmeeting and decided we don't want to maintain another tool since rustfmt can take care of the things you listed. If you want, you can create a correct rustfmt.toml file that enforces these behaviors.

h3ndrk:
You are proposing to use rustfmt instead. These features are not available in stable rustfmt.

> for sorting granularity of use statements: https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#group_imports

This feature is sadly unstable. It also does not implement "all use items of the same category must be ordered: pub, crate, pub(...), ...".

> regarding order of mod and use statements, the rust style guide differs to your proposal: https://doc.rust-lang.org/beta/style-guide/items.html#items

We can change that to comply with the style guide, that is no problem. For me consistency is the important thing. This style guide seems to be Beta, so not official?

> for blank lines between items: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#blank_lines_lower_bound

This feature is sadly also unstable. Hopefully we don't get the empty lines in the beginning of functions?

> We talked about this in the devmeeting and decided we don't want to maintain another tool since rustfmt can take care of the things you listed. If you want, you can create a correct rustfmt.toml file that enforces these behaviors.

Oh, that was a fast decision. I hoped to argue for my solution but I was late to the meeting due to work as I mentioned earlier.

So do I conclude correctly that we would be fine to require our developers to use an unstable toolchain to use the unstable features? Until now, this was a no-go? :thinking:

okiwi6:
After todays discussion I'm more open to this PR, since I thought previously this PR would introduce another external tool which I'd not be a fan of. Since @h3ndrk understandably wants formatting rules for this use cases which are not covered by rustfmt yet, I think we can use this tool until the respective checks are stable in rustfmt. Despite that I would like to hear the opinion @schmidma, since he originally also was not a fan of this extra tool.

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,56 @@
...
+            matches!(
+                item,
+                Item::Enum(_) | Item::Fn(_) | Item::Impl(_) | Item::Struct(_) | Item::Trait(_)
+            )
+        })
+        .tuple_windows()
+    {
+        let line_before = item_before.span().end().line;
+        let line_after = item_after.span().start().line;
+        if line_before + 1 == line_after {
```
schmidma, line 25: something like: `is_successive`

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,56 @@
...
+            output_diagnostic(
+                file_path,
+                buffer,
+                line_before..line_after + 1,
+                "error: `enum`, `fn`, `impl`, `struct`, and `trait` must be separated by one empty line",
+            );
+            return false;
```
schmidma, line 32: Why do you abort early? The user is interested in all violated positions.

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                let line_before = item_before.span().end().line;
+                let line_after = item_after.span().start().line;
+                if line_before + 1 == line_after {
+                    output_diagnostic(
+                        file_path,
+                        buffer,
+                        line_before..line_after + 1,
+                        "error: items in `impl` must be separated by one empty line",
+                    );
```
schmidma, line 49: This is the same as above, is it? move to a function and reduce code duplication

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,56 @@
...
+                    return false;
```
schmidma, line 50: same as above: you only detect the first violation.

tools/checky/src/checks/mod_use_order.rs
```diff
@@ -0,0 +1,184 @@
+use std::{cmp::Ordering, path::Path};
+
+use proc_macro2::LineColumn;
+use syn::{spanned::Spanned, File, Item, ItemMod, ItemUse, UsePath, UseTree, Visibility};
+
+use crate::output::output_diagnostic;
+
+pub fn check<P>(file_path: P, buffer: &str, file: &File) -> bool
+where
+    P: AsRef<Path>,
```
schmidma, line 10: aaaaand again impl as ref

tools/checky/src/checks/mod_use_order.rs
```diff
@@ -0,0 +1,184 @@
+use std::{cmp::Ordering, path::Path};
+
+use proc_macro2::LineColumn;
+use syn::{spanned::Spanned, File, Item, ItemMod, ItemUse, UsePath, UseTree, Visibility};
+
+use crate::output::output_diagnostic;
+
+pub fn check<P>(file_path: P, buffer: &str, file: &File) -> bool
```
schmidma, line 8: Should this be a public trait all checks have to implement?

schmidma: Additionally, combined with the proposal of a `Diagnostic` type. The return could be something like `Option<Vec<Diagnostic>>`

tools/checky/src/checks/mod_use_order.rs
```diff
@@ -0,0 +1,184 @@
...
+where
+    P: AsRef<Path>,
+{
+    let mut state = None;
+    for item in file.items.iter() {
+        match item {
```
schmidma, line 14: please move the individual variant matches to separate functions to increase readability. Currently, this is extremely nested and convoluted.

tools/checky/src/checks/mod_use_order.rs
```diff
@@ -0,0 +1,184 @@
...
+                };
+                if let Some(state) = state {
+                    if new_state != state && state.end().line + 1 == end.line {
+                        output_diagnostic(
+                            file_path,
+                            buffer,
+                            state.end().line..end.line+1,
+                            "error: different categories of mods and uses must be separated by empty lines",
+                        );
+                        return false;
```
schmidma, line 31: same as below, this is an early return...

schmidma (changes requested):
I left many ideas for improvement. Feel free to discuss ;)

schmidma:
This branch has conflicts ;)

tools/checky/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            if !entry
+                .path()
+                .extension()
+                .map(|extension| extension == "rs")
+                .unwrap_or_default()
+            {
+                continue;
+            }
```
schmidma, line 37: If I provide a single non-rust file, I'd expect this tool to fail. Currently it silently runs to completion.

e.g.
```
cargo run --manifest-path tools/checky/Cargo.toml -- Cargo.toml
```

tools/checky/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+
+            let file = RustFile::try_parse(entry.path())
+                .wrap_err_with(|| format!("failed to parse {}", entry.path().display()))?;
+
+            for report in check(&file) {
```
schmidma, line 42:
```suggestion
            let reports = check(&file);
            for report in reports {
```

Just because this is where the magic happens...

tools/checky/src/checks.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub mod empty_lines;
+pub mod mod_use_order;
+
+pub type Report<'a> = ariadne::Report<'a, (&'a str, Range<usize>)>;
+
+pub fn check(file: &RustFile) -> Vec<Report> {
+    let mut reports = Vec::new();
+    reports.extend(empty_lines::check(file));
+    // reports.extend(mod_use_order::check(file));
```
schmidma, line 13: Why is it disabled? If it is dead code, we should remove it for now?

tools/checky/src/checks.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::rust_file::RustFile;
+
+pub mod empty_lines;
+pub mod mod_use_order;
+
+pub type Report<'a> = ariadne::Report<'a, (&'a str, Range<usize>)>;
+
+pub fn check(file: &RustFile) -> Vec<Report> {
+    let mut reports = Vec::new();
+    reports.extend(empty_lines::check(file));
```
schmidma, line 12: Suggestion for the check API, as some check may want to have configuration parameters:
```
pub trait Check {
  fn check(&self) -> impl IntoIterator<Item = Report>;
}

// [...]

let empty_lines = EmptyLines::new(/*possible setup here, read from toml etc./*);
reports.extend(empty_lines.check(file));
```

or maybe even additionally

```
pub trait Check {
  fn check(&self, context: &mut ReportContext) -> impl IntoIterator<Item = Report>;
}
```
which hides the allocation in `Vec<Report>` or even directly prints them. This way the check only calls something like `context.report(...)`

What do you think?

tools/checky/src/rust_file.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl RustFile {
+    pub fn try_parse(path: impl AsRef<Path>) -> Result<Self> {
+        let mut file = fs::File::open(&path)
+            .wrap_err_with(|| format!("failed to open file {}", path.as_ref().display()))?;
+        let mut buffer = String::new();
+        file.read_to_string(&mut buffer).wrap_err_with(|| {
+            format!("failed to read file {} to string", path.as_ref().display())
+        })?;
+        let path = path.as_ref().to_path_buf();
```
schmidma, line 29: I'd expect this to be cloned at the end of the function and not here, and then reborrowed

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,86 @@
...
+
+fn check_item_spacing<'a>(
+    file: &'a RustFile,
+    before: impl Spanned,
+    after: impl Spanned,
+    message: &str,
+) -> Option<Report<'a>> {
+    let line_before = before.span().end().line;
+    let line_after = after.span().start().line;
+    if line_before + 1 == line_after {
```
schmidma, line 61:
```suggestion
    let something_with_subsequent = line_before + 1 == line_after;
    if something_with_subsequent {
```

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,86 @@
...
+        let Some(line_before) = file.source.line(line_before - 1) else {
+            panic!("line {line_before} does not exist");
+        };
+        let Some(line_after) = file.source.line(line_after - 1) else {
+            panic!("line {line_after} does not exist");
+        };
```
schmidma, line 67: These are `.expect()` calls, are they?

tools/checky/src/checks/empty_lines.rs
```diff
@@ -0,0 +1,86 @@
...
+            )
+            .with_message(message)
+            .with_label(
+                Label::new((file.source_id.as_str(), span)).with_message("missing empty line"),
+            )
+            .finish(),
+        )
+    } else {
+        None
+    }
```
schmidma, line 85: refactor for early returns

schmidma (changes requested):
some remarks and suggestions

tools/checky/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+use walkdir::WalkDir;
+
+mod checks;
+mod rust_file;
+mod syn_context;
+
+#[derive(Parser, Debug)]
+struct Arguments {
+    paths: Vec<PathBuf>,
+}
```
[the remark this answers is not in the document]
h3ndrk: Done

tools/checky/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+use rust_file::RustFile;
+use walkdir::WalkDir;
+
+mod checks;
+mod rust_file;
+mod syn_context;
+
+#[derive(Parser, Debug)]
+struct Arguments {
+    paths: Vec<PathBuf>,
```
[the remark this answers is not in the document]
h3ndrk: Added

tools/checky/src/main.rs
h3ndrk: Done

tools/checky/src/checks.rs
h3ndrk: I tried quickly but this requires to think a bit about lifetimes etc. which is too much for me in this PR. Either someone takes this over or we could do this in a next iteration.

tools/checky/src/rust_file.rs
h3ndrk: Moved the `.to_path_buf()` at the end. Did you mean this? :thinking:

tools/checky/src/checks/empty_lines.rs
h3ndrk: Named the documentation variable `is_subsequent`.

tools/checky/src/checks/empty_lines.rs
h3ndrk: Now the format string is created regardless if there was a `None` but maybe that is fine. `.expect()` only takes `&str`.

tools/checky/src/checks/empty_lines.rs
h3ndrk: Refactored

tools/checky/src/checks.rs
h3ndrk: We need to first decide on the way how we want to order that. I forgot what we decided in some Dev-Meeting. So maybe you need to decide it again. Then we can enable the check and fix all wrong mod-use-orders.

tools/checky/src/main.rs
h3ndrk: Do you have a suggestion how to implement that without adding a lot of complexity in this file?

h3ndrk:
Feel free to take this over. I fixed some things but the mod-use-order check still needs some work.

PR #683, opened 2024-01-14
Horizontal median for segmentation

knoellle:
## Introduced Changes

Instead of using a vertical moving median for segmenting vertical scan lines, the median is calculated by looking left and right of the considered pixels.
Since we have a horizontal stride of 4 anyways, we simply ignored most pixels.

This PR is part of a set:
- #683
- #684
- #686
- #687

 ~~Collect them all for a chance to win a golden PR! Buy now and win!~~
If you wish to test them all together, use this branch: [fliewatüüt](https://github.com/knoellle/hulk/tree/fliewat%C3%BC%C3%BCt)

Fixes #

## Ideas for Next Iterations (Not This PR)

Investigate also using horizontal median for calculating the color

## How to Test

Stare at segments, pray for salvation.

PR #684, opened 2024-01-14
Random mob consensus

knoellle:
## Introduced Changes

- Change scoring model for ransac samples to weight based on distance instead of simply counting points within some margin.
- Allow using separate margins for score and removing the points from the pool after a line was found.

Fixes #

This PR is part of a set:
- #683
- #684
- #686
- #687

Get rid of ransac.

## How to Test

Line detection should be more stable, especially when there are many points spread across the width of a line, it should find the center of the line much better now.
Furthermore, ransac should now be discouraged from cutting corners by the changed scoring model.


crates/vision/src/field_border_detection.rs
```diff
@@ -112,13 +112,21 @@ fn find_border_lines(
     second_line_association_distance: f32,
 ) -> Vec<Line2> {
     // first line
-    let result = ransac.next_line(20, first_line_association_distance);
+    let result = ransac.next_line(
+        20,
```
okiwi6, line 116: Not your changes, but maybe we can make this a parameter? :)

okiwi6 (commented):
Will continue review tomorrow

crates/vision/src/line_detection.rs
```diff
@@ -38,8 +38,10 @@ pub struct CycleContext {
     gradient_alignment: Parameter<f32, "line_detection.$cycler_instance.gradient_alignment">,
     maximum_distance_to_robot:
         Parameter<f32, "line_detection.$cycler_instance.maximum_distance_to_robot">,
-    maximum_fit_distance_in_pixels:
+    maximum_fit_distance_in_field:
```
okiwi6, line 41: rebasing error?

crates/vision/src/ransac.rs
```diff
@@ -24,34 +24,42 @@ impl Ransac {
...
+                    .filter(|&point| {
+                        line.squared_distance_to_point(*point) <= maximum_score_distance_squared
                     })
-                    .count();
+                    .map(|point| 1.0 - line.distance_to_point(*point) / maximum_score_distance)
+                    .sum();
                 (line, score)
             })
-            .max_by_key(|scored_line| scored_line.1)
+            .max_by(|line_1, line_2| line_1.1.total_cmp(&line_2.1))
```
okiwi6, line 58: since score is not a member of a line, maybe prefer pattern matching here (`(_, score_1), ...`)

okiwi6 (changes requested):

crates/vision/src/field_border_detection.rs
knoellle: Done

crates/vision/src/line_detection.rs
knoellle: Undone.

okiwi6 (approved):
Nice, LGTM

PR #686, opened 2024-01-15
Merge Segments for line detection

knoellle:
## Introduced Changes

What it says. We originally were merging consecutive non-field segments directly in the segment filter but realized that feet detection and ball detection rely on counting segments.
Now the merging happens on the fly when generating seed points for ransac in the line detection node.

This PR is part of a set:
- #683
- #684
- #686
- #687

## Main

image

## This branch

image

Fixes #

Investigate merging segments based on distance in robot coordinates instead of distance in pixels.

## How to Test

Have a look at images with wide (bottom camera), noisy (sprayed for example) lines.
On main, there would be many small ransac seed points all over the place but on this branch they should be a lot closer to the center of the line.


schluis (changes requested):
Very nice.

Since you're editing this file, do you think it's feasable to use the new `CoordinateSystems` here?

etc/parameters/default.json
```diff
@@ -119,7 +120,8 @@
       "maximum_gap_on_line": 30.0,
       "maximum_number_of_lines": 10,
       "maximum_projected_segment_length": 0.3,
-      "minimum_number_of_points_on_line": 4
+      "minimum_number_of_points_on_line": 4,
+      "maximum_merge_gap": 30
```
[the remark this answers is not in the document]
knoellle: Done

crates/vision/src/line_detection.rs
```diff
@@ -234,6 +237,47 @@ fn get_gradient(image: &YCbCr422Image, point: Point2<u16>) -> Vector2<f32> {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut current = self.iterator.next()?;
+
+        while let Some(next) = self.iterator.peek().copied() {
+            if next.start - current.end >= self.maximum_merge_gap {
+                break;
+            }
+
+            let _ = self.iterator.next();
```
schluis, line 271: I think you should be able to remove `let _ =`

crates/vision/src/line_detection.rs
```diff
@@ -331,110 +379,9 @@ fn is_segment_shorter_than(
...
-            &filtered_segments,
-            &image,
-            check_line_segments_projection,
-            maximum_projected_segment_length,
-            check_edge_gradient,
-            gradient_alignment,
-        );
-        assert_eq!(line_points.len(), 32);
-    }
-
```
schluis, line 437: Why did you remove the tests? Because they failed? 😆

schluis (changes requested):

crates/vision/src/line_detection.rs
knoellle: Yes, but I think this is more readable. I explicitly and intentionally discard the return value here.

crates/vision/src/line_detection.rs
schluis: Okay, please elaborate...

knoellle: Since we were previously applying the median in the same direction we were constructing segments, we were actually using a sort of moving median window.
However, since the median direction and segment direction are now orthogonal, every pixel is only inspected once (but more pixels are looked at in total) now.

Yes, we can construct new test data which would make these tests "work" again, but we already have tests without median and the medians are tested separately anyways.
Given that the two are orthogonal now, I see little value in trying to repair these tests.

schluis: Okay, makes sence 👍🏻

schluis (approved):

PR #687, opened 2024-01-15
Ransac lines in ground coordinates

knoellle:
## Introduced Changes

What it says.
On current main we use pixel coordinates for ransac and the associated scoring.
This doesn't really make sense because the inclusion margin for points should depend on physical distance instead of pixel distance.

This PR changes all calculations in line detection to ground coordinates (on the field but relative to the robot).
Note that some debug information has to be transformed back into pixel coordinates.

Fixes #

This PR is part of a set:
- #683
- #684
- #686
- #687


crates/vision/src/line_detection.rs
```diff
@@ -77,7 +77,11 @@ impl LineDetection {
             *context.gradient_alignment,
         );
         if context.lines_in_image.is_subscribed() {
-            image_lines.points = line_points.clone();
+            image_lines.points = line_points
+                .iter()
+                .cloned()
```
okiwi6, line 82: use `copied()` here

okiwi6 (changes requested):
Minor remark otherwise looks great

okiwi6 (approved):
LGTM

PR #689, opened 2024-01-17
Abstract mirroring Joints using the Mirror trait

schmidma:
## Introduced Changes

Add a new `Mirror` trait to abstract mirroring joints.

## ToDo / Known Issues

none

## Ideas for Next Iterations (Not This PR)

none

## How to Test

compile, this is only refactoring

PR #691, opened 2024-01-17
Add LowerBody and UpperBody joints

schmidma:
## Introduced Changes

Add structs for the lower and upper body to better structure joints.

This is based on #689 

## ToDo / Known Issues

none

none

## How to Test

just compile, this is only refactoring

### DIFF
--- a/crates/types/src/joints/body.rs
+++ b/crates/types/src/joints/body.rs
@@ -27,6 +27,17 @@ pub struct BodyJoints<T> {
     pub right_leg: LegJoints<T>,
 }
 
+impl<T> BodyJoints<T> {
+    pub fn from_lower_and_upper(lower: LowerBodyJoints<T>, upper: UpperBodyJoints<T>) -> Self {
+        Self {
+            left_arm: upper.left_arm,
+            right_arm: upper.right_arm,
+            left_leg: lower.left_leg,
+            right_leg: lower.right_leg,
+        }
+    }
+}
+
 impl<T> BodyJoints<T>
 where
     T: Clone,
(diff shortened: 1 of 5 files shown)


crates/types/src/joints/mirror.rs
```diff
@@ -0,0 +1,3 @@
+pub trait Mirror {
+    fn mirrored(self) -> Self;
+}
```
schluis, line 3: What do you think of moving all `impl Mirror for x<y>` to this file?

crates/types/src/joints/body.rs
```diff
@@ -125,3 +136,45 @@ impl Div<f32> for BodyJoints<f32> {
...
 }
+
+#[derive(
+    Clone, Copy, Debug, Default, Deserialize, PartialEq, Eq, Serialize, SerializeHierarchy,
+)]
+#[serialize_hierarchy(bound = "T: SerializeHierarchy + Serialize, for<'de> T: Deserialize<'de>")]
+pub struct LowerBodyJoints<T> {
+    pub left_leg: LegJoints<T>,
+    pub right_leg: LegJoints<T>,
+}
```
schluis, line 147: Naming suggestion: `LegPairJoints` and `ArmPairJoints`

crates/types/src/joints/body.rs
```diff
@@ -27,6 +27,17 @@ pub struct BodyJoints<T> {
     pub right_leg: LegJoints<T>,
 }
 
+impl<T> BodyJoints<T> {
+    pub fn from_lower_and_upper(lower: LowerBodyJoints<T>, upper: UpperBodyJoints<T>) -> Self {
```
schluis, line 31: Naming suggestion: `from_arm_and_leg_pairs`

schluis (changes requested):
nice 👌🏻

crates/types/src/joints/mirror.rs
schmidma: I'd like to keep it at the individual types, as this is the usual way to implement traits for types we own.

crates/types/src/joints/leg.rs
```diff
@@ -146,8 +148,8 @@ impl Div<LegJoints<f32>> for LegJoints<f32> {
     }
 }
 
-impl LegJoints<f32> {
-    pub fn mirrored(self) -> Self {
+impl Mirror for LegJoints<f32> {
```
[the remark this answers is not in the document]
schmidma: huh?! okay... what formatting style do you have in mind? Why should it be there?

crates/types/src/joints/body.rs
schmidma: I also thought about that while implementing the types. My thoughts on that were that it is not clear what part is actually still body or which is already leg/arm. For example the hip-yaw-pitch is not clearly belonging to any leg of the NAO robot.
I personally like the Lower and Upper Body more.

crates/types/src/joints/body.rs
schmidma: see replying comment above

crates/types/src/joints/mirror.rs
schluis: Okay

crates/types/src/joints/body.rs
schluis: My thought was quite simple: it containt two `LegJoints`. `LowerBodyJoints` feels a bit vague. But you decide...

crates/types/src/joints/leg.rs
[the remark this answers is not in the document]
schluis: Because I think the structure is more consistent.
I suggest (generally):
1. `pub struct x {}`
2. `impl x {}`
3. `impl y for x {}`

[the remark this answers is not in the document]
schluis: Also, I've just noticed that there's no blank line in the legs.rs between the impl for mirror and the next impl.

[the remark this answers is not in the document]
schluis: Yep, feel free to do that :)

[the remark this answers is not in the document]
schmidma: Done. Sorted: first bare impls, then trait impls

schluis (approved):
👌🏻

PR #693, opened 2024-01-17
New OS versioN

PatrickGoettsch:
## Introduced Changes

as it says


julianschuler:
We do not increment the patch version of the SDK, since the SDK is compatible with the OS as long as the major and minor version match. This is also the reason the CI fails, since there only exists an SDK with path version 0. I suggest to close this PR in favor of #694.

PR #710, opened 2024-01-31
Framework Time Measurement

oleflb:
## Introduced Changes
Adds instrumental profiling for all framework nodes subscribable at a new `cycle_times` output.
Only works in twix when #709 is merged.

Fixes #

## ToDo / Known Issues

## How to Test

* Check that cycle time outputs are generated and meaningful


okiwi6:
@okiwi6, @knoellle, @h3ndrk discussed about this today, we have not yet found a solution. @h3ndrk is not a fan that timings are exposed via the output interface, since he argues that strictly speaking this is not an output. @knoellle argues that this information is strictly bound to the node execution and therefore would also make sense next to the other outputs.
This needs further discussion.

oleflb:
superseded by #1823

PR #711, opened 2024-02-03
Correct values for center-of-mass computation

schmidma:
## Introduced Changes

Correct signs and values for center of mass computation.

## ToDo / Known Issues

none

none

## How to Test

Subscribe and look at the center of mass via twix and reason whether it is correct :P

### DIFF
--- a/crates/types/src/robot_dimensions.rs
+++ b/crates/types/src/robot_dimensions.rs
@@ -4,7 +4,7 @@ use nalgebra::{vector, Vector3};
 pub struct RobotDimensions {}
 
 impl RobotDimensions {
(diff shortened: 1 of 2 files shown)


MaikRe:
I haven't fully gone in detail on my center of mass data but something definitely looked off with a center of mass clearly way outside of the torso. So just for fun I took the old center of masses and changed the webots proto to match those, i.e. removing a 0 after the comma. Clearly it's not supposed to look like that.

https://github.com/HULKs/hulk/assets/87246463/64343d73-db23-41d5-a3e3-c170da39de33

MaikRe (approved):
LGTM

PR #715, opened 2024-02-07
Coordinate Systems Rollout

knoellle:
## Introduced Changes

This PR takes the concepts introduced in #611 and applies them to the whole repository.
It was previously suggested to do this whenever we touch other pieces of code, however, I think due to the ripple effects of wrapping things like `robot_to_field`, this isn't a good idea and I'd rather do it all in one richtig draufhauen.
Since this resulted in a large number of changed lines, I will outline the most important changes here.

### Coordinate Systems

Most of the codebase now uses nalgebra types like Point, Vector, or Isometry only wrapped in `Framed` and `Transform` from #611.
I have defined all of the coordinate systems that are used in our codebase in `types::coordinate_systems`, except for those used in tests which are always defined locally within the test module.

According to @schmidma, his intention was to wrap all types that exist within a coordinate system in a `Framed` type, i.e. `Framed<Ground, Circle>`.
However, I think it makes more sense to wrap only the foreign (nalgebra) types and instead make our types generic over the Frame (where applicable) which then contain nalgebra types wrapped with the frame generic.

There are mainly two reasons for this, which I will demonstrate using the example of the `Circle` type, which previously looked like this:
```rust
pub struct Circle {
    pub center: Point2<f32>,
    pub radius: f32,
}
 ```

Firstly, when accessing members of the wrapped type `Framed<Circle>`, you would have the `.inner` indirection:
`my_circle.inner.center`, which is more verbose **and**  loses the frame, as the `center` member is not itself wrapped.
Secondly, you cannot call the inner type's member functions without the `.inner` indirection or reimplementing them by `impl Framed<Circle> { ... }`.
This would be fine for type we only sometimes want to use in their wrapped form, but this is mostly for types in our `geometry` crate. Since these should always exist within a specified coordinate system anyways, there would be no point to have separate implementations for the wrapped and non-wrapped types.

If we instead have a Circle containing a `Framed<Point2>` as its `center` member, we can simply write `my_circle.center` and get a wrapped nalgebra type.
The new circle type thus looks like this:
```rust
pub struct Circle<Frame> {
    pub center: Framed<Frame, Point2<f32>>,
    pub radius: f32,
}
```

However, integrating coordinate systems with our geometry types also makes them impossible to use without specifying a coordinate system.
Depending on how seriously we want to take this idea, this is either a blessing or a curse.

I also implemented a bunch of delegating functions specifically for specific wrapped nalgebra types, so instead of `my_vector.inner.normalize().framed()` (which also loses track of the coordinate system) you can simply write `my_vector.normalize()` which returns a normalized vector tagged with the same coordinate system.
These are implemented in separate modules within the `coordinate_systems` crate.

There is some room for interpretation with the coordinate systems, especially with regards to poses and other frames of reference that change over time.
Before this PR, I never realized that we actually have two different kinds of poses, one relative to the field and one relative to the current robot pose.
The former is simply a `Transform<Ground, Field>`, but the latter would technically be a `Transform<NewGround, OldGround>`.
However, I have used `Transform<Ground, Ground>` in this PR, because even though it reduces safety slightly, it makes dealing with different poses much easier.
Even if we did use different types here, it wouldn't provide perfect safety, as positions from different KickPose coordinate systems still shouldn't be added together, even if they are both `Framed<KickPose, Point2>`s.

### Renaming `robot_to_field` -> `ground_to_field`

We have had this issue in our code for quite a while, where we accidentally called the "Ground" coordinate system "Robot".
This PR renames all occurrences of `robot_to_field` to `ground_to_field` alongside renaming variables `..._in_robot` to `..._in_ground`.
I may have missed some occurrences of the latter case.
Also, note that there still is a "Robot" coordinate frame, its just not the same as what we used to call "Robot" and should only appear in parts of the code related to kinematics.

### Limb projector

The limb projector previously used the same (unmirrored) polygons for left and right limbs, which only happens to work because the current parameters happen to be symmetric along the X-Z plane, which at least for the foot isn't accurate.
This PR does not change the polygon parameters but does mirror them for the right side of the body along with using the proper coordinate systems for each.
I also added an image overlay in twix which renders the polygons projected into the image.

### approx_derive

This PR changes our implementation for deriving the traits from the `approx` crate to work for structs without any members.
That change was necessary to derive these traits on the marker types used for coordinate system.

Fixes rabsolute bugs

## ToDo / Known Issues

- Add doc comments to the coordinate systems explaining what they are and how they relate to each other.

- Most of the changes in this PR stem from adding `.framed()` to tag  nalgebra types with a coordinate system.
But since this is a comprehensive rollout, most nalgebra variables/members are already wrapped, so most occurences of `.framed()` follow directly after a `point![...]` or `vector![...]` macro.
This could be alleviated by creating our own macros which work like the nalgebra ones but produce framed types instead, however, I have yet to come up with good names for these.
- I have implemented a bunch of accessor/delegate functions, but there are still more that can be added to prevent `a.inner.b.framed()` scenarios.

## How to Test

Since most of the changes essentially just add compile time safety, the risk of breakage should be fairly minimal and compilability checking is already done by the CI.
However, I did make a mistake during the great wrappening, see [this commit](https://github.com/HULKs/hulk/commit/33c6c20d40394843c8a93fa9caf94ca6f9178a8f) of which there are possibly more.
I already tested this on a robot, but I think its a good idea to have others test this too.


schmidma (commented):
Some intermediate comments

crates/control/src/ball_filter.rs
```diff
@@ -258,21 +263,21 @@ impl BallFilter {
...
         hypothesis: &mut Hypothesis,
-        detected_position: Point2<f32>,
+        detected_position: Point2<Ground>,
         detection_time: SystemTime,
         configuration: &BallFilterParameters,
     ) {
         hypothesis.moving_state.update(
             Matrix2x4::identity(),
-            detected_position.coords,
+            detected_position.inner.coords,
```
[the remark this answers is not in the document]
knoellle: I thought we agreed not to frame the multivariate distributions. Because of that, we need a raw nalgebra vector here, either through `.coords().inner` or through `.inner.coords`.

crates/control/src/behavior/support.rs
```diff
@@ -44,17 +45,17 @@ fn support_pose(
...
+) -> Option<Pose<Ground>> {
+    let ground_to_field = world_state.robot.ground_to_field?;
     let ball = world_state
         .rule_ball
         .or(world_state.ball)
-        .unwrap_or_else(|| BallState::new_at_center(robot_to_field));
+        .unwrap_or_else(|| BallState::new_at_center(ground_to_field));
     let side = field_side.unwrap_or_else(|| ball.field_side.opposite());
-    let offset_vector = UnitComplex::new(match side {
+    let offset_vector = UnitComplex::<Field, Field>::new(match side {
```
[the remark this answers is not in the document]
knoellle: I don't think Orientation fits here. We create a rotation which is then used to transform a point. Orientations aren't supposed to do that, that's the whole point.

We could define a new coordinate system here which the rotation then transforms to field coordinates, but a good name for that remains elusive and I'm not convinced we gain much by that.

crates/control/src/ball_filter.rs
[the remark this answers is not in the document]
schmidma: sorry, mixed that up. resolving...

crates/control/src/behavior/support.rs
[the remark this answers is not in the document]
schmidma: true, i agree

crates/control/src/localization.rs
```diff
@@ -895,67 +912,71 @@ fn get_translation_and_rotation_measurement(
         _ => field_mark_correspondence.measured_line_in_field,
     };
     let measured_line_in_field_vector = measured_line_in_field.1 - measured_line_in_field.0;
-    let signed_distance_to_line = measured_line_in_field
-        .signed_distance_to_point(Point2::from(robot_to_field.translation.vector));
+    let signed_distance_to_line =
+        measured_line_in_field.signed_distance_to_point(ground_to_field * Point2::origin());
```
schmidma, line 916: this is your current `.origin()` is it?

crates/control/src/localization.rs
```diff
@@ -895,67 +912,71 @@ fn get_translation_and_rotation_measurement(
...
     match field_mark_line_direction {
         Direction::PositiveX => {
-            vector![
-                field_mark_line.0.y + signed_distance_to_line,
-                (-measured_line_in_field_vector.y).atan2(measured_line_in_field_vector.x)
-                    + robot_to_field.rotation.angle()
+            nalgebra::vector![
+                field_mark_line.0.y() + signed_distance_to_line,
+                (-measured_line_in_field_vector.y()).atan2(measured_line_in_field_vector.x())
```
schmidma, line 921: you said something about replacing atan?

crates/control/src/motion/look_at.rs
```diff
@@ -89,11 +91,11 @@ impl LookAt {
         }
 
         let (target, camera) = match head_motion {
-            HeadMotion::LookAt { target, camera } => (*target, *camera),
+            HeadMotion::LookAt { target, camera } => (target, camera),
             HeadMotion::LookLeftAndRightOf { target } => {
                 let left_right_shift = vector![
                     0.0,
-                    f32::tan(*context.glance_angle) * distance(target, &Point2::origin())
+                    f32::tan(*context.glance_angle) * distance(target, Point2::origin())
```
schmidma, line 98: same as `.norm()`?

crates/control/src/role_assignment.rs
```diff
@@ -236,14 +240,15 @@ impl RoleAssignment {
         } else {
             for spl_message in spl_messages {
                 self.last_received_spl_striker_message = Some(cycle_start_time);
-                let sender_position =
-                    (robot_to_field.inverse() * spl_message.robot_to_field) * Point2::origin();
+                let sender_position = ground_to_field.inverse()
+                    * spl_message.ground_to_field.framed_transform()
+                    * Framed::<Ground, _>::origin();
```
schmidma, line 245: Point?

crates/spl_network_messages/src/lib.rs
```diff
@@ -23,7 +23,7 @@ pub use visual_referee_message::{VisualRefereeDecision, VisualRefereeMessage};
 pub struct HulkMessage {
     pub player_number: PlayerNumber,
     pub fallen: bool,
-    pub robot_to_field: Isometry2<f32>,
+    pub ground_to_field: Isometry2<f32>,
     pub ball_position: Option<BallPosition>,
```
schmidma, line 27: These should be framed...

crates/control/src/role_assignment.rs
```diff
@@ -608,50 +613,50 @@ fn decide_if_claiming_striker_or_other_role(
...
+    ball: Option<&BallPosition<Ground>>,
+    ground_to_field: Isometry2<Ground, Field>,
     cycle_start_time: SystemTime,
-) -> Option<BallPosition> {
+) -> Option<BallPosition<Field>> {
     ball.as_ref().map(|ball| BallPosition {
-        position: (current_pose * ball.position),
-        velocity: Vector2::zeros(),
+        position: (ground_to_field * ball.position),
+        velocity: Framed::zeros(),
```
schmidma, line 659: Vector

crates/control/src/rule_obstacle_composer.rs
```diff
@@ -75,7 +76,7 @@ impl RuleObstacleComposer {
                 _,
             ) => {
                 let center_circle_obstacle = RuleObstacle::Circle(Circle::new(
-                    Point2::origin(),
+                    Framed::origin(),
```
schmidma, line 79: Point ... I think you should search for all occurrences of `Framed` and try to replace them if appropriate

crates/coordinate_systems/src/framed.rs
```diff
@@ -49,17 +70,182 @@ where
...
+}
+
+impl<Frame, Inner> Eq for Framed<Frame, Inner> where Inner: Eq {}
+
+impl<Frame, Inner> Hash for Framed<Frame, Inner>
+where
+    Inner: Hash,
+{
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // TODO: do we want to hash the state? does it even do anything?
```
schmidma, line 238: todo

crates/coordinate_systems/src/transform_nalgebra.rs
```diff
@@ -0,0 +1,85 @@
...
+{
+    pub fn identity() -> Self {
+        Self::wrap(nalgebra::Isometry::identity())
+    }
+
+    pub fn inverse(&self) -> Transform<To, From, Isometry<Scalar, Rotation, DIMENSION>> {
+        Transform::<To, From, _>::wrap(self.inner.inverse())
+    }
+
+    // TODO: naming kloppen
```
schmidma, line 22: todo

crates/geometry/src/arc.rs
```diff
@@ -46,23 +53,23 @@ impl RelativeEq for Arc {
...
 
-    pub fn length(&self, orientation: Orientation) -> f32 {
+    pub fn length(&self, orientation: Direction) -> f32 {
         let vector_start = self.start - self.circle.center;
         let vector_end = self.end - self.circle.center;
 
-        let angle_x_axis_to_start = vector_start.y.atan2(vector_start.x);
-        let mut angle = vector_end.y.atan2(vector_end.x) - angle_x_axis_to_start;
+        let angle_x_axis_to_start = vector_start.y().atan2(vector_start.x());
+        let mut angle = vector_end.y().atan2(vector_end.x()) - angle_x_axis_to_start;
```
schmidma, line 66: you said something about replacing atan?

crates/geometry/src/line_segment.rs
```diff
@@ -1,15 +1,24 @@
...
-use crate::{arc::Arc, orientation::Orientation};
+use coordinate_systems::{vector, Point2, Vector2};
 
-#[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct LineSegment(pub Point2<f32>, pub Point2<f32>);
+use crate::{arc::Arc, direction::Direction};
 
-impl AbsDiffEq for LineSegment {
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+#[serde(bound = "")]
```
schmidma, line 11: necessary?

crates/geometry/src/line_segment.rs
```diff
@@ -1,15 +1,24 @@
...
+pub struct LineSegment<Frame>(pub Point2<Frame>, pub Point2<Frame>);
+
+// Manual implementation required because the derived version imposes Frame to be PartialEq
```
schmidma, line 14: but it is ... Maybe you missed it when adding it?

crates/geometry/src/two_line_segments.rs
```diff
@@ -5,10 +5,18 @@ use serialize_hierarchy::{Error, SerializeHierarchy};
 
 use crate::line_segment::LineSegment;
 
-#[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct TwoLineSegments(pub LineSegment, pub LineSegment);
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+#[serde(bound = "")]
+pub struct TwoLineSegments<Frame>(pub LineSegment<Frame>, pub LineSegment<Frame>);
+
+// Manual implementation required because the derived version imposes Frame to be PartialEq
```
schmidma, line 12: same here

crates/geometry/src/two_line_segments.rs
```diff
@@ -5,10 +5,18 @@ use serialize_hierarchy::{Error, SerializeHierarchy};
 
 use crate::line_segment::LineSegment;
 
-#[derive(Clone, Copy, Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct TwoLineSegments(pub LineSegment, pub LineSegment);
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+#[serde(bound = "")]
```
schmidma, line 9: and here...

crates/projection/src/lib.rs
```diff
@@ -14,137 +18,142 @@ pub enum Error {
...
+        resolution: Point2<Pixel, u32>,
     ) -> Result<f32, Error> {
         let robot_coordinates =
             self.pixel_to_ground_with_z(pixel_coordinates, radius_in_robot_coordinates)?;
-        let camera_coordinates =
-            self.ground_to_camera * point![robot_coordinates.x, robot_coordinates.y, 0.0];
-        let distance = camera_coordinates.coords.norm();
+        let camera_coordinates = self.ground_to_camera
+            * point![robot_coordinates.inner.x, robot_coordinates.inner.y, 0.0];
+        let distance = camera_coordinates.inner.coords.norm();
```
schmidma, line 152: remove .inner?

crates/projection/tests/projection.rs
```diff
@@ -1,29 +1,30 @@
...
+    image_size: Point2<Pixel, f32>,
 ) -> CameraMatrix {
     CameraMatrix::from_normalized_focal_and_center(
         focal_length,
         optical_center,
         image_size,
-        Isometry3::identity(),
-        Isometry3::identity(),
-        Isometry3::identity(),
+        Isometry3::identity().framed_transform(),
```
schmidma, line 16: use the coordinate_systems Isometry3?

crates/types/src/ball.rs
```diff
@@ -1,23 +1,29 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, SerializeHierarchy, AbsDiffEq, RelativeEq)]
 #[abs_diff_eq(epsilon = "f32")]
 pub struct Ball {
-    pub position: Point2<f32>,
-    pub image_location: Circle,
+    pub position: Point2<Ground>,
+    pub image_location: Circle<Pixel>,
+}
+
+impl PartialEq for Ball {
```
schmidma, line 25: derive?

crates/types/src/ball_filter.rs
```diff
@@ -34,11 +36,14 @@ impl Hypothesis {
...
-    pub fn selected_ball_position(&self, configuration: &BallFilterParameters) -> BallPosition {
+    pub fn selected_ball_position(
+        &self,
+        configuration: &BallFilterParameters,
+    ) -> BallPosition<Ground> {
         let selected_state = self.selected_state(configuration);
 
         BallPosition {
-            position: Point2::from(selected_state.mean.xy()),
+            position: Vector::wrap(selected_state.mean.xy()).as_point(),
```
schmidma, line 46: Is there no `Point::from` or `wrap`?

crates/types/src/camera_matrix.rs
```diff
@@ -34,14 +40,16 @@ impl CameraMatrices {
...
-    pub focal_length: Vector2<f32>,
-    pub optical_center: Point2<f32>,
-    pub field_of_view: Vector2<f32>,
+    pub camera_to_head: Isometry3<Camera, Head>,
+    pub camera_to_ground: Isometry3<Camera, Ground>,
+    pub ground_to_camera: Isometry3<Ground, Camera>,
+    pub camera_to_robot: Isometry3<Camera, Robot>,
+    pub robot_to_camera: Isometry3<Robot, Camera>,
+    // TODO: I don't know enough about this to say which
+    // (if any) coordinate system these belong to
```
schmidma, line 49: todo

schmidma: they definitely do, maybe we take this opportunity and discuss with ole and his latest insights?

crates/types/src/interpolated.rs
```diff
@@ -18,10 +21,10 @@ impl Interpolated {
     const ARGUMENT_FIRST_HALF_OPPONENT_HALF_TOWARDS_OWN_GOAL: Point2<f32> = point![3.0, 0.0];
     const ARGUMENT_FIRST_HALF_OPPONENT_HALF_AWAY_OWN_GOAL: Point2<f32> = point![3.0, PI];
 
-    pub fn evaluate_at(&self, robot_to_field: Isometry2<f32>) -> f32 {
+    pub fn evaluate_at(&self, ground_to_field: Transform<Ground, Field, Isometry2<f32>>) -> f32 {
```
schmidma, line 24: Use new typedef

crates/types/src/line_data.rs
```diff
@@ -1,15 +1,18 @@
...
+use crate::{
+    coordinate_systems::{Ground, Pixel},
+    line::Line2,
+};
 
 #[derive(Clone, Default, Debug, Serialize, Deserialize, SerializeHierarchy)]
 pub struct LineData {
-    pub lines_in_robot: Vec<Line2>,
-    pub used_vertical_filtered_segments: HashSet<Point2<u16>>,
+    pub lines_in_ground: Vec<Line2<Ground>>,
```
schmidma, line 14:
```suggestion
    pub lines: Vec<Line2<Ground>>,
```

crates/types/src/perspective_grid_candidates.rs
```diff
@@ -69,23 +72,23 @@ mod tests {
             PerspectiveGridCandidates {
                 candidates: vec![
                     Circle {
-                        center: point![0.0, 0.0],
+                        center: point![0.0, 0.0].framed(),
```
schmidma, line 75: use the wrapped macro

crates/types/src/robot_dimensions.rs
```diff
@@ -3,6 +3,7 @@ use nalgebra::{vector, Vector3};
 #[derive(Debug)]
 pub struct RobotDimensions {}
 
+// TODO: consider wrapping in coordinate systems
```
schmidma, line 6: We definitely should do that!

crates/types/src/robot_kinematics.rs
```diff
@@ -1,40 +1,48 @@
...
+    RightForearm, RightHip, RightPelvis, RightShoulder, RightSole, RightThigh, RightTibia,
+    RightUpperArm, RightWrist, Robot, Torso,
+};
+
 #[derive(Debug, Clone, Default, SerializeHierarchy, Serialize, Deserialize)]
 pub struct RobotKinematics {
     // head
-    pub neck_to_robot: Isometry3<f32>,
-    pub head_to_robot: Isometry3<f32>,
+    pub neck_to_robot: Transform<Neck, Robot, Isometry3<f32>>,
```
schmidma, line 16: Use the new syntax with wrapped types

crates/types/src/robot_masses.rs
```diff
@@ -1,141 +1,145 @@
...
+    RightWrist, Torso,
+};
+use coordinate_systems::{point, Point3};
 
 #[derive(Debug)]
-pub struct RobotMass {
+pub struct RobotMass<Frame> {
     pub mass: f32,
-    pub center: Point3<f32>,
+    pub center: Point3<Frame, f32>,
```
schmidma, line 12: omit the f32

crates/vision/src/field_border_detection.rs
```diff
@@ -137,25 +139,24 @@ fn find_border_lines(
...
 
-fn find_centre_of_group(group: &[Point2<f32>]) -> Point2<f32> {
-    Point2::<f32> {
-        coords: group
-            .iter()
-            .map(|point| point.coords)
-            .sum::<Vector2<f32>>()
-            .unscale(group.len() as f32),
-    }
+fn find_centre_of_group(group: &[Point2<Pixel>]) -> Point2<Pixel> {
```
schmidma, line 149:
```suggestion
fn find_center_of_group(group: &[Point2<Pixel>]) -> Point2<Pixel> {
```

crates/vision/src/field_border_detection.rs
```diff
@@ -137,25 +139,24 @@ fn find_border_lines(
...
+    group
+        .iter()
+        .map(|point| point.coords())
+        .sum::<Vector2<Pixel>>()
```
schmidma, line 153:
```suggestion
        .sum::<Vector2<_>>()
```

crates/vision/src/field_color_detection.rs
```diff
@@ -1,13 +1,18 @@
...
+use types::{
+    coordinate_systems::{Field, Ground},
+    field_color::FieldColor,
+    interpolated::Interpolated,
+};
 
 #[derive(Deserialize, Serialize)]
 pub struct FieldColorDetection {
-    robot_to_field_of_home_after_coin_toss_before_second_half: Isometry2<f32>,
+    ground_to_field_of_home_after_coin_toss_before_second_half: Isometry2<Ground, Field>,
```
schmidma, line 15: :exploding_head:

schmidma (commented):
puh... done :steam_locomotive:

crates/control/src/behavior/support.rs
```diff
@@ -44,17 +45,17 @@ fn support_pose(
...
         Side::Left => -FRAC_PI_4,
         Side::Right => FRAC_PI_4,
-    }) * -(Vector2::x() * distance_to_ball);
+    }) * -(Vector2::x_axis() * distance_to_ball);
```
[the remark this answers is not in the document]
knoellle: Yes, it does work in this case.
I prefer this way too, done.

crates/control/src/behavior/walk_to_penalty_kick.rs
```diff
@@ -14,15 +14,15 @@ pub fn execute(
     path_obstacles_output: &mut AdditionalOutput<Vec<PathObstacle>>,
     field_dimensions: &FieldDimensions,
 ) -> Option<MotionCommand> {
-    let robot_to_field = world_state.robot.robot_to_field?;
-    let kick_off_pose = Isometry2::translation(
+    let ground_to_field = world_state.robot.ground_to_field?;
+    let kick_off_pose = Pose::from_position(point![
```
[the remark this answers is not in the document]
knoellle: I removed the from_position function.

crates/control/src/behavior/walk_to_pose.rs
```diff
@@ -81,20 +82,20 @@ impl<'cycle> WalkPathPlanner<'cycle> {
...
             * point![
-                target_in_field.x.clamp(-x_max, x_max),
-                target_in_field.y.clamp(-y_max, y_max)
+                target_in_field.x().clamp(-x_max, x_max),
+                target_in_field.y().clamp(-y_max, y_max)
             ];
 
         let path = planner
-            .plan(Point2::origin(), clamped_target_in_robot)
+            .plan(Framed::origin(), clamped_target_in_robot)
```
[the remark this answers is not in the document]
knoellle: Replaced all of those

crates/control/src/behavior/walk_to_pose.rs
```diff
@@ -81,20 +82,20 @@ impl<'cycle> WalkPathPlanner<'cycle> {
...
             .unwrap();
         path_obstacles_output.fill_if_subscribed(|| planner.obstacles.clone());
-        path.unwrap_or_else(|| direct_path(Point2::origin(), Point2::origin()))
+        path.unwrap_or_else(|| direct_path(Framed::origin(), Framed::origin()))
```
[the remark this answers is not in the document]
knoellle: Replaced all of those

crates/coordinate_systems/src/framed_nalgebra.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn from_cos_sin_unchecked(cos: f32, sin: f32) -> Self {
+        Self::wrap(nalgebra::UnitComplex::from_cos_sin_unchecked(cos, sin))
+    }
+
+    pub fn from_vector(direction: Vector2<Frame>) -> Self {
+        Orientation::wrap(nalgebra::UnitComplex::rotation_between(
+            &nalgebra::Vector2::x_axis(),
+            &direction.inner,
+        ))
+    }
```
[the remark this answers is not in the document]
knoellle: This implication is what defines this whole type, its not just a property of this constructor.
This is how all our poses have worked since forever: If the rotational part is an identity mapping, the forward vector aligns with the X axis.

crates/control/src/ground_provider.rs
```diff
@@ -25,8 +27,8 @@ pub struct CycleContext {
 #[context]
 #[derive(Default)]
 pub struct MainOutputs {
-    pub robot_to_ground: MainOutput<Option<Isometry3<f32>>>,
-    pub ground_to_robot: MainOutput<Option<Isometry3<f32>>>,
+    pub robot_to_ground: MainOutput<Option<Transform<Robot, Ground, Isometry3<f32>>>>,
+    pub ground_to_robot: MainOutput<Option<Transform<Ground, Robot, Isometry3<f32>>>>,
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/ground_provider.rs
```diff
@@ -40,25 +42,31 @@ impl GroundProvider {
         let imu_pitch = imu_roll_pitch.y;
 
         let left_sole_to_robot = context.robot_kinematics.left_sole_to_robot;
+        // TODO: coordinate systems
```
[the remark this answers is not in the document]
knoellle: Done, made some rather large changes in this file, please have a look at it again.

crates/control/src/localization.rs
```diff
@@ -84,23 +85,23 @@ pub struct CycleContext {
     player_number: Parameter<PlayerNumber, "player_number">,
     score_per_good_match: Parameter<f32, "localization.score_per_good_match">,
     use_line_measurements: Parameter<bool, "localization.use_line_measurements">,
-    injected_robot_to_field_of_home_after_coin_toss_before_second_half: Parameter<
-        Option<Isometry2<f32>>,
-        "injected_robot_to_field_of_home_after_coin_toss_before_second_half?",
+    injected_ground_to_field_of_home_after_coin_toss_before_second_half: Parameter<
+        Option<Isometry2<Ground, Field>>,
+        "injected_ground_to_field_of_home_after_coin_toss_before_second_half?",
     >,
```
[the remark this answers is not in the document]
knoellle: Agreed, but not in this PR.

crates/control/src/kick_selector.rs
```diff
@@ -421,44 +427,49 @@ fn is_inside_any_obstacle(
     })
 }
 
-fn mirror_kick_offset(kick_offset: Vector2<f32>) -> Vector2<f32> {
-    vector![kick_offset.x, -kick_offset.y]
+fn mirror_kick_offset(kick_offset: nalgebra::Vector2<f32>) -> nalgebra::Vector2<f32> {
+    nalgebra::vector![kick_offset.x, -kick_offset.y]
 }
 
+// TODO: not sure where to draw the line here between coordinate-safe and pure nalgebra operations
```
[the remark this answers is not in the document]
knoellle: Do you think the current implementation is fine?

crates/spl_network_messages/src/lib.rs
knoellle: You mean something like this?

```suggestion
    pub pose: Pose<field>,
    pub ball_position: Option<BallPosition>,
```

I thought about that, but this is incompatible with our current crate layout.
We currently have the coordinate systems defined in `types`, which depends on the `spl_network_messages` crate, so we cannot use `Field` here.

crates/control/src/localization.rs
```diff
@@ -147,11 +148,11 @@ impl Localization {
                     kicking_team: Team::Hulks,
                 }),
             ) => {
-                let penalty_shoot_out_striker_pose = Isometry2::translation(
+                let penalty_shoot_out_striker_pose = Pose::from_position(point![
```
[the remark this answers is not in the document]
knoellle: Are you sure about this?
I think having variable names like `initial` or `penalty_shoot_out` hurts the understandability of the code.
I do agree that we shouldn't just repeat type names in variable names, but I don't think that applies here.

crates/control/src/motion/look_at.rs
knoellle: It is, yes. Do you find `target.coords().norm()` more readable than `distance(target, Point2::origin())`?

crates/control/src/localization.rs
```diff
@@ -275,17 +276,18 @@ impl Localization {
                     scored_state.score *= *context.hypothesis_prediction_score_reduction_factor;
                 }
                 if *context.use_line_measurements {
-                    let robot_to_field = scored_state.state.as_isometry();
+                    let ground_to_field: coordinate_systems::Isometry2<Ground, Field> =
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/localization.rs
knoellle: Done

crates/control/src/kick_selector.rs
```diff
@@ -190,38 +192,38 @@ fn generate_decisions_for_instant_kicks(
...
 
-            let is_inside_field = field_dimensions.is_inside_field(robot_to_field * target);
+            let is_inside_field = field_dimensions.is_inside_field(ground_to_field * target);
             let ball_to_target = LineSegment(ball_position, target);
             let is_intersecting_with_an_obstacle = obstacle_circles
                 .iter()
                 .any(|circle| circle.intersects_line_segment(&ball_to_target));
             let opponent_goal_center =
-                robot_to_field.inverse() * point![field_dimensions.length / 2.0, 0.0];
+                ground_to_field.inverse() * point![field_dimensions.length / 2.0, 0.0];
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/kick_selector.rs
```diff
@@ -190,38 +192,38 @@ fn generate_decisions_for_instant_kicks(
         .filter_map(|(&kicking_side, &variant)| {
             let kick_info = &in_walk_kicks[variant];
             let shot_angle = match kicking_side {
-                Side::Left => UnitComplex::new(kick_info.shot_angle),
-                Side::Right => UnitComplex::new(kick_info.shot_angle).inverse(),
+                Side::Left => UnitComplex::<Ground, Ground>::new(kick_info.shot_angle),
+                Side::Right => UnitComplex::<Ground, Ground>::new(kick_info.shot_angle).inverse(),
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/localization.rs
knoellle: Yes, when passing the result of the atan2 directly to a UnitComplex constructor or something like that.
However, in this area of the code, atan2 seems to be used in non-standard ways, swapping the arguments or inverting axes.
I'm not convinced replacing atan2 here makes things better necessarily.

crates/control/src/role_assignment.rs
knoellle: Done

crates/control/src/role_assignment.rs
knoellle: Done

crates/control/src/rule_obstacle_composer.rs
knoellle: Done, I checked all references to the `origin` and `zeros` constructors.

crates/coordinate_systems/src/transform_nalgebra.rs
knoellle: Do you have a suggestion?

crates/geometry/src/arc.rs
knoellle: There are two ways to calculate angles between vectors:
`Vector::angle` always returns the smaller angle between the two vectors and the angle is unsigned.
`UnitComplex::rotation_between` returns a rotation that can transform one vector to the other and is effectively just a wrapper around atan2.
I don't particularly like the idea of doing `UnitComplex::rotation_between().angle()`, but perhaps we can add a yet unnamed function to our vector type which just returns the signed angle.

What do you think?

crates/vision/src/field_color_detection.rs
knoellle: I'm all for removing the interpolation, but if we do want to add something like this again in the future, I think the correct way would be to introduce another coordinate system (`World` maybe) and then have a `field_to_world` transform which is either an identity or a 180 degree rotation.
Then we can do the interpolation in "world"-space.

crates/control/src/ground_provider.rs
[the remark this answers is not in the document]
schmidma: I read again, let us discuss as that in person ;)

crates/control/src/kick_selector.rs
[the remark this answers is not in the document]
schmidma: No, i think we can move that to the safe world.

crates/control/src/motion/look_at.rs
schmidma: lets keep it `distance`

crates/spl_network_messages/src/lib.rs
schmidma: yes, then we may move the coordinate systems to a separate crate ;)

crates/coordinate_systems/src/transform_nalgebra.rs
schmidma: `ground_to_field.as_pose().position()`? maybe?

crates/calibration/src/lines.rs
```diff
@@ -1,19 +1,20 @@
...
+pub struct Lines<Frame> {
+    pub border_line: Line2<Frame>,
+    pub goal_box_line: Line2<Frame>,
+    pub connecting_line: Line2<Frame>,
 }
 
-impl Lines {
-    pub fn to_projected(&self, matrix: &CameraMatrix) -> Result<Self, LinesError> {
+impl Lines<Pixel> {
+    pub fn to_projected(&self, matrix: &CameraMatrix) -> Result<Lines<Ground>, LinesError> {
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/behavior/defend.rs
```diff
@@ -218,52 +220,52 @@ fn defend_goal_pose(
...
         distance_to_target,
     );
-    Some(robot_to_field.inverse() * defend_pose)
+    Some(ground_to_field.inverse() * defend_pose)
 }
 
 pub fn block_on_circle(
-    ball_position: Point2<f32>,
-    target: Point2<f32>,
+    ball_position: Point2<Field, f32>,
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/behavior/dribble.rs
```diff
@@ -55,9 +55,9 @@ pub fn execute(
     );
     let orientation_mode = match hybrid_orientation_mode {
         types::motion_command::OrientationMode::AlignWithPath
-            if ball_position.coords.norm() > 0.0 =>
+            if ball_position.coords().norm() > 0.0 =>
         {
-            OrientationMode::Override(Point2::origin().look_at(&ball_position))
+            OrientationMode::Override(Framed::origin().look_at(&ball_position))
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/behavior/search.rs
```diff
@@ -59,19 +60,19 @@ pub fn execute(
...
+    let ground_to_field = world_state.robot.ground_to_field?;
     let search_role = assign_search_role(world_state);
     let search_position = search_role
-        .map(|role| role.to_position(robot_to_field, field_dimensions))
+        .map(|role| role.to_position(ground_to_field, field_dimensions))
         .unwrap_or(point![0.0, 0.0]);
     let head = HeadMotion::SearchForLostBall;
     if let Some(SearchRole::Goal) = search_role {
-        let goal_pose = robot_to_field.inverse() * Isometry2::from(search_position.coords);
+        let goal_pose = Pose::from_position(search_position);
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/behavior/stand.rs
```diff
@@ -32,9 +32,9 @@ pub fn execute(
                     let penalty_spot_x =
                         field_dimensions.length / 2.0 - field_dimensions.penalty_marker_distance;
                     let penalty_spot_location = point![side_factor * penalty_spot_x, 0.0];
-                    robot_to_field.inverse() * penalty_spot_location
+                    ground_to_field.inverse() * penalty_spot_location
                 }
-                _ => robot_to_field.inverse() * Point2::origin(),
+                _ => ground_to_field.inverse().origin(),
```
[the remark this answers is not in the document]
knoellle: Done

crates/coordinate_systems/src/transform_nalgebra.rs
knoellle: Done

crates/geometry/src/line_segment.rs
knoellle: Right, this was before I gave up and just derived every necessary trait for all of our frames.
I had hoped that would not be necessary, but tag type parameters that don't actually appear in the struct hierarchy are sadly not supported well by derive macros.

crates/spl_network_messages/src/lib.rs
knoellle: ... which opens a new naming problem. 😅 
We already have `crates/coordinate_systems`...

crates/coordinate_systems/src/framed.rs
knoellle: Removed.

crates/types/src/robot_masses.rs
knoellle: Removed

crates/geometry/src/line_segment.rs
knoellle: Replaced most manual trait implementations with derives on all Line types

crates/projection/tests/projection.rs
knoellle: Done

crates/types/src/interpolated.rs
knoellle: Done, also cleaned up the tests in this file w.r.t. using nalgebra types.

crates/types/src/ball_filter.rs
knoellle: Done

crates/types/src/ball.rs
knoellle: Done

crates/geometry/src/line_segment.rs
knoellle: Removed

crates/geometry/src/two_line_segments.rs
knoellle: Removed

crates/geometry/src/two_line_segments.rs
knoellle: Removed

crates/vision/src/field_border_detection.rs
knoellle: Done

crates/vision/src/field_border_detection.rs
knoellle: Done

crates/types/src/robot_kinematics.rs
knoellle: Done

crates/types/src/robot_dimensions.rs
knoellle: Agreed, but last time I tried this I ran into const issues.

crates/types/src/perspective_grid_candidates.rs
knoellle: Done

crates/types/src/line_data.rs
knoellle: Done

crates/control/src/ground_provider.rs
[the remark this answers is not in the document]
knoellle: Done

crates/types/src/camera_matrix.rs
knoellle: Removed comment, lets do this later

crates/types/src/robot_dimensions.rs
knoellle: Done

crates/spl_network_messages/src/lib.rs
knoellle: Done

schmidma (approved):
Was soll schon schiefgehen

PR #666, opened 2024-01-11
Make game controller state consistent

ichbinJosi:
## Introduced Changes

Implements a consistently used `FilteredGameStateController` to enable a single layer abstraction architecture.

## How to Test

Check if everything still works as expected regarding anything concerned with `GameStates` 🤡

PR #722, opened 2024-02-07
Fix setting internal state in game controller state filter

knoellle:
## Introduced Changes

#666 broke the game controller state filter.
The internal state would be set to Initial on startup but was never updated.
Since it was never updated, it would stay in Initial forever and because we always believe the game controller when the internal state is Initial, the filter was effectively disabled.

This PR fixes the issue by actually storing the newly calculated internal state.

Fixes #666 

## How to Test

Robots should react to a whistle in set again and sit down when game controller sends finished.


julianschuler (approved):
LGTM!

PR #727, opened 2024-02-24
Correct camera projection

okiwi6:
## Introduced Changes

The horizon check was done on the ground plane, even if the `pixel_to_ground_with_z` function was used.
This PR corrects the horizon check to be done on the elevated plane.
This also makes it possible to have the projection plane above the camera. That is why the `AboveHorizon` error was renamed since the projection plane can have the horizon on the upper or lower side.

Fixes #

## ToDo / Known Issues

## How to Test

* A test was added that fails with the old implementation, but succeeds with the new one

### DIFF
--- a/crates/projection/src/lib.rs
+++ b/crates/projection/src/lib.rs
@@ -9,7 +9,7 @@ pub enum Error {
     TooClose,
     #[error("position is behind the camera")]
     BehindCamera,
-    #[error("the pixel position cannot be projected on the projection plane")]
+    #[error("the pixel position cannot be projected onto the projection plane")]
     NotOnProjectionPlane,
 }
 
--- a/crates/projection/src/lib.rs
+++ b/crates/projection/src/lib.rs
@@ -1,4 +1,5 @@
-use nalgebra::{point, vector, Point2, Point3, Vector2, Vector3};
+use approx::relative_eq;
+use nalgebra::{point, vector, Isometry3, Point2, Point3, Vector2, Vector3};
 use thiserror::Error;
 use types::camera_matrix::CameraMatrix;
 
@@ -8,8 +9,8 @@ pub enum Error {
     TooClose,
     #[error("position is behind the camera")]
     BehindCamera,
-    #[error("the pixel position is above the horion and cannot be projected to the ground")]
-    AboveHorizon,
+    #[error("the pixel position cannot be projected to the ground")]
+    NotOnProjectionPlane,
 }
 
 pub trait Projection {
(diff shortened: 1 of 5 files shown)


crates/projection/src/lib.rs
```diff
@@ -70,20 +71,24 @@ impl Projection for CameraMatrix {
...
 
-        let distance_to_plane = z - self.camera_to_ground.translation.z;
-        let slope = distance_to_plane / camera_ray_over_ground.z;
-        let intersection_point =
-            self.camera_to_ground.translation.vector + camera_ray_over_ground * slope;
-        Ok(point![intersection_point.x, intersection_point.y])
+        let intersection_scalar = -camera_position.z / camera_ray_over_ground.z;
+
+        if intersection_scalar < 0.0 {
+            return Err(Error::NotOnProjectionPlane);
```
knoellle, line 86: I'm not whether this should return `NotOnProjectionPlane` or `BehindCamera`...

crates/projection/src/lib.rs
```diff
@@ -8,8 +9,8 @@ pub enum Error {
     TooClose,
     #[error("position is behind the camera")]
     BehindCamera,
-    #[error("the pixel position is above the horion and cannot be projected to the ground")]
-    AboveHorizon,
+    #[error("the pixel position cannot be projected to the ground")]
```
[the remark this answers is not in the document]
okiwi6: fixed

crates/projection/src/lib.rs
okiwi6: You're right, `BehindCamera` makes more sense here.

crates/projection/src/lib.rs
```diff
@@ -9,7 +9,7 @@ pub enum Error {
     TooClose,
     #[error("position is behind the camera")]
     BehindCamera,
-    #[error("the pixel position cannot be projected to the ground")]
+    #[error("the pixel position cannot be projected on the projection plane")]
```
knoellle, line 12: "to", "on"... "onto"? 😄 
image

knoellle (commented):
Tested this by projecting a grid in ground coordinates to pixels and if successful, projecting it back to ground, asserting success.
Also assert_relative_eq-ed that the resulting ground position is close to the input.

Seems to work as expected, we can merge this once the error message discussion is resolved.

crates/projection/src/lib.rs
okiwi6: done

knoellle:
> If you already have the code, we could maybe also add this as a testcase

Can do, but we'll have to think about a camera matrix to use. I just hacked my testing into the line detection by overwriting the line points with my grid and moving the robot head by hand to see if it looks like a grid on the ground.

Without the human visual confirmation, this test would also just check whether your function is a correct inverse of the ground_with_z_to_pixel function but doesn't test whether they are correct.

knoellle (approved):
Sieht gut zu mir

PR #728, opened 2024-02-25
Add Intrisic Matrix

oleflb:
## Introduced Changes
Based on #727 

This PR will be required for the robot filtering.
It adds the intrinsic matrix to the `CameraMatrix` struct. Consequently, the `impl Projection` also uses the intrinsic matrix.
For now I've decided against flipping the extrinsic coordinate system of our matrix (we don't follow typical conventions of having z along the optical axis). Instead, the flipping now happens at the same position as previously, but now it is encoded in the intrinsic matrix.

Fixes #

## ToDo / Known Issues

## How to Test

* `cargo test` :)


oleflb:
Superseded by #769

PR #729, opened 2024-02-25
Distant rainbows

knoellle:
## Introduced Changes

Cargo as well as other programs detect whether their stdout is connected to a tty/pty or file/pipe/etc.
If stdout is a terminal session (tty/pty), programs often add colors and other visual niceties.

However, ssh does not allocate a pty on the target machine by default, which causes cargo commands to lose color as well as the progress indicator (X/Y jobs done, currently running jobs) during remote compilation.

This PR add detection for tty-ish-ness to the remoteWorkspace script and instructs ssh to allocate a pty.

Fixes #

## ToDo / Known Issues

## How to Test

- [Set up a remote workspace](https://hulks.de/hulk/tooling/pepsi/#remote-compile)
- Run
```sh
# should show colored cargo output
pepsi build --remote

# should show plain cargo output since stdout is not a tty
pepsi build --remote | cat
```

### DIFF
--- a/scripts/remoteWorkspace
+++ b/scripts/remoteWorkspace
@@ -69,8 +69,17 @@ remotePath="$(echo $remoteWorkspace | cut -d':' -f 2-)"
 # send local state to remote
 rsync -ah --info progress2 --exclude='.git' --filter="dir-merge,- .gitignore" --delete . "$remoteWorkspace"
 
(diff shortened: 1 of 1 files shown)


Vivituhh (approved):
Very nice feature :)

PR #738, opened 2024-03-10
Context Privacy 2

knoellle:
## Introduced Changes

Context (heh): #535 

Removes `pub` tags from all `CreationContext` and `CycleContext` struct members since they are unnecessary and prevent detection of unused fields.

~~Also adds a check to our source analyzer which makes `pub` fields in contexts a compilation error with the exception of `MainOutput` fields.
I had wanted to make this a warning, but it appears that [warnings from proc macros are only emitted on nightly](https://docs.rs/proc-macro-error/latest/proc_macro_error/index.html#limitations).
There exists a [proc-macro-warning](https://crates.io/crates/proc-macro-warning) crate, but that only supports deprecation warnings via a hack.~~

~~An alternative approach I considered is be to have the context macro strip the pub tag. This would allow detecting unused fields but also allows inconsistency and at least in my editor screws with the syntax highlighting.~~

Fixes #

Find a way to emit warnings

## How to Test

~~Making any context field `pub` should result in a an error.~~

### DIFF
--- a/crates/control/src/foot_bumper_filter.rs
+++ b/crates/control/src/foot_bumper_filter.rs
@@ -10,7 +10,7 @@ use context_attribute::context;
 use framework::{AdditionalOutput, MainOutput};
 
 use types::{
-    cycle_time::CycleTime, fall_state::FallState, foot_bumper_obstacle::FootBumperObstacle,
+    fall_state::FallState, foot_bumper_obstacle::FootBumperObstacle,
     foot_bumper_values::FootBumperValues, sensor_data::SensorData,
 };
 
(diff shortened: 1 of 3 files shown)


crates/context_attribute/src/lib.rs
```diff
@@ -56,6 +56,11 @@ pub fn context(_attributes: TokenStream, input: TokenStream) -> TokenStream {
                     );
                 };
 
+                if field_type.as_str() != "MainOutput" {
+                    if let Visibility::Public(vis) = &field.vis {
+                        abort!(vis.pub_token, "context fields should not be public");
+                    }
+                }
```
schmidma, line 63: I am not sure about this being an error...

schmidma (changes requested):
Let's discuss... (just removing the existing `pub`s i am definitely fine with)

crates/context_attribute/src/lib.rs
knoellle: After discussion, I removed the check. This branch now only contains the removal of `pub` and unused context fields.

schmidma (approved):

PR #743, opened 2024-03-12
Update to nix 0.28

schmidma:
## Introduced Changes

https://github.com/nix-rust/nix/pull/2022 is landed in the latest release of `nix`. We can thus update to the latest release...


knoellle (approved):
Wörks, LGTM

PR #747, opened 2024-03-16
Temporary download names

knoellle:
## Introduced Changes

#746 prevents uploading broken or truncated images. In my experience, the most common source of broken images or sdk files are interrupted downloads.
This PR should prevent most issues arising from interrupted downloads by using different filenames as download target and for checking whether the file has been downloaded already.
After the download finishes, the temporary file is moved into the correct place atomically, preventing half-finished files from ending up in the final location.

Fixes #

- Maybe do the same atomic move after long interruptible process for sdk installation.
- Resumable downloads? Probably not worth it unless we run into Iran 2018 internet conditions.

## How to Test

Use with either `pepsi gammaray` or `pepsi sdk install` and abort during download. Then run the same command again and observe the result.
On main, gammaray would try to upload the truncated image file to the nao (careful, this can leave the robot in a broken state, improved by #746) and sdk installation would complain about broken file permissions, which isn't exactly helpful unless you already know what the issue is.
With this PR, the download should be simply reattempted upon running the command a second time.

### DIFF
--- a/crates/repository/src/lib.rs
+++ b/crates/repository/src/lib.rs
@@ -335,7 +335,7 @@ impl Repository {
             }
             install_sdk(installer_path, &sdk)
                 .await
-                .wrap_err("Failed to install SDK")?;
+                .wrap_err("failed to install SDK")?;
         }
         Ok(())
     }
@@ -374,7 +374,7 @@ impl Repository {
         let hardware_ids_path = self.root.join("etc/parameters/hardware_ids.json");
         let mut hardware_ids = File::open(&hardware_ids_path)
             .await
-            .wrap_err_with(|| format!("Failed to open {}", hardware_ids_path.display()))?;
+            .wrap_err_with(|| format!("failed to open {}", hardware_ids_path.display()))?;
         let mut contents = vec![];
         hardware_ids.read_to_end(&mut contents).await?;
         let hardware_ids_with_string_keys: HashMap<String, HardwareIds> = from_slice(&contents)?;
@@ -384,7 +384,7 @@ impl Repository {
                 Ok((
                     nao_number
                         .parse()
-                        .wrap_err_with(|| format!("Failed to parse NAO number: {nao_number:?}"))?,
+                        .wrap_err_with(|| format!("failed to parse NAO number: {nao_number:?}"))?,
                     hardware_ids,
                 ))
             })
@@ -503,7 +503,7 @@ async fn download_with_fallback(
             .arg(url)
             .status()
             .await
-            .wrap_err("Failed to spawn command")?;
+            .wrap_err("failed to spawn command")?;
 
         if status.success() {
             return Ok(());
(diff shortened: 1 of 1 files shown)


crates/repository/src/lib.rs
```diff
@@ -503,7 +503,7 @@ async fn download_with_fallback(
             .arg(url)
             .status()
             .await
-            .context("Failed to spawn command")?;
+            .wrap_err("Failed to spawn command")?;
```
schmidma, line 506:
```suggestion
            .wrap_err("failed to spawn command")?;
```

crates/repository/src/lib.rs
```diff
@@ -521,16 +521,23 @@ async fn download_image(
     if !downloads_directory.as_ref().exists() {
         create_dir_all(&downloads_directory)
             .await
-            .context("Failed to create download directory")?;
+            .wrap_err("Failed to create download directory")?;
```
schmidma, line 524:
```suggestion
            .wrap_err("failed to create download directory")?;
```

crates/repository/src/lib.rs
```diff
@@ -554,20 +561,25 @@ async fn download_sdk(
...
         format!("https://github.com/HULKs/meta-hulks/releases/download/{version}/{installer_name}"),
     ];
 
     println!("Downloading SDK from {}", urls[0]);
-    download_with_fallback(&installer_path, urls, CONNECT_TIMEOUT).await?;
+    download_with_fallback(&download_path, urls, CONNECT_TIMEOUT).await?;
+
+    set_permissions(&download_path, Permissions::from_mode(0o755))
+        .await
+        .wrap_err("Failed to make installer executable")?;
```
schmidma, line 578:
```suggestion
        .wrap_err("failed to make installer executable")?;
```

schmidma, line 578: i'd first move and then modify permissions, what do you think?

crates/repository/src/lib.rs
```diff
@@ -579,7 +591,7 @@ async fn install_sdk(
         .arg(installation_directory.as_ref().as_os_str())
         .status()
         .await
-        .context("Failed to spawn command")?;
+        .wrap_err("Failed to spawn command")?;
```
schmidma, line 594:
```suggestion
        .wrap_err("failed to spawn command")?;
```

crates/repository/src/lib.rs
```diff
@@ -591,11 +603,11 @@ async fn create_symlink(source: &Path, destination: &Path) -> Result<()> {
     if destination.read_link().is_ok() {
         remove_file(&destination)
             .await
-            .context("Failed to remove current symlink")?;
+            .wrap_err("Failed to remove current symlink")?;
```
schmidma, line 606:
```suggestion
            .wrap_err("failed to remove current symlink")?;
```

crates/repository/src/lib.rs
```diff
@@ -591,11 +603,11 @@ async fn create_symlink(source: &Path, destination: &Path) -> Result<()> {
...
     }
     symlink(&source, &destination)
         .await
-        .context("Failed to create symlink")?;
+        .wrap_err("Failed to create symlink")?;
```
schmidma, line 610:
```suggestion
        .wrap_err("failed to create symlink")?;
```

schmidma (approved):

PR #748, opened 2024-03-17
Upgrade to 7.0.0 scarthgap

schmidma:
## Introduced Changes

upgrades to the our latest yocto release

### DIFF
--- a/crates/code_generation/src/run.rs
+++ b/crates/code_generation/src/run.rs
@@ -46,6 +46,10 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
 
             #start_cyclers
 
+            #[cfg(feature = "systemd")]
+            systemd::daemon::notify(false, [(systemd::daemon::STATE_READY, "1")].iter())
+                .wrap_err("failed to contact SystemD for ready notification")?;
+
             let mut encountered_error = false;
             #join_cyclers
             match recording_thread.join() {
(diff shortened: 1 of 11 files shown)


crates/code_generation/src/run.rs
```diff
@@ -46,6 +46,10 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
 
             #start_cyclers
 
+            #[cfg(feature = "systemd")]
+            systemd::daemon::notify(false, [(systemd::daemon::STATE_READY, "1")].iter())
```
knoellle, line 50: Use [`std::iter::once()`](https://doc.rust-lang.org/std/iter/fn.once.html) here.

crates/constants/src/lib.rs
```diff
@@ -13,8 +13,8 @@ pub const HULA_DBUS_PATH: &str = "/org/hulks/HuLA";
 pub const HULA_DBUS_SERVICE: &str = "org.hulks.hula";
 pub const HULA_SOCKET_PATH: &str = "/tmp/hula";
 pub const OS_RELEASE_PATH: &str = "/etc/os-release";
-pub const OS_VERSION: &str = "5.9.1";
-pub const SDK_VERSION: &str = "5.9.0";
+pub const OS_VERSION: &str = "6.0.0";
+pub const SDK_VERSION: &str = "6.0.0";
```
knoellle, line 17: 7.0.0

knoellle (changes requested):

crates/constants/src/lib.rs
schmidma: Done

crates/code_generation/src/run.rs
schmidma: Done, changed it also from my copy source in hula ;)

knoellle (approved):
LGTM

PR #749, opened 2024-03-17
Colorize Aliveness Output

schmidma:
## Introduced Changes

Restructures the aliveness output and adds colors :)


oleflb (commented):
Generally LGTM. Do you think it poses a problem that `pepsi aliveness > output` will not work properly now anymore because of the colors?

knoellle:
https://docs.rs/owo-colors/4.0.0/owo_colors/#only-style-on-supported-terminals

schmidma:
Is aliveness output actually written to a file anywhere? Your suggestion, @knoellle, sounds like a lot of overhead to wrap all styling expressions with this conditional thingy?!

knoellle:
> Is aliveness output actually written to a file anywhere?

I'm not aware of anyone using it like that.
For further processing, there's always the `--json` flag, whose output should never contain ANSI escape sequences.
And in cases where people use `pepsi aliveness` indirectly, i.e. with `watch/viddy` or saving it to a file, it is likely shown to a user as-is at some point anyways, in which case colors are likely desirable.

One thing to note though: `watch pepsi aliveness` will look terrible, since `watch` doesn't render colors correctly unless explicitly enabled with `-c/--color`.
[Viddy](https://github.com/sachaos/viddy) has colored output by default, but with watch, you will always have to use `watch -c pepsi aliveness`.

knoellle:
Don't forget to bump the pepsi version

tools/pepsi/src/aliveness.rs
```diff
@@ -58,93 +58,137 @@ pub async fn aliveness(arguments: Arguments) -> Result<(), Error> {
...
+const BATTERY_CHARGE_WARN: f32 = 0.7;
+const CHARGING_ICON: &str = "󱐋";
+const DISCHARGING_ICON: &str = "󰁽";
+const OS_ICON: &str = "󱑞";
+const ALL_OK_ICON: &str = "✔";
+const UNKNOWN_CHARGE_ICON: &str = "󰁽?";
+const NETWORK_ICON: &str = "󰖩 ";
+const TEMPERATURE_ICON: &str = "";
+const TEMPERATURE_WARN_THRESHOLD: f32 = 40.0;
+const TEMPERATURE_ERROR_THRESHOLD: f32 = 60.0;
```
schluis, line 76: Please use the values the ears use:
```            
const TEMPERATURE_LEVEL_ONE: f32 = 76.0;
const TEMPERATURE_LEVEL_TWO: f32 = 80.0;
const TEMPERATURE_LEVEL_THREE: f32 = 90.0;
```

oleflb (approved):
LGTM

tools/pepsi/src/aliveness.rs
schmidma: I added `45` for warning as this is as soon as the robot was used. and `80` for error level as this robot should first be cooled down before using it.

PR #768, opened 2024-03-23
Hybrid alignment fix

### LINKED ISSUE #643: Fix hybrid alignment
When we run to the kick-off position after scoring a goal, the striker seems to oscillate in the hybrid orientation when he reaches the end of his path. This means that he oscillates between turning left and right and sometimes never reaches the final kick-off position, resulting in an illegal position.

knoellle:
## Introduced Changes

Hybrid alignment was completely bronek when approaching a target position from the opposite direction. See associated issue for problem details.

This PR changes the hybrid alignment behavior to slowly reduce the allowed deviation from the target orientation based on the distance to the target.
Effectively, there is a cone of allowed orientations around the target orientations which shrinks when the robot gets closer.
The requested orientation then the current robot orientation (Identity in Ground), clamped to this allowed region.

Fixes #643 

## How to Test

Set up game with at least a kick off striker and a game controller running.
After the striker scores a goal, it runs back to the kick off position, if the robot crosses the X axis of the target position within the hybrid align distance, it should no longer try to turn the other way and get stuck.
In every other situation, it should behave just like before.

### DIFF
--- a/crates/control/src/behavior/walk_to_pose.rs
+++ b/crates/control/src/behavior/walk_to_pose.rs
@@ -3,7 +3,7 @@ use std::f32::consts::PI;
 use coordinate_systems::{Field, Ground};
 use filtering::hysteresis::less_than_with_hysteresis;
 use framework::AdditionalOutput;
-use linear_algebra::{point, Isometry2, Point, Point2, Pose, UnitComplex};
+use linear_algebra::{point, Isometry2, Orientation2, Point, Point2, Pose};
 use types::{
     field_dimensions::FieldDimensions,
     motion_command::ArmMotion,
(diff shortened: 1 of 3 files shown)


crates/control/src/behavior/walk_to_pose.rs
```diff
@@ -211,10 +213,96 @@ pub fn hybrid_alignment(
...
+            (FRAC_PI_2, FRAC_PI_2),
+            (-FRAC_PI_2, FRAC_PI_2),
+            (PI, FRAC_PI_2),
+            (-PI, FRAC_PI_2),
+        ];
+
+        for (input, angle_limit) in testcases {
+            let input = Orientation2::new(input);
+            let center = Orientation2::new(0.0);
+            assert_relative_eq!(
```
oleflb, line 284: also compare sign here? (or separate test)

oleflb (changes requested):
Nice, great improvement in the hybrid alignment! Just one remark on the testcase.

crates/control/src/behavior/walk_to_pose.rs
knoellle: Done

oleflb (approved):
:+1:

Issue #760, opened 2024-03-20
Postgame fails to run dmesg on image 7.0.0


h3ndrk:
`dmesg` needs root permissions now. A `sudo` in the `nao` crate fixes it but do we want this? Maybe something like https://unix.stackexchange.com/a/390187 in Yocto?

PR #770, opened 2024-03-25
Run dmesg with sudo

knoellle:
## Introduced Changes

What it says, more infos here: #760
Keeping issue open to find better solution.

Fixes #

## How to Test

`pepsi logs download ...` should work now


h3ndrk (approved):
Should work

PR #771, opened 2024-03-25
Allow specifying log path

knoellle:
## Introduced Changes

This PR adds an option to the hulk_nao binary which lets the caller specify where logs should be stored.
It also removes the timestamp from replay files.
The idea is to move the timestamp from the individual file names to the directory file name.

Since the log path defaults to just `logs`, for the time being, this will lead to replay files overwriting older ones until we deploy a new image where the `launchHULK` script creates a directory for each run and points the hulk binary to save its logs there.

Fixes #

## How to Test

build hulk for nao and upload, then run the following on the nao:
```
cd hulks
bin/hulk --log-path logs/my_awesome_replay
```
This should put all of the bincode files into the specified directory.

### DIFF
--- a/crates/hulk_nao/src/main.rs
+++ b/crates/hulk_nao/src/main.rs
@@ -1,6 +1,12 @@
 #![recursion_limit = "256"]
-use std::{env::args, fs::File, io::stdout, sync::Arc};
+use std::{
+    fs::File,
+    io::stdout,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
+use clap::Parser;
 use color_eyre::{
     eyre::{Result, WrapErr},
     install,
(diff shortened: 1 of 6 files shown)


h3ndrk:
Instead of adding command line parsing etc., we could continue to use the `framework.json` which can already be passed as parameter and would be a single source of truth for such paths.

knoellle:
I don't think that's such a good idea. The framework json is for persistent configuration. The whole point of this PR is to use a different log path for each execution where the path contains some sort of timestamp.
I don't think having the launchscript rewrite the framework.json is a good idea.

crates/hulk_nao/src/main.rs
```diff
@@ -39,12 +45,18 @@ pub fn setup_logger() -> Result<(), fern::InitError> {
...
 }
 
+#[derive(Parser)]
+struct Arguments {
+    #[arg(short, long, default_value = "logs")]
+    log_path: String,
+
+    #[arg(short, long, default_value = "etc/parameters/framework.json")]
+    framework_parameters_path: String,
+}
```
h3ndrk, line 55: Does this work?

```suggestion
struct Arguments {
    #[arg(short, long, default_value = "logs")]
    log_path: PathBuf,

    #[arg(short, long, default_value = "etc/parameters/framework.json")]
    framework_parameters_path: PathBuf,
}
```

knoellle: I tried that first but sadly not, the field type must support Display, which PathBuf doesn't.

knoellle: Huh...
Looks like this doesn't work with `default_value_t`, but does with `default_value`, changed.

schmidma (approved):
LGTM

PR #772, opened 2024-03-25
Replay parameters

### LINKED ISSUE #764: Simplify ID passing for parameter retrieval during replay
Since recording files only contain the inputs at the beginning of cyclers and the node states, but not the outputs of each node, the outputs are recomputed during replay. This requires configuration parameters, ideally the same as during recording. When assuming the parameters locally are the same as on the NAO, the replayer just needs the IDs to retrieve the same parameters. Currently, the head and body IDs need to be entered in the replayer. Obtaining them is very tedious and should be improved.

knoellle:
## Introduced Changes

Based on #771
The hulk binary now saves the flattened initial parameters to the log path.
Flattened in this case refers to the merging of default.json and location/robot -specific parameters.

Replay is then pointed to the log directory as the parameters directory from which it loads the bincode files and initial parameters.

Fixes #764 

## How to Test

Create a replay with this branch, ideally on a nao, download it and start the replayer with it.
Check via twix whether the parameters are correct, e.g. check if the `camera_matrix_parameters` match the robot-specific ones in `etc/parameters/head.[...].json`

### DIFF
--- a/crates/code_generation/src/execution.rs
+++ b/crates/code_generation/src/execution.rs
@@ -249,11 +249,13 @@ fn generate_recording_thread(cyclers: &Cyclers) -> TokenStream {
     quote! {
         {
             let keep_running = keep_running.clone();
+            let parameters_reader = communication_server.get_parameters_reader();
             std::thread::Builder::new()
                 .name("Recording".to_string())
                 .spawn(move || -> color_eyre::Result<()> {
                     let result = (|| {
                         use std::io::Write;
+                        std::fs::write(log_path.as_ref().join("default.json"), serde_json::to_string_pretty(&*parameters_reader.next())?)?;
                         #(#file_creations)*
                         for recording_frame in recording_receiver {
                             match recording_frame {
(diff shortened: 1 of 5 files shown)


crates/code_generation/src/execution.rs
```diff
@@ -247,12 +249,13 @@ fn generate_recording_thread(cyclers: &Cyclers) -> TokenStream {
     quote! {
         {
             let keep_running = keep_running.clone();
+            let parameters_reader = communication_server.get_parameters_reader();
```
h3ndrk, line 252: Is this dropped correctly s.t. nothing leaks and prevents graceful shutdown? Hard to see via GitHub app from my side... 🤪

knoellle: I didn't notice an issue like that, but good point. I'll have a look at this again.

knoellle: The reader is moved into the recording thread and is dropped when the recording thread exits.
We could also fetch the parameters here and then move them by value into the new thread if you prefer.

h3ndrk: Ok and probably the recording thread terminates when all cyclers terminate. Since this is the same point in time when the other parameter readers are dropped, should be fine then.

PR #774, opened 2024-03-25
Interpret EOF errors as soft errors

### LINKED ISSUE #767: Enable the recording index to handle truncated files
If NAOs die without a chance to write all recording data to disk, recording files may be truncated at non-frame-aligned lengths. Currently, the recording index of the replayer produces a hard error if encountering such files. In these situations, the replayer cannot be used. Instead we should harden the recording index to handle truncated files more softly s.t. the replayer can be used in such situations.

h3ndrk:
## Introduced Changes

Fixes #767

## ToDo / Known Issues

- How to handle soft errors?

None

## How to Test

Use this dumper application for generating a synthetical recording file with correct format:

```rust
use std::{
    fs::File,
    io::{Seek, Write},
    time::SystemTime,
};

use bincode::serialize_into;

fn main() {
    let mut file = File::create("logs/Control.bincode").unwrap();
    dbg!(file.stream_position().unwrap());
    serialize_into(&mut file, &SystemTime::now()).unwrap();
    dbg!(file.stream_position().unwrap());
    serialize_into(&mut file, &42_usize).unwrap();
    dbg!(file.stream_position().unwrap());
    file.write_all(&[42; 42]).unwrap();
    dbg!(file.stream_position().unwrap());
    serialize_into(&mut file, &SystemTime::now()).unwrap();
    dbg!(file.stream_position().unwrap());
    serialize_into(&mut file, &42_usize).unwrap();
    dbg!(file.stream_position().unwrap());
    file.write_all(&[42; 42]).unwrap();
    dbg!(file.stream_position().unwrap());
}
```

```toml
[package]
name = "dumper"
version.workspace = true
edition.workspace = true
license.workspace = true
homepage.workspace = true

[dependencies]
bincode = { workspace = true }
```

Then use the following truncated file sizes for testing:

- Happy path: complete file, should work :tm:
- `truncate -s 100 ...`: this truncates the data, seeking beyond the end of file will not yield any errors
- `truncate -s 82 ...`: now the file has zero data and is truncated exactly after the header, should be same as the previous case
- `truncate -s 81 ...`: the header misses one byte, parsing the length will fail, but the recording index should still contain the first complete frame, the second frame is ignored (soft error)
- `truncate -s 73 ...`: same as the previous case but with the timestamp

### DIFF
--- a/crates/framework/src/recording_index.rs
+++ b/crates/framework/src/recording_index.rs
@@ -38,11 +38,13 @@ impl RecordingIndex {
                 end_of_file_error_as_option(deserialize_from(&mut recording_file))
                     .wrap_err("failed to deserialize timestamp")?
             else {
+                eprintln!("unexpected end of file of recording file while deserializing timestamp");
                 break;
             };
             let Some(length) = end_of_file_error_as_option(deserialize_from(&mut recording_file))
                 .wrap_err("failed to deserialize data length")?
             else {
+                eprintln!("unexpected end of file of recording file while deserializing length");
                 break;
             };
             let header_offset = recording_file
@@ -52,6 +54,10 @@ impl RecordingIndex {
             recording_file
                 .seek(SeekFrom::Current(length as i64))
                 .wrap_err("failed to seek to end of data")?;
+            if offset + header_offset + length as u64 > end_of_file_offset {
+                eprintln!("unexpected end of file of recording file");
+                break;
+            }
             frames.push(RecordingFrameMetadata {
                 timestamp,
                 offset: offset.try_into().unwrap(),
--- a/crates/framework/src/recording_index.rs
+++ b/crates/framework/src/recording_index.rs
@@ -1,12 +1,12 @@
 use std::{
     fs::File,
-    io::{Read, Seek, SeekFrom},
+    io::{self, Read, Seek, SeekFrom},
     path::Path,
     time::SystemTime,
 };
 
-use bincode::deserialize_from;
-use color_eyre::{eyre::WrapErr, Result};
+use bincode::{deserialize_from, Error};
+use color_eyre::eyre::WrapErr;
 
 #[derive(Debug)]
 pub struct RecordingIndex {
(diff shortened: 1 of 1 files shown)


crates/framework/src/recording_index.rs
```diff
@@ -34,10 +34,17 @@ impl RecordingIndex {
...
         while offset < end_of_file_offset {
-            let timestamp = deserialize_from(&mut recording_file)
-                .wrap_err("failed to deserialize timestamp")?;
-            let length = deserialize_from(&mut recording_file)
-                .wrap_err("failed to deserialize data length")?;
+            let Some(timestamp) =
+                end_of_file_error_as_option(deserialize_from(&mut recording_file))
+                    .wrap_err("failed to deserialize timestamp")?
+            else {
+                break;
```
h3ndrk, line 41: Handle the soft error differently?

knoellle:
While this implementation successfully prevents errors if the file is truncated within the recording frame *headers*, the replayer still crashes when trying to load an incomplete frame where the file is truncated within the data segment.
I think a frame should not be added if parsing time and length is successful but `offset + length > end_of_file_offset`.

The recording index constructor also doesn't return any indication whether the file is intact or if there was an error, I think the user should be told if there was a problem.
Maybe we can return a tri-state enum with `Ok(index)`, `Partial(index)`, `Error(error)`?

h3ndrk:
Yes, truncated data is not correctly handled.

If we want to return the error, we  need to do two things:

1. propagate the error through all abstraction layers until we inspect it or display it somewhere (mixed generated/static code, color_eyre Results etc.)
2. accumulate the errors because n recording indices but 1 replayer

That means some effort. Alternative is to ignore errors which is obviously simpler. Where do we want to set the trade-off? How important are error messages? How does the user benefit from them? The file cannot be recovered/untruncated. I think, the most important detail is that we can still load these files, not to know that they are truncated. What do you think?

h3ndrk:
I thought about logging, but was hesitant to include it into the framework crate but maybe that is fine? Or `eprintln!`?

Yes, I meant to not include a truncated frame in the index. Otherwise we would run into errors at load time, as you mentioned.

crates/framework/src/recording_index.rs
```diff
@@ -34,17 +34,30 @@ impl RecordingIndex {
 
         let mut offset = 0;
         while offset < end_of_file_offset {
```
knoellle, line 36: `end_of_file_offset` -> `file_size/length`?

knoellle (approved):
LGTM

PR #780, opened 2024-03-29
Improvements to `linear_algebra`

schmidma:
## Introduced Changes

Several improvements to our `linear_algebra` crate. Including splitting up modules for better readability...


crates/linear_algebra/src/isometry.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// 3 Dimension
+
+impl<From, To, T> Isometry3<From, To, T>
+where
+    T::Element: SimdRealField,
+    T: SimdRealField + Copy,
+{
+    pub fn from_parts(translation: Vector3<To, T>, orientation: Orientation3<To, T>) -> Self {
```
knoellle, line 85: I don't like that this function has a different name than the 2D variant.
Do you think it makes sense to stray from the original nalgebra interface here?

crates/linear_algebra/src/orientation.rs
```diff
@@ -0,0 +1,74 @@
...
+{
+    pub fn new(angle: T) -> Self {
+        Self::wrap(nalgebra::UnitComplex::new(angle))
+    }
+
+    pub fn identity() -> Self {
+        Self::wrap(nalgebra::UnitComplex::identity())
+    }
+
+    pub fn inverse(&self) -> Self {
```
knoellle, line 21: I don't think it makes sense to "invert" an orientation.

```suggestion
    pub fn mirror(&self) -> Self {
```

crates/linear_algebra/src/orientation.rs
```diff
@@ -0,0 +1,74 @@
...
+        Self::wrap(nalgebra::UnitQuaternion::new(axis_angle.inner))
+    }
+
+    pub fn from_euler_angles(roll: T, pitch: T, yaw: T) -> Self {
+        Self::wrap(nalgebra::UnitQuaternion::from_euler_angles(
+            roll, pitch, yaw,
+        ))
+    }
+
+    pub fn inverse(&self) -> Self {
```
knoellle, line 71: I don't think it makes sense to "invert" an orientation.

```suggestion
    pub fn mirror(&self) -> Self {
```

crates/linear_algebra/src/vector.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    pub fn xz(&self) -> Vector2<Frame, Scalar> {
+        Vector2::wrap(self.inner.xz())
+    }
+
+    pub fn yz(&self) -> Vector2<Frame, Scalar> {
+        Vector2::wrap(self.inner.yz())
+    }
+}
```
knoellle, line 161: You split this from the `x`, `y`, `z` block above, but they are in the same block for the `Point` implementation, why?

knoellle (changes requested):

crates/linear_algebra/src/framed.rs
```diff
@@ -94,33 +94,33 @@ where
     }
 }
 
-impl<Frame, Inner> Mul<f32> for Framed<Frame, Inner>
+impl<Frame, Inner, T> Mul<T> for Framed<Frame, Inner>
```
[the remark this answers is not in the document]
schmidma: Just checking, why do you think the second one is dangerous? Because we combine *safe* elements with *unsafe* computation?

crates/linear_algebra/src/isometry.rs
schmidma: Then I'd rather suggest to stray from the interface in the 2d variant. I like the `from_parts` naming. If you agree I'd change the `Isometry2::new` to `Isometry2::from_parts`?

crates/linear_algebra/src/orientation.rs
schmidma: I agree. Done.

crates/linear_algebra/src/orientation.rs
schmidma: I agree. Done.

crates/linear_algebra/src/vector.rs
schmidma: I was just lazy... I first tried to have minimal trait bounds. But because of readability I just assumed some basic bounds. To further increase readability, I now changed it to merge even more basic constraints. like `Copy + Zero + One` etc... -> YAGNI

crates/linear_algebra/src/framed.rs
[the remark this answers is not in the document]
knoellle: Yes.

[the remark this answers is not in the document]
schmidma: Okay, I added a
```rust
T: SimdComplexField
```
to the `Mul`, `MulAssign` and `Div` implementations. This constrains the `T` type to only scalar values. What do you think? IMHO this is in line with all the other nalgebra styled trait bounds.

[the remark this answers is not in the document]
schmidma: Na... I take that back... Let me try for a minute. `u32` for example does not implement that...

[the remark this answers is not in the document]
schmidma: Okay, I went with `num_traits::Num`. This includes all numbers.

crates/linear_algebra/src/isometry.rs
schmidma: Done

knoellle (approved):
LGTM

PR #781, opened 2024-03-29
Move lines to `geometry`

schmidma:
What it says, just moving

### DIFF
--- a/crates/geometry/src/line.rs
+++ b/crates/geometry/src/line.rs
@@ -16,6 +16,7 @@ pub struct Line<Frame, const DIMENSION: usize>(
 );
 
 pub type Line2<Frame> = Line<Frame, 2>;
+pub type Line3<Frame> = Line<Frame, 3>;
 
 impl<Frame> Line2<Frame> {
     pub fn signed_acute_angle(&self, other: Self) -> f32 {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,9 +123,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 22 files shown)


crates/geometry/src/line.rs
```diff
@@ -143,9 +144,10 @@ impl<Frame, const DIMENSION: usize> Line<Frame, DIMENSION> {
     }
 }
 
-impl<From, To, const DIMENSION: usize> Mul<Line<From, DIMENSION>> for Isometry<From, To, DIMENSION>
+impl<From, To, const DIMENSION: usize, Inner> Mul<Line<From, DIMENSION>>
+    for Transform<From, To, Inner>
 where
-    Isometry<From, To, DIMENSION>: Mul<Point<From, DIMENSION>, Output = Point<To, DIMENSION>>,
+    Self: Mul<Point<From, DIMENSION>, Output = Point<To, DIMENSION>> + Copy,
```
knoellle, line 150: Isn't this the same change as in #780?

crates/geometry/src/line.rs
```diff
@@ -17,6 +17,7 @@ pub struct Line<Frame, const DIMENSION: usize>(
 );
 
 pub type Line2<Frame> = Line<Frame, 2>;
+pub type Line3<Frame> = Line<Frame, 3>;
```
knoellle, line 20: Is this used anywhere?

knoellle (changes requested):
I like it but have some minor remarks

crates/geometry/src/line.rs
schmidma: True, I reverted this change here. Should be the same as in main now...

crates/geometry/src/line.rs
schmidma: No, just there for completeness ;) Did not hurt, and the next person is lucky

knoellle (approved):
LGTM

knoellle (approved):
LGTM

PR #791, opened 2024-04-06
Add Y axis offset for penalty poses

### LINKED ISSUE #645: Penalty Rule 2024
In the 2024 rules the robot is now placed outside the field looking inside when penalized.
We still assume the robot stands on the line.

MaikRe:
## Introduced Changes

What it says.
Also added covariance separately from initial poses because it is now more common that X position is is not at penalty spot.

Fixes #645 

Covariance might need adjustments as testing is conducted, distance must be adjusted based on field that is being played on. If the border of the field is smaller than 50cm, the robots will not be placed closer to field.

## How to Test

Penalize robot in test game, placing it in different positions left and right of the penalty spot and outside of the field (20cm for SMD). Ensure robot localization still settles to being accurate.

### DIFF
--- a/crates/control/src/localization.rs
+++ b/crates/control/src/localization.rs
@@ -83,6 +83,9 @@ pub struct CycleContext {
     minimum_fit_error: Parameter<f32, "localization.minimum_fit_error">,
     odometry_noise: Parameter<Vector3<f32>, "localization.odometry_noise">,
     player_number: Parameter<PlayerNumber, "player_number">,
+    penalized_distance: Parameter<f32, "localization.penalized_distance">,
+    penalized_hypothesis_covariance:
+        Parameter<Matrix3<f32>, "localization.initial_hypothesis_covariance">,
     score_per_good_match: Parameter<f32, "localization.score_per_good_match">,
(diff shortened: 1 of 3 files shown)


crates/control/src/localization.rs
```diff
@@ -83,6 +83,9 @@ pub struct CycleContext {
     minimum_fit_error: Parameter<f32, "localization.minimum_fit_error">,
     odometry_noise: Parameter<Vector3<f32>, "localization.odometry_noise">,
     player_number: Parameter<PlayerNumber, "player_number">,
+    penalized_distance: Parameter<f32, "localization.penalized_distance">,
+    penalized_hypothesis_covariance:
+        Parameter<Matrix3<f32>, "localization.initial_hypothesis_covariance">,
```
h3ndrk, line 88: Shouldn't this be

```suggestion
        Parameter<Matrix3<f32>, "localization.penalized_hypothesis_covariance">,
```

h3ndrk (changes requested):

crates/control/src/localization.rs
MaikRe: Yes

h3ndrk (approved):

Issue #147, opened 2023-03-16
Improve type `TwoLineSegments`


h3ndrk:
This tuple type requires a manual `SerializeHierarchy` implementation. Maybe convert to a struct type and derive it simply?

tuxbotix:
@h3ndrk This might sound like a stupid question, but is there a reason/ limitation that prevents us from extending `SerializedHierarchy` to support tuple structs or is there a technical reason not to? While I was doing #793 , I felt that tuple struct would still be cool to have.

I was wondering if field names can be made up with `0`, `1`, etc for serialization purposes or mark these cases as tuples, or use `(a, b, c)` syntax?

h3ndrk:
I don't know of any technical limitation. The derive macro should™️ be implementable with `.0` etc. For deciding whether it is worth, have a look how often we would derive it for tuple structs and weigh it against the macro implementation effort.

tuxbotix:
Thanks for the explanation. I think such a derive might ease the pains of refactoring a large amount of code (see #793).
You also have a good point on "how often", I'll give it a try to see how much pain can one get into :bomb:

PR #793, opened 2024-04-06
Line with SerializedHierarchy

tuxbotix:
## Introduced Changes

It relates to https://github.com/HULKs/hulk/issues/147, but Fixing `Line` first.

I'm not convinced if the naming is the best, but 0 or 1 are invalid field names making them impossible to use.

Fixes #

## ToDo / Known Issues
..

Complete coverage for : https://github.com/HULKs/hulk/issues/147
Or find an alternative approach to handle Tuple structs with SerializedHierarchy!

## How to Test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -917,6 +917,8 @@ dependencies = [
  "linear_algebra",
  "nalgebra",
  "projection",
+ "serde",
+ "serialize_hierarchy",
  "thiserror",
  "types",
 ]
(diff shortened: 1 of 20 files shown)


knoellle (changes requested):
Why don't you just manually implement the trait for `Line`?
Then you wouldn't have to any of the usages.

tuxbotix:
> Why don't you just manually implement the trait for `Line`? Then you wouldn't have to any of the usages.

@knoellle I forgot to mention, https://github.com/HULKs/hulk/pull/797 can kill this MR as that introduces `SerializeHierarchy` for enum structs. Could you please take a look and lets figure out what's the best option?

A workaround option is to add `leaf` annotation to the line types but I personally prefer the above MR as the best option.

schmidma:
Superseded by #987

Issue #127, opened 2023-03-08
Behavior for automatic calibration


h3ndrk:
Some parts are already done in an old branch.

- Button presses with a state machine
- Walking to fixed positions
- Looking at specific features on the field
- While looking at features trigger other modules and start calibration
- Wait for calibration to finish
- Maybe have a calibration calculation phase at the end

Depends on all issues with label `calibration:Manual`.

Time Estimation: 8 weeks

schmidma:
Rethink for K1. Closing for now

PR #798, opened 2024-04-07
Automatic calibration controller (Der dessen name nicht genannt werden darf II)

### LINKED ISSUE #131: Control calibration node for automatic calibration
- Accumulates all vision inputs and the camera matrices etc.
- Algorithms
    - Penalty/Goal Box Calibration
        - Fitting algorithm was working on a small test data set
        - A lot of data has been accumulated from a badly lighted field
    - Center Circle Calibration
        - Works on some extracted data with some noise

Depends on #130 

Time Estimation: 8 weeks

tuxbotix:
## Introduced Changes

The current work in #695 mainly provides the calibration line detection and extracting measurements (although the PR originally meant to implement the integration too).

- This PR will provide the behaviour and coordination components
    - Extracting measurements or running actual calibration is out of the scope
- ~#695 can be cleaned for calibration line detection and measurement extraction.~ -> Replaced by #1057

Whichever PR gets merged last, will connect the final integration (writing calibration results, validation, etc) or as a 3rd PR.

Depends on #1076 (for datatypes & workflow testing) - It can be decoupled if nessesary.
Fixes #131

State changes:
- Controller's commands (blue thin stripe is "capture"
- other two are vision cyclers, Yellow is "captured value" -> fake measurement generator randomly deciding to send or not.
image

## Workflow

The controller node tracks the state of the primary state and issues the calibration commands when the primary state is set to Calibrate.

Commands & State

- *Inactive*
  - Do nothing when the primary state is anything other than `Calibrate`
  - Transition to Initialize when the primary state goes to `Calibrate` and also set the cycle time.
- *Initialize* `{ started time }`
  - Active during initialization, waits for `initial_stabilization_delay` until the robot finishes movement and stabilizes.
  - Once the timeout is reached, transition to LookAt with the stored LookAt command list. If nothing is found, jump to _Processing_ phase.
- *LookAt* `{ target, camera, dispatch_time }`
  - Move the head to the given target.
  - Waits `look_at_dispatch_waiting` from `dispatch_time` to ensure the head movement is complete.
  - Once the timeout is complete, switch to 'Capture'
- *Capture* `{ dispatch_time }`
  - The vision node extracts measurements using the camera images. Shall do necessary filtering in this phase as well.
  - Waits until a measurement arrives with this information & stores it.
      - This may need improvements - See TODOs
  - Transition to the next *LookAt* once capture information arrives or jump to Processing if all look-at commands are executed.
- *Processing*
  - Runs the solver, gets the results and persists to storage if satisfactory.
  - Transition to *Finish* once above is complete.
- *Finish*
  - End state. Does nothing until exiting calibration primary phase.
  - Auto-change primary state??

## ToDo / Known Issues

- [x] Verify that the use of `PerceptionInput` is correct
- [x] Handle when a measurement cannot be obtained from a given position (retry?, move to next look-at?
    - vision retries and says `RetriesExceeded` if it can't find any measurement.
    - If both cameras fail to find a measurement, it'll goto the next one.
- [x] Consider where to run the actual calibration processing -> on the node, on a future or a separate cycler
    - Done on the controller as of now, seems to be fast enough*
- [x] CalibrationCommand (also tracks state) could be improved.
    - Further actions added for Next Iterations
- [x] LookAt commands use a timeout to decide the end of the movement. Consider alternatives
   - Won't consider anything else for this PR
- [x] The initial phase (after activating calibration) uses a timeout (instead of checking if the pose is reached and stable). Consider alternatives
   - Won't consider anything else for this PR
- [x] Remove logging and replace with `AdditionalOutputs`
- [x] Nested `match` and `if` blocks could be made more elegant/ flatter
    - Done as much as I can. Reviewer could help!

- Make look-at commands configurable (got a fixed list for now)
- Finalize the integration and include tests for the entire workflow.
- ~Should consider adding camera to the `capture` command.~ (already done)
- Decide on keeping the timeout for look-at or check for reaching the look-at status (joint value?)
- Decide on keeping the timeout for initial to calibration start - might want to look at IMU, etc values for stability?
- Should we automatically switch to `Initial` state after reaching `Finish` state in calibration?
    - https://github.com/HULKs/hulk/issues/1092

## How to Test

Webots or a NAO:
- On Twix, subscribe to:
    - `Control.main_outputs.calibration_command` to look at the calibration phase
    - `Control.main_outputs.primary_state` to verify the transition from `initial` to `calibration` (useful for Webots)
- Enter the calibration mode from Initial, (press `ctrl + shift + x` on webots for 1 second of "simulated time")
- If on webots, `pepsi run` will print the state changes and measurement collection on the terminal


schmidma:
This looks extremely wip, but resides in "Request for Review"... Do you expect of getting a review on your code in the current state? If not, please move to "In Progress"

tuxbotix:
Hi,
Yes it is quite WIP and as written on Telegram when I opened it, this was
1.5 days work.

What I wanted by putting for request for review is to get first remarks and
feedback on the architecture and the implementation approach before
investing any more time.

If it is better, I can put this back to in development and discuss with you
and/or whoever else interested to decide if this is a good approach to
continue.

I simply wanted to avoid a repeat of https://github.com/HULKs/nao/pull/2951
where the devised approach wasn't all that great. At the same time, trying
the implementation on this PR was much faster to answer the questions I had
about this approach regarding the framework.

tuxbotix:
Fixed majority of the TODO/ WIP stuff. Works on Webots (Will test on a NAO today)

It would be great if a review can be done on this status to confirm the design choices and finalize this PR.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+                    self.current_measurements.push(measurement);
+                }
+
+                if goto_next_position_top || goto_next_position_bottom {
+                    Some(self.try_transition_to_look_at(*context.cycle_time))
+                } else {
+                    None
+                }
+            }
+            CalibrationCommand::PROCESS => {
```
knoellle, line 189: Why do you need this state?
Can't you directly do the processing when there are no more targets to look at?

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+                    Some(CalibrationCommand::INITIALIZE {
+                        started_time: *current_cycle_time,
+                    })
+                } else {
+                    None
+                }
+            }
+            CalibrationCommand::INITIALIZE {
+                started_time: activated_time,
+            } => match context.primary_state {
```
knoellle, line 140: You don't need this match statement.
If we are not in `PrimaryState::Calibration`, this code will never execute anyway since `cycle` returns early.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+            current_measurements: vec![],
+            current_primary_phase_is_calibration: false,
+            last_calibration_corrections: None,
+        })
+    }
+
+    pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
+        let calibration_grame_state_active =
+            matches!(context.primary_state, PrimaryState::Calibration);
+        if !calibration_grame_state_active {
```
knoellle, line 77:
```suggestion
        if *context.primary_state != PrimaryState::Calibration {
```

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+    fn get_next_command(
+        &mut self,
+        primary_state_transition_to_calibration: bool,
+        current_cycle_time: &CycleTime,
+        context: &CycleContext,
+    ) -> Option<CalibrationCommand> {
+        let look_at_dispatch_waiting = context.parameters.look_at_dispatch_wait_duration;
+        let initial_stabilization_delay = context
+            .parameters
+            .initial_to_calibration_stabilization_delay;
```
knoellle, line 122: Why are these two different names?
Also, do you really need this binding? I would suggest just using `context.stabilization_delay` in place of this.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+                if time_diff > look_at_dispatch_waiting {
+                    Some(CalibrationCommand::CAPTURE {
+                        dispatch_time: *current_cycle_time,
+                    })
+                } else {
+                    None
+                }
+            }
+            CalibrationCommand::CAPTURE { dispatch_time } => {
+                // TODO decide a better way to handle the two cameras. Perhaps set a capture position like in the look-at command?
```
knoellle, line 170: Are you capturing with both cameras at the same time?
If you are trying to look at a position with the top camera but VisionBottom runs once and successfully captures a measurement. Wouldn't you then transition to LookAt again and start moving the head without ever getting data from VisionTop?

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+    ];
+
+    let attach_camera_to_lookat =
+        |point: &Point2<Ground>,
+         camera_position: &CameraPosition|
+         -> (Point2<Ground>, Option<CameraPosition>) { (*point, Some(*camera_position)) };
+
+    Ok(look_at_points
+        .iter()
+        .map(|&point| attach_camera_to_lookat(&point, &CameraPosition::Top))
```
knoellle, line 297:
```suggestion
        .map(|&point| (point, CameraPosition::Top))
```

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+        .collect_vec())
```
knoellle, line 298: You don't need the specialization here.
```suggestion
        .collect())
```

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+        point!(3.0, 0.5),
+        point!(1.0, -0.5),
+    ];
+
+    let attach_camera_to_lookat =
+        |point: &Point2<Ground>,
+         camera_position: &CameraPosition|
+         -> (Point2<Ground>, Option<CameraPosition>) { (*point, Some(*camera_position)) };
+
+    Ok(look_at_points
```
knoellle, line 295: Why is this function fallible?

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+            retries_exceeded = true;
+            None
+        }
+    });
+
+    (measurement, retries_exceeded || value_found)
+}
+
+// TODO Add fancier logic to either set this via parameters OR detect the location, walk, etc
+fn generate_look_at_list() -> Result<Vec<(Point2<Ground>, Option<CameraPosition>)>> {
```
knoellle, line 280: Why is the camera position optional?

knoellle (changes requested):

crates/types/src/parameters.rs
```diff
@@ -374,3 +374,18 @@ pub struct SearchSuggestorParameters {
...
+    pub max_retries: usize,
+}
+
+#[derive(
+    Clone, Debug, Default, Deserialize, Serialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub struct CameraCalibrationControllerParameters {
+    pub look_at_dispatch_wait_duration: Duration,
+    pub initial_to_calibration_stabilization_delay: Duration,
+}
```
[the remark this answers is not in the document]
tuxbotix: I was wondering about this today morning, the existence of the other types already got me confused. Will remove.

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+        dispatch_time: CycleTime,
+    },
+    PROCESS,
+    FINISH,
+}
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum CalibrationCaptureResponse<Measurement> {
```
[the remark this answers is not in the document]
tuxbotix: In this PR, but once #1076 is merged, there are two choices, and I felt having this generalisation for now makes it easy to swap out measurement type only in the vision and control module, not everywhere.

crates/calibration/src/lines.rs
```diff
@@ -1,12 +1,18 @@
...
-#[derive(Clone)]
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathIntrospect, PathDeserialize,
+)]
 pub struct Lines<Frame> {
     pub border_line: Line2<Frame>,
+    #[path_serde(leaf)]
     pub goal_box_line: Line2<Frame>,
+    #[path_serde(leaf)]
     pub connecting_line: Line2<Frame>,
```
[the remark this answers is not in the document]
tuxbotix: Pretty much my lack of clear understanding to make it work without leaf, will check again (I originally added these when serializehierarchy was there)

crates/control/src/calibration_controller.rs
tuxbotix: Correct, the issue I'm facing is that we might not have the right look-at list or such and not capture anything.
I'll change this to capture with specific camera (same as look-at camera). if it fails (retries exceeded), then we move on.

This means, if we don't have enough measurements, solver should abort and inform the user (i.e. audio.)

crates/control/src/calibration_controller.rs
tuxbotix: Current look-at behavious has it optional, for convenience, I left it optional in the context of look-at command.

Since the system works in this state, I'll make it mandatory and tighten the control.

crates/calibration/src/lines.rs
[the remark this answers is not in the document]
tuxbotix: Okay, I looked again. The issue is that, although I've added the PathIntrospect trait to `Line`, it is still complaining. When it was with `serializedhierarchy` this was due to `Line` being a Tuple struct.
`the trait bound `geometry::line::Line<Frame, 2>: PathIntrospect` is not satisfied`

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[default]
+    Idling,
+    CommandRecieved {
+        dispatch_time: CycleTime,
+        output: Option<Measurement>,
```
knoellle, line 37:
```suggestion
        measurement: Option<Measurement>,
```

knoellle (changes requested):

crates/calibration/src/lines.rs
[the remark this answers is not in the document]
knoellle: If I remove the leaf attributes, it builds fine on my machine and clippy doesn't complain either... 😅

crates/types/src/parameters.rs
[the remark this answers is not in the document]
tuxbotix: Fixed

crates/control/src/world_state_composer.rs
```diff
@@ -80,6 +82,7 @@ impl WorldStateComposer {
             instant_kick_decisions: context.instant_kick_decisions.cloned(),
             filtered_game_controller_state: context.filtered_game_controller_state.copied(),
             hypothetical_ball_positions: context.hypothetical_ball_position.clone(),
+            calibration_command: Some(context.calibration_command.clone()),
```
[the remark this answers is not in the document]
tuxbotix: Fixed, I initially looked at `KickDecision` and how it was optional, didn't have the enum for command when I originally wrote this part.

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{camera_position::CameraPosition, cycle_time::CycleTime};
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum CalibrationCommand {
+    #[default]
+    INACTIVE,
```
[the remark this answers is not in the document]
tuxbotix: Fixed

crates/control/src/calibration_controller.rs
tuxbotix: Honestly, I don't remember either, and looking at it again, it doesn't have to be.

crates/control/src/calibration_controller.rs
tuxbotix: Done

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+                None
+            }
+        }
+        CalibrationCaptureResponse::RetriesExceeded { dispatch_time: _ } => {
+            retries_exceeded = true;
+            None
+        }
+    });
+
+    (measurement, retries_exceeded || value_found)
```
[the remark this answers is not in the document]
tuxbotix: Ooh Didn't know about this syntax :D

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    fn get_next_look_at(&mut self) -> Option<(Point2<Ground>, Option<CameraPosition>)> {
+        let current_index = self.current_look_at_index;
+        self.current_look_at_index += 1;
+        if current_index < self.look_at_list.len() {
+            Some(self.look_at_list[current_index])
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
tuxbotix: *facepalm*

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+        .find(|&measurement| match measurement {
+            CalibrationCaptureResponse::CommandRecieved {
+                dispatch_time,
+                output: _,
+            } => original_dispatch_time.start_time == dispatch_time.start_time,
+            CalibrationCaptureResponse::Idling => false,
+            CalibrationCaptureResponse::RetriesExceeded { dispatch_time } => {
+                original_dispatch_time.start_time == dispatch_time.start_time
+            }
+        });
```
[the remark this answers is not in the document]
tuxbotix: Yes, I think this would be fine as there can only be one response from the vision cycler.

That said, I should reconsider using `filter_map` as this filtering needs a lot of chaining.

crates/control/src/calibration_controller.rs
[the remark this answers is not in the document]
tuxbotix: Fixed

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+                Some(CalibrationCommand::FINISH)
+            }
+            CalibrationCommand::FINISH => None,
+        }
+    }
+
+    fn try_transition_to_look_at(&mut self, dispatch_time: CycleTime) -> CalibrationCommand {
+        let next_look_at = self.get_next_look_at();
+
+        match next_look_at {
```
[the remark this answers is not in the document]
tuxbotix: Fixed

crates/vision/src/calibration_measurement_provider.rs
```diff
@@ -0,0 +1,147 @@
...
+    calibration_command: Input<CalibrationCommand, "control", "calibration_command">,
+    camera_position: Parameter<CameraPosition, "image_receiver.$cycler_instance.camera_position">,
+    field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
+    max_retries: Parameter<usize, "calibration_measurement_provider.max_retries">,
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct CalibrationMeasurementProvider {
+    last_capture_command_time_and_retries: Option<(SystemTime, usize)>,
+}
```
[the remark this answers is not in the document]
tuxbotix: Fixed, it is no longer optional.

crates/types/src/calibration.rs
tuxbotix: Done.

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+    FINISH,
+}
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum CalibrationCaptureResponse<Measurement> {
+    #[default]
+    Idling,
+    CommandRecieved {
```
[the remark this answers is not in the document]
tuxbotix: I moved the retry logic into the controller and simplified the vision cycler.

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+)]
+pub enum CalibrationCommand {
+    #[default]
+    INACTIVE,
+    INITIALIZE {
+        started_time: CycleTime,
+    },
+    LOOKAT {
+        target: Point2<Ground>,
+        camera: Option<CameraPosition>,
```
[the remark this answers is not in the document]
tuxbotix: Correct, fixed and made it mandatory

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+    PROCESS,
+    FINISH,
+}
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum CalibrationCaptureResponse<Measurement> {
+    #[default]
+    Idling,
```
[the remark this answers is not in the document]
tuxbotix: Fixed, not sure if I replied this already. Github is acting weird

crates/calibration/src/lines.rs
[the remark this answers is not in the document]
tuxbotix: Huh, strange! it works, but my LSP says not. Looks like I need to check for updates

crates/control/src/calibration_controller.rs
tuxbotix: Fixed, limited to specific camera

crates/control/src/calibration_controller.rs
tuxbotix: Fixed.

crates/control/src/calibration_controller.rs
tuxbotix: Ah yes,

crates/control/src/calibration_controller.rs
tuxbotix: Fixed

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,270 @@
...
+        .persistent
+        .iter()
+        .flat_map(|(_cycle_timestamp, measurements)| measurements.iter())
+        .flatten()
+        .find(|&measurement| {
+            let CalibrationCaptureResponse {
+                dispatch_time,
+                measurement: _,
+            } = measurement;
+            original_dispatch_time.start_time == dispatch_time.start_time
```
knoellle, line 249:
```suggestion
            original_dispatch_time.start_time == measurement.dispatch_time.start_time
```

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// TODO Add fancier logic to either set this via parameters OR detect the location, walk, etc
+fn generate_look_at_list() -> Vec<(Point2<Ground>, CameraPosition)> {
+    let look_at_points: Vec<Point2<Ground>> = vec![
+        point!(1.0, 0.0),
+        point!(1.0, -0.5),
+        point!(3.0, -0.5),
+        point!(3.0, 0.0),
+        point!(3.0, 0.5),
+        point!(1.0, -0.5),
```
knoellle, line 263: While technically all types of parentheses work for macros like this, please use `[]` to keep things consistent.

crates/control/src/calibration_controller.rs
tuxbotix: Uh that was a stupid one by me!  :3 fixed!

crates/control/src/calibration_controller.rs
tuxbotix: Oh yes, I am also not sure why I ended up doing it, might be some issue with my LSP that day :P. Fixed

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+            look_at_list: generate_look_at_list().unwrap(),
+            current_look_at_index: 0,
+            current_measurements: vec![],
+            current_primary_phase_is_calibration: false,
+            last_calibration_corrections: None,
+        })
+    }
+
+    pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
+        let calibration_grame_state_active =
```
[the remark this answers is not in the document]
tuxbotix: Oh no IDK how the heck this happened.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        let primary_state_transitioned_to_calibration =
+            calibration_grame_state_active && !self.current_primary_phase_is_calibration;
+        self.current_primary_phase_is_calibration = calibration_grame_state_active;
+
+        if primary_state_transitioned_to_calibration {
+            self.current_measurements = vec![];
+        }
+
+        let current_cycle_time = context.cycle_time;
```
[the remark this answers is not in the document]
tuxbotix: This is used to uniquely identify calibration commands of the same speciality (i.e. `Capture`). Helps to know if the receiver is still getting the same one or it has changed.
Also could be some random string or whatever.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+        let calibration_grame_state_active =
+            matches!(context.primary_state, PrimaryState::Calibration);
+        if !calibration_grame_state_active {
+            self.current_calibration_command = CalibrationCommand::INACTIVE;
+            return Ok(MainOutputs::default());
+        }
+
+        let primary_state_transitioned_to_calibration =
+            calibration_grame_state_active && !self.current_primary_phase_is_calibration;
+        self.current_primary_phase_is_calibration = calibration_grame_state_active;
```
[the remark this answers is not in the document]
tuxbotix: Yup cleared that mess including above. I wrote the big match statement inside the `cycle()` and then split it out, also added the early return later. That's why there were redundant statements and states.

crates/types/src/calibration.rs
```diff
@@ -0,0 +1,42 @@
...
+use linear_algebra::Point2;
+use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
+use serde::{Deserialize, Serialize};
+
+use crate::{camera_position::CameraPosition, cycle_time::CycleTime};
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum CalibrationCommand {
```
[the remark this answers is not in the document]
tuxbotix: I agree with you, and have the same concern. This was simply repeated at the WIP state to make things easy to test at the beginning. And I needed to verify that these states were enough for the job.

Now, the question would be how to avoid repitition of defining `LookAt` and `Capture` in the internal state and output command. I'll think of some ideas and try to simplify this.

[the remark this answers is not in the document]
knoellle: I think that is a small amount of duplication to pay in exchange for a much better separation of states and commands.
Don't overthink it 😉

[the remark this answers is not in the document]
tuxbotix: I fixed this, but there are a couple of things I'm not sure in terms of style:
1. the internal state enum is defined in the controller's module (no point to have it in a different crate)
2. IDK whether to keep the definition it at end of the file, top of the file or just before the `impl ... ` for the structs (what I've done now)

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,285 @@
...
+}
+
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {
+    pub calibration_command: MainOutput<Option<CalibrationCommand>>,
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub enum InternalCalibrationState {
```
tuxbotix, line 65: @knoellle This is what I mentioned, is it okay to have it here or should I move it elsewhere in the file?

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,285 @@
...
+            .calibration_measurements
+            .fill_if_subscribed(|| self.current_measurements.clone());
+        context
+            .last_calibration_corrections
+            .fill_if_subscribed(|| self.last_calibration_corrections);
+
+        let command: Option<CalibrationCommand> = self.current_calibration_state.clone().into();
+        Ok(MainOutputs {
+            calibration_command: command.into(),
+        })
```
tuxbotix, line 140: I don't like this clone().into() and .into() again, any ideas what can be done better?

crates/control/src/calibration_controller.rs
knoellle: Its fine here, but you can remove the `pub` and the "internal" prefix.

crates/control/src/calibration_controller.rs
knoellle: Replace the From implementation with a function that takes &self.
From moves the subject, so that'll get rid of one of the `into`s.
Having a well named function instead of "into" should also make it more clear what is happening and you don't have to write the type here anymore.

The one `into` for the main output is normal.

tuxbotix: Done, probably not the best function name, but should convey the message.

crates/control/src/calibration_controller.rs
tuxbotix: Done, I realised I've unnecessarily made the fields public .

crates/control/src/calibration_controller.rs
tuxbotix: In principle we can and will simplify things at a glance. But, the function that gets the next available look-at is called in two places (initialise-> look-at), and (capture -> look-at). So it was a bit more obvious to see where the processing happens by having a seperate state.

The drawback is that the processing starts one cycle later.

crates/control/src/calibration_controller.rs
tuxbotix: Renamed and removed the bindings.

crates/control/src/behavior/calibrate.rs
```diff
@@ -1,12 +1,28 @@
...
 };
 
 pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
-    match world_state.robot.primary_state {
-        PrimaryState::Calibration => Some(MotionCommand::Stand { head: Unstiff }),
-        _ => None,
+    match (
+        world_state.robot.primary_state,
+        &world_state.calibration_command,
+    ) {
```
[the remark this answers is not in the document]
tuxbotix: I tried with let-else and flatten it, please let me know if this is alright :)

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+            .fill_if_subscribed(|| self.current_measurements.clone());
+        context
+            .last_calibration_corrections
+            .fill_if_subscribed(|| self.last_calibration_corrections);
+        Ok(MainOutputs {
+            calibration_command: self.current_calibration_command.clone().into(),
+        })
+    }
+
+    fn get_next_command(
```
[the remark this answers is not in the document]
tuxbotix: Done, I added copy trait to the internal enum to reduce extra cloning. Let me know if there are pitfalls in using copy for the datatypes in the enum.

crates/control/src/behavior/calibrate.rs
```diff
@@ -6,20 +6,17 @@ use types::{
...
-                _ => HeadMotion::Center,
-            };
-
-            Some(MotionCommand::Stand { head: head_motion })
-        }
-        _ => None,
-    }
+    let (PrimaryState::Calibration, Some(CalibrationCommand::LookAt { target, camera, .. })) = (
+        world_state.robot.primary_state,
+        world_state.calibration_command.clone(),
```
knoellle, line 11: You don't need to clone here if you make the `CalibrationCommand` `Clone`.

crates/control/src/behavior/calibrate.rs
```diff
@@ -6,20 +6,17 @@ use types::{
...
-                // TODO Add walk-to-penalty area/ centre circle during CalibrationCommand::INITIALIZE
-                _ => HeadMotion::Center,
-            };
-
-            Some(MotionCommand::Stand { head: head_motion })
-        }
-        _ => None,
-    }
+    let (PrimaryState::Calibration, Some(CalibrationCommand::LookAt { target, camera, .. })) = (
+        world_state.robot.primary_state,
```
knoellle, line 10: I would split the primary state check and the calibration command check.
You can also take the calibration command by reference

crates/control/src/calibration_controller.rs
knoellle:
```
    fn get_next_look_at_or_processing(
        &mut self,
        dispatch_time: CycleTime,
    ) -> Option<CalibrationState> {
        self.get_next_look_at()
            .map(|(target, camera)| CalibrationState::LookAt {
                camera,
                target,
                dispatch_time,
            })
    }
```

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,273 @@
...
+                    *context.field_dimensions,
+                );
+
+                self.last_calibration_corrections = Some(solved_result);
+                return CalibrationState::Finish;
+            }
+            CalibrationState::Finish => {}
+        }
+
+        return self.current_calibration_state;
```
knoellle, line 219: You should decide if this function should change the `self.current_calibration_state` or return the new state.
Currently you do both and then assign the return value to `self.current_calibration_state` again.

tuxbotix: I noticed the same today xD and changed as follows:
1. Move the big-match into `cycle`, 
2. but take the larger parts into own functions -> i.e. process capture, etc
3. While we eliminated `Process`, I didn't want to do that part in an anonymous function, so a new function called `calibrate` is done.

I still think this can be refactored better, but if we don't have any other serious issues, we could try to finalize this :)

P.S. In case you feel moving the `match` is too much, i can revert that commit.

crates/control/src/behavior/calibrate.rs
tuxbotix: Done

crates/control/src/behavior/calibrate.rs
tuxbotix: Done

crates/control/src/behavior/calibrate.rs
```diff
@@ -1,12 +1,24 @@
...
-    match world_state.robot.primary_state {
-        PrimaryState::Calibration => Some(MotionCommand::Stand { head: Unstiff }),
-        _ => None,
+    if PrimaryState::Calibration != world_state.robot.primary_state {
+        return None;
     }
+
+    let Some(CalibrationCommand::LookAt { target, camera, .. }) = world_state.calibration_command
+    else {
+        return None;
```
knoellle, line 15: This currently causes a panic because there is no other action that feels responsible to produce a motion command in calibration mode.

tuxbotix: Oh I know what happened, there were 3 actions in the old version,

* If not calibration -> nothing 
* If calibration and look at -> look at head motion 
* If calibration but not look at -> head motion center

So the last condition has to be split to ternary expression style with restructuring to handle the last two cases.

knoellle: In the third case, the robot would already start to look away from the target while vision tries to capture measurements, wouldn't it?

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
+        if *context.primary_state != PrimaryState::Calibration {
+            self.current_calibration_state = CalibrationState::Inactive;
+            return Ok(MainOutputs::default());
+        }
+
+        match self.current_calibration_state {
+            CalibrationState::Inactive => {
+                self.current_measurements = vec![];
```
knoellle, line 125: You should also reset the current_look_at_index here.
Currently, when trying to calibrate twice, the measurements are cleared, but since there are no more targets left, you immediately transition from Initialize to Finish and end up with no measurements at all.

crates/control/src/calibration_controller.rs
```diff
@@ -0,0 +1,277 @@
...
+    cycle_time::CycleTime,
+    field_dimensions::FieldDimensions,
+    primary_state::PrimaryState,
+};
+
+#[derive(Deserialize, Serialize)]
+pub struct CalibrationController {
+    current_look_at_index: usize,
+    current_calibration_state: CalibrationState,
+    current_measurements: Vec<Measurement>,
```
knoellle, line 23: I'd remove the prefix of these fields.
They represent the state and there is no need to differentiate them from past or futures states.

crates/control/src/calibration_controller.rs
tuxbotix: Oh damn! Good catch! Should define a reset function or a seperate struct with the mutable parts so re-start means just create it freshly

knoellle: Having a separate struct that can just be reassigned is probably the cleanest approach, yes.

crates/control/src/behavior/calibrate.rs
tuxbotix: Hmm that's also right.
I'll re check the older working state and make the nessesary changes. Unstiff or no HeadMotion might be better.

knoellle: Neither of those sound like a good idea, you want neither head nor body to move during the capture.
That's the whole point of waiting in the LookAt state, isn't it?

knoellle:
Apart from walking to positions on the field, this PR also seems to largely fix #127.

crates/control/src/behavior/calibrate.rs
tuxbotix: Oh damn, I had forgotten it! One option is to go with "Unstiff" -> not reliable, head may or may not move.

Or Combine look at and capture
```rust
enum State{

LookAtAndCapture {
  dispatch_time: ...,
  target: point![1.0, 0.0],
  camera: CameraPostion:Top,
  trigger: true // Controller flips this On when the timeout is reached. used by the vision cycler to commence capturing
}
```

crates/control/src/calibration_controller.rs
tuxbotix: Yeah, makes more sense. Fixed.

crates/control/src/calibration_controller.rs
tuxbotix: Did it, looks a little odd, but I think it isn't too bad:
```rust

#[derive(Clone, Debug, Default, Serialize, Deserialize)]
struct StateTracking {
    look_at_index: usize,
    calibration_state: CalibrationState,
    measurements: Vec<Measurement>,
    last_capture_retries: u32,
}
```

crates/control/src/behavior/calibrate.rs
tuxbotix: *or*, I tried this, the command is now called `CalibrationCaptureCommand`, IDK if this is the best way to do it but simplify the consumer's logic.

As for the internal state, I kept the seperate states as I felt it was more readable in that context.

crates/types/src/calibration.rs
```diff
@@ -9,16 +9,11 @@ use crate::{camera_position::CameraPosition, cycle_time::CycleTime};
...
-    LookAt {
-        target: Point2<Ground>,
-        camera: CameraPosition,
-        dispatch_time: CycleTime,
-    },
-    Capture {
-        camera: CameraPosition,
-        dispatch_time: CycleTime,
-    },
+pub struct CalibrationCaptureCommand {
```
knoellle, line 12: Is it really a "capture" command?
I think the old name was better.

tuxbotix: Makes sense :D reverted it.

knoellle (approved):
So, nu aber 😄
LGTM

PR #801, opened 2024-04-07
Add jump to penalty kick

### LINKED ISSUE #250: Penalty Box Violations
Some robots violated the Penalty box during the penalty kick while attacking team was opponents. In the behavior simulator with all 7 players this is not the case.
This problem should be investigated in:
* Real life with all 7 robots active
* In the behavior simulator with no goalkeeper and one defender penalized to match test game situtation

### LINKED ISSUE #788: Jumping Keeper for Penalty Kick
Enable the ``Action::PrepareJump`` and ``Action::Jump`` for the keeper during a penalty kick

MaikRe:
## Introduced Changes

Allow keeper to jump during a penalty kick.
Prevent role changes during penalty kick so that keeper does not become striker and leave the goal.

Fixes #788 
Fixes #250 

## ToDo / Known Issues

Does not apply to replacement keepers. However, as these are not allowed to touch the ball with their hands anyway this is probably the desired behavior.

## How to Test

Test in game penalty kick and check if keeper jumps

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -9,7 +9,7 @@ use framework::{MainOutput, PerceptionInput};
 use hardware::NetworkInterface;
 use linear_algebra::{Isometry2, Point2, Vector};
 use spl_network_messages::{
-    GameControllerReturnMessage, GamePhase, HulkMessage, Penalty, PlayerNumber, Team,
+    GameControllerReturnMessage, GamePhase, HulkMessage, Penalty, PlayerNumber, SubState, Team,
 };
 use types::{
     ball_position::BallPosition,
@@ -355,6 +355,9 @@ fn process_role_state_machine(
                 kicking_team: Team::Opponent,
             } => return (Role::Keeper, false, None),
             _ => {}
+        };
+        if let Some(SubState::PenaltyKick) = game_controller_state.sub_state {
+            return (current_role, false, None);
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/control/src/role_assignment.rs
```diff
@@ -355,6 +355,9 @@ fn process_role_state_machine(
                 kicking_team: Team::Opponent,
             } => return (Role::Keeper, false, None),
             _ => {}
+        };
+        if let Some(SubState::PenaltyKick) = game_controller_state.sub_state {
+            return (current_role, false, None);
```
schluis, line 360: How does this solve the issue? Did the keeper switch his role during our test? Did you test it again?

schluis (changes requested):

crates/control/src/role_assignment.rs
MaikRe: I checked with the replayer and yes the role changed. I am currently trying to test this in the behavior sim but #666 broke all of the files except for golden goal and demonstration so now I'm down the rabbit hole of fixing all of these files as well. Will updated when I have more info

MaikRe: #811 and #812 allow the behavior simulator to now show that the keeper does in fact prepare to jump. Since the behavior simulator does not have a penalty shot direction the action jump is never carried out so this needs to be tested in the lab. However yes this is the correct fix

MaikRe: On second thought, yes this allows the keeper to prepare jump, but he won't actually jump because the penaltyshotdirection is restricted to the gamephase penalty shootout. I will implement this as well

MaikRe: Just tested this, works as expected now!

schluis (approved):
Dange Chef

PR #809, opened 2024-04-07
Realtime priority for realtime cyclers

oleflb:
## Introduced Changes

Makes cyclers with realtime kind also use a realtime scheduler with priority 20. This should (?) fix issues of other cyclers stealing ressources from Control.

Fixes #

## ToDo / Known Issues
- We need the capability `sys_cap_nice` on the `hulk` binary before merging this PR. Otherwise we can only launch with sudo

## How to Test
* Check if process starts up and robot behaves normally

### DIFF
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -200,7 +200,7 @@ fn generate_node_fields(cycler: &Cycler) -> TokenStream {
 fn generate_implementation(cycler: &Cycler, cyclers: &Cyclers, mode: Execution) -> TokenStream {
     let new_method = generate_new_method(cycler, cyclers, mode);
     let start_method = match mode {
-        Execution::None | Execution::Run => generate_start_method(),
+        Execution::None | Execution::Run => generate_start_method(cycler.kind),
         Execution::Replay => Default::default(),
     };
     let cycle_method = generate_cycle_method(cycler, cyclers, mode);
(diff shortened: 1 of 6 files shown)


crates/code_generation/src/cyclers.rs
```diff
@@ -379,7 +379,27 @@ fn generate_consumer_identifiers(cyclers: &Cyclers) -> Vec<Ident> {
         .collect()
 }
 
-fn generate_start_method() -> TokenStream {
+fn generate_start_method(cycler_kind: CyclerKind) -> TokenStream {
+    let scheduler_tokens = match cycler_kind {
+        CyclerKind::Perception => quote! {
+            unsafe {
+                let priority = libc::sched_param {
+                    sched_priority: 10,
```
h3ndrk, line 387: This is 10 but the PR description mentions 20?

crates/framework/Cargo.toml
h3ndrk, line 1: Sort alphabetically and add final newline according to POSIX standard?

h3ndrk (changes requested):

crates/framework/Cargo.toml
oleflb: done

crates/code_generation/src/cyclers.rs
oleflb: hoppla, meant to write 10 in the description, that way it has lower priority compared to hula

h3ndrk (commented):
Code schaut gut zu mir :handshake: We're testing tomorrow

h3ndrk (approved):

PR #811, opened 2024-04-08
Fix behavior simulator files

MaikRe:
## Introduced Changes

Use FilteredGameControllerState in behavior simulator and throw out filtered game state and game controller state to match the rest of the repo.

Adapts behavior files to match

Fixes all of the behavior files broken by #666

## How to Test

Test behavior files.

### DIFF
--- a/tools/behavior_simulator/src/state.rs
+++ b/tools/behavior_simulator/src/state.rs
@@ -12,16 +12,13 @@ use coordinate_systems::{Field, Head};
 use geometry::line_segment::LineSegment;
 use linear_algebra::{vector, Isometry2, Orientation2, Point2, Rotation2, Vector2};
 use serialize_hierarchy::SerializeHierarchy;
-use spl_network_messages::{GamePhase, GameState, HulkMessage, PlayerNumber, Team};
+use spl_network_messages::{GamePhase, HulkMessage, PlayerNumber, Team};
 use types::{
     ball_position::BallPosition,
     filtered_game_controller_state::FilteredGameControllerState,
     filtered_game_state::FilteredGameState,
-    game_controller_state::GameControllerState,
     messages::{IncomingMessage, OutgoingMessage},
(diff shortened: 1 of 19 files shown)


MaikRe:
@knoellle I am having the issue that I technically have to do this
```lua
    if state.cycle_count == 1600 then
        state.filtered_game_controller_state.game_state = {
            Set
        }
    end
```
The same way that I do
```lua    
if state.cycle_count == 100 then
        state.filtered_game_controller_state.game_state = {
            Ready = {
                kicking_team = "Hulks",
            }
        }
    end
```
The bottom one serializes properly but the top one doesn't. Neither does "Set" or Set ={} or whatever. Do you have any suggestions of how to fix this?

MaikRe:
NVM I found my issue

tools/behavior_simulator/src/state.rs
```diff
@@ -326,9 +317,9 @@ impl Default for State {
                 six: None,
                 seven: None,
             },
-            remaining_amount_of_messages: 1200,
+            remaining_number_of_messages: 1200,
             sub_state: None,
-            hulks_team_is_home_after_coin_toss: false,
```
schluis, line 331: Nice change, please also rename all other instances of this in the repo in the files
- `game_controller_state.rs`
- `game_controller_state_message.rs`
- `game_controller_state_filter.rs`
- `game_controller_filter.rs`

schluis (changes requested):
Thanks! 🚀

tools/behavior_simulator/src/state.rs
MaikRe: This renaming was part of #666 and I just used the naming from that. I think since my focus here is the behavior simulator that change should be a different PR. What do you think?

schluis: Yep okay 👍🏻

schluis (approved):

PR #812, opened 2024-04-08
Increase penalty box size to prevent keeper getting stuck

MaikRe:
## Introduced Changes

Increase the penalty box size by 20cm and shift accordingly
The keeper was getting stuck on the boundary or thinking it was outside the penalty box and would not defend a penalty kick properly.

Fixes #

## How to Test

Optimally merge #811 first.
It can be seen that the keeper now sticks to the line when defending the penalty kick rather than staying behind it and failing to defend properly.

### DIFF
--- a/crates/control/src/rule_obstacle_composer.rs
+++ b/crates/control/src/rule_obstacle_composer.rs
@@ -23,9 +23,11 @@ pub struct CycleContext {
         RequiredInput<Option<FilteredGameControllerState>, "filtered_game_controller_state?">,
     ball_state: Input<Option<BallState>, "ball_state?">,
 
-    center_circle_obstacle_increase: Parameter<f32, "center_circle_obstacle_increase">,
+    center_circle_obstacle_increase:
+        Parameter<f32, "rule_obstacles.center_circle_obstacle_increase">,
     field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
-    free_kick_obstacle_radius: Parameter<f32, "free_kick_obstacle_radius">,
+    free_kick_obstacle_radius: Parameter<f32, "rule_obstacles.free_kick_obstacle_radius">,
+    penaltykick_box_increase: Parameter<f32, "rule_obstacles.penaltykick_box_increase">,
 }
 
 #[context]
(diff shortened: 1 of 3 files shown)


crates/control/src/rule_obstacle_composer.rs
```diff
@@ -122,11 +122,11 @@ pub fn create_penalty_box(field_dimensions: &FieldDimensions, kicking_team: Team
     };
     let half_field_length = field_dimensions.length / 2.0;
     let half_penalty_area_length = field_dimensions.penalty_area_length / 2.0;
-    let center_x = side_factor * (half_field_length - half_penalty_area_length);
+    let center_x = side_factor * (half_field_length - half_penalty_area_length + 0.1);
     RuleObstacle::Rectangle(Rectangle::new_with_center_and_size(
         point![center_x, 0.0],
         vector![
-            field_dimensions.penalty_area_length,
+            field_dimensions.penalty_area_length + 0.2,
```
schluis, line 129: Please convert these to modifiable parameters ;)

schluis (changes requested):

crates/control/src/rule_obstacle_composer.rs
MaikRe: Did so, and I grouped the rule obstacle relevant parameters into one part of the ```default.json```. Do you think I should make a struct to group these further?

schluis: Nice, is fine for me like this 👍🏻

etc/parameters/default.json
```diff
@@ -1160,6 +1160,9 @@
     }
   },
   "angular_velocity_smoothing_factor": 0.1,
-  "center_circle_obstacle_increase": 1.2,
-  "free_kick_obstacle_radius": 0.75
+  "rule_obstacles": {
+    "center_circle_obstacle_increase": 1.2,
+    "free_kick_obstacle_radius": 0.75,
+    "penaltykick_box_increase": 0.2
```
schluis, line 1166: Please make both `increase`s consistent, the `center_circle_obstacle_increase` is multiplicative and the new `penaltykick_box_increase` is additive.

MaikRe: Renamed

schluis (approved):

schluis:
Nice, thanks! 👍🏻

PR #813, opened 2024-04-09
Add tests for each behavior file

MaikRe:
## Introduced Changes

Add all behavior test files to be tested in CI

Fixes #

## How to Test

If the CI is green it means I did it right


crates/code_generation/src/write_to_file.rs
```diff
@@ -40,3 +40,20 @@ impl WriteToFile for TokenStream {
         Ok(())
     }
 }
+
+impl WriteToFile for String {
```
MaikRe, line 44: This is basically just code duplication of impl for TokenStream. How can I reuse that function?

knoellle (approved):
Merg

PR #814, opened 2024-04-10
Behavior in CI autogen

MaikRe:
## Introduced Changes

Auto generate CI tests for behavior sim files
Correct issues in one behavior file from #488 

Improves #813 

## ToDo / Known Issues

~This is sort of a hack because the formatting for the test code is all done through string. I will gladly take feedback on how to make this prettier. Unfortunately the quote macro is not compatible with newlines and rustfmt doesn't like having #[test] on the same line as the function definition so this is the only reasonable solution I found.~

## How to Test

CI do be green


knoellle:
I'm not sure what you mean by rustfmt being unhappy with `#[test]`.

This seems to compile, generate, format, compile again, and run fine for me:
```rs
    let paths: TokenStream = read_dir("../../tests/behavior")
        .expect("")
        .map(|file| {
            let path = file.unwrap().path();
            let name = path.file_stem().unwrap().to_string_lossy().to_owned();
            let test_file = path.file_name().unwrap().to_string_lossy().into_owned();
            let path = format!("../../tests/behavior/{test_file}");

            let function_name = format_ident!("test_{}", name);

            quote! {
                #[test]
                fn #function_name() -> color_eyre::Result<()> {
                    test_scenario(#path)
                }
            }
        })
        .collect();
    paths
        .write_to_file("behavior_files.rs")
        .wrap_err("failed to write generated tests to file")?;
```

MaikRe:
I had tried this in a weird multi level way like how some of our other code generation seems to work and was probably nesting something wrong. Your recommendation works as expected!

crates/code_generation/src/write_to_file.rs
```diff
@@ -40,3 +40,20 @@ impl WriteToFile for TokenStream {
...
+            write!(file, "{self}")?;
+        }
+
+        let status = Command::new("rustfmt").arg(file_path).status()?;
+        if !status.success() {
+            return Err(Error::RustFmt);
+        }
+        Ok(())
+    }
+}
```
knoellle, line 59: Can this be removed now?

MaikRe: Yes

knoellle (approved):
LGTM

PR #820, opened 2024-04-12
deprecate meta-hulks in favor of meta-nao

schmidma:
the ci image action should query for releases from meta-nao


knoellle (approved):
Schaut mich gut an

PR #822, opened 2024-04-13
Allow transition from finished to initial

### LINKED ISSUE #815: Add state transition from finished to initial
As discussed during our test game today.

Currently, if we deploy to our robots while the game controller is still in finished (from a previous game/half), we can never leave finished and have to deploy again.

Narcha:
## Introduced Changes

what the title says

Fixes #815 

## How to Test

Put a nao on the field, and play around with the game controller.
Things to try:
- undo finishes
  - this sets the game controller state to `Playing`, not `Initial`. The nao stays in `Finish`.
- connect a nao while the game controller is currently in `Finish`, then start a new game


MaikRe (approved):
Works as intended!

PR #825, opened 2024-04-14
Stay in Calibration

### LINKED ISSUE #824: Transition to PrimaryState::Calibration with game controller running
When the game controller is running, the transition from `PrimaryState::Initial` to `PrimaryState::Calibration` does not work properly. The transition happens but is immediately followed by a transition back to whatever the game controller state is.

MaikRe:
## Introduced Changes

One liner. When in calibration stay in calibration as per SPL rule state transitions.

Fixes #824 

## How to Test

Enter calibration while a game controller is running and the NAO is connected to the wireless network of that game controller. Before this change the NAO would immediately go back into initial and now stays in calibration.

### DIFF
--- a/crates/control/src/primary_state_filter.rs
+++ b/crates/control/src/primary_state_filter.rs
@@ -72,6 +72,8 @@ impl PrimaryStateFilter {
                 PrimaryState::Unstiff
             }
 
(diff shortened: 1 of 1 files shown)


oleflb (approved):
Abfahrt :steam_locomotive:

Issue #818, opened 2024-04-10
Require multiple successive referee pose detections


ThagonDuarte:
Require that at least a certain amount of successive referee pose detections have to occur, before commiting to a detection and sending a message.

oleflb:
Implemented by #831

PR #831, opened 2024-04-14
Visual referee message transmission decision

Vivituhh:
## Introduced Changes
Introduces a queue to reduce the number of messages sent in `Initial` and reduce false positives.
At the moment the queue has a length of 4 and if there are 3 detections out of those 4 the robot changes to `Ready`. Please suggest other values if this is not sufficient.
A detected pose can be seen by two purple leds in the left eye of the NAO. 

Depends on #806 

Addresses #818

## ToDo / Known Issues

## How to Test
Test with NAO if it takes longer as in #806 to change from `Initial` -> `Ready` with a running game controller. 
And can use `Control.additional_outputs.referee_pose_queue` in twix to see the queue.

### DIFF
--- a/crates/control/src/behavior/initial.rs
+++ b/crates/control/src/behavior/initial.rs
@@ -1,5 +1,6 @@
 use coordinate_systems::{Field, Ground};
 use linear_algebra::{Isometry2, Point2};
+use spl_network_messages::PlayerNumber;
 use types::{
     camera_position::CameraPosition,
     filtered_game_controller_state::FilteredGameControllerState,
@@ -22,6 +23,7 @@ pub fn execute(
             world_state.robot.ground_to_field,
             world_state.filtered_game_controller_state,
             expected_referee_position,
+            world_state.clone(),
         )
         .unwrap_or(MotionCommand::Initial {
             head: HeadMotion::Center,
@@ -34,6 +36,7 @@ fn look_at_referee(
     ground_to_field: Option<Isometry2<Ground, Field>>,
     filtered_game_controller_state: Option<FilteredGameControllerState>,
     expected_referee_position: Option<Point2<Field>>,
+    world_state: WorldState,
 ) -> Option<MotionCommand> {
     let ground_to_field = ground_to_field?;
     let expected_referee_position = expected_referee_position?;
@@ -46,6 +49,19 @@ fn look_at_referee(
     if position.y().signum() == expected_referee_position.y().signum() {
         return None;
     };
+
+    let own_team_is_home_after_coin_toss = world_state
+        .filtered_game_controller_state?
+        .own_team_is_home_after_coin_toss;
+
+    match (
+        world_state.robot.player_number,
+        own_team_is_home_after_coin_toss,
+    ) {
+        (PlayerNumber::Four | PlayerNumber::Seven, true) => {}
+        (PlayerNumber::Two | PlayerNumber::Six, false) => {}
+        _ => return None,
+    }
 
     Some(MotionCommand::Initial {
         head: HeadMotion::LookAt {
--- a/crates/control/src/behavior/initial.rs
+++ b/crates/control/src/behavior/initial.rs
@@ -1,5 +1,5 @@
-use coordinate_systems::{Field, Ground};
-use linear_algebra::{Isometry2, Point2};
+use coordinate_systems::Field;
+use linear_algebra::Point2;
 use spl_network_messages::PlayerNumber;
 use types::{
     camera_position::CameraPosition,
@@ -18,24 +18,20 @@ pub fn execute(
     }
 
     Some(
-        look_at_referee(
-            world_state.robot.ground_to_field,
-            expected_referee_position,
-            world_state.clone(),
-        )
-        .unwrap_or(MotionCommand::Initial {
-            head: HeadMotion::Center,
-            should_look_for_referee: false,
-        }),
+        look_at_referee(expected_referee_position, world_state.clone()).unwrap_or(
+            MotionCommand::Initial {
+                head: HeadMotion::Center,
+                should_look_for_referee: false,
+            },
+        ),
     )
 }
 
 fn look_at_referee(
-    ground_to_field: Option<Isometry2<Ground, Field>>,
     expected_referee_position: Option<Point2<Field>>,
     world_state: WorldState,
 ) -> Option<MotionCommand> {
-    let ground_to_field = ground_to_field?;
+    let ground_to_field = world_state.robot.ground_to_field?;
     let expected_referee_position = expected_referee_position?;
     if world_state.filtered_game_controller_state?.game_state != FilteredGameState::Initial {
         return None;
--- a/crates/control/src/behavior/initial.rs
+++ b/crates/control/src/behavior/initial.rs
@@ -3,7 +3,6 @@ use linear_algebra::{Isometry2, Point2};
 use spl_network_messages::PlayerNumber;
 use types::{
     camera_position::CameraPosition,
-    filtered_game_controller_state::FilteredGameControllerState,
     filtered_game_state::FilteredGameState,
     motion_command::{HeadMotion, ImageRegion, MotionCommand},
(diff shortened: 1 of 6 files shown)


crates/object_detection/src/pose_interpretation.rs
```diff
@@ -84,7 +93,19 @@ impl PoseInterpretation {
...
+            .push_front(matches!(pose_type, PoseType::OverheadArms));
+
+        self.detected_over_arm_poses_queue
+            .truncate(*context.referee_pose_queue_length);
+        let detected_referee_position_count = self
+            .detected_over_arm_poses_queue
+            .iter()
+            .filter(|x| **x)
+            .count();
+        dbg!(detected_referee_position_count);
```
schluis, line 106: Forgotten dbg

schluis (changes requested):
Looks good, nice 👍🏻

crates/object_detection/src/pose_interpretation.rs
Vivituhh: This is resolved

oleflb:
@schluis do you review this?

schluis:
Jep, can do that

crates/control/src/referee_pose_detection_filter.rs
```diff
@@ -1,32 +1,30 @@
...
-    messages::{IncomingMessage, OutgoingMessage},
-    players::Players,
+    cycle_time::CycleTime, fall_state::FallState, messages::IncomingMessage, players::Players,
     pose_kinds::PoseKind,
 };
 
 #[derive(Deserialize, Serialize)]
 pub struct RefereePoseDetectionFilter {
     detection_times: Players<Option<SystemTime>>,
+    own_detected_above_arm_poses_queue: VecDeque<bool>,
```
schluis, line 20: Why `own_...`? Can that be omitted?

crates/control/src/referee_pose_detection_filter.rs
```diff
@@ -48,18 +46,31 @@ pub struct CycleContext {
...
+    >,
+
+    referee_pose_queue: AdditionalOutput<VecDeque<bool>, "referee_pose_queue">,
 }
 
 #[context]
 #[derive(Default)]
 pub struct MainOutputs {
     pub is_referee_initial_pose_detected: MainOutput<bool>,
+    pub is_own_referee_initial_pose_detected: MainOutput<bool>,
```
schluis, line 64: Renaming suggestion:
```
pub filtered_is_referee_initial_pose_detected: MainOutput<bool>,
pub is_referee_initial_pose_detected: MainOutput<bool>,
```

crates/control/src/referee_pose_detection_filter.rs
```diff
@@ -82,37 +93,40 @@ impl RefereePoseDetectionFilter {
...
         Ok(MainOutputs {
             is_referee_initial_pose_detected: is_referee_initial_pose_detected.into(),
+            is_own_referee_initial_pose_detected: is_own_referee_initial_pose_detected.into(),
         })
     }
 
-    fn update(&mut self, context: &CycleContext<impl NetworkInterface>) -> Result<()> {
-        let time_tagged_persistent_messages =
-            unpack_message_tree(&context.network_message.persistent);
+    fn update(&mut self, context: &CycleContext<impl NetworkInterface>) -> Result<bool> {
```
schluis, line 106: You don't need the `Result`, just return `bool`

schluis (changes requested):
Very nice, thank you! 😊

crates/control/src/behavior/initial.rs
```diff
@@ -47,6 +50,19 @@ fn look_at_referee(
         return None;
     };
 
+    let own_team_is_home_after_coin_toss = world_state
+        .filtered_game_controller_state?
+        .own_team_is_home_after_coin_toss;
+
+    match (
+        world_state.robot.player_number,
```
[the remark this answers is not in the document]
Vivituhh: The idea (which was actually not implemented in this PR) was that the positions on which sides the robots start in `Initial` are given by the player number and not the role, if I'm not mistaken. Because we want to see the referee on the other side of the field and not next to us, and also only the inner most two robots should look there.

crates/control/src/referee_pose_detection_filter.rs
Vivituhh: Done

crates/control/src/referee_pose_detection_filter.rs
Vivituhh: In this case I would argue that we lose the discriminability.

crates/control/src/referee_pose_detection_filter.rs
Vivituhh: This was because we distinguish between own detections of the referee and the detections of the other robots (received via a message). I think this can be omitted.

Done

crates/control/src/led_status.rs
```diff
@@ -293,17 +305,20 @@ impl LedStatus {
...
-                            .unwrap_or_else(|| ball_background_color.unwrap_or(Rgb::BLACK)),
+                        color_at_0: referee_color.unwrap_or(Rgb::BLACK),
                         color_at_45: ball_color_top
                             .unwrap_or_else(|| ball_background_color.unwrap_or(Rgb::BLACK)),
                         color_at_90: ball_background_color.unwrap_or(Rgb::BLACK),
                         color_at_135: ball_color_bottom
                             .unwrap_or_else(|| ball_background_color.unwrap_or(Rgb::BLACK)),
-                        color_at_180: ball_color_bottom
-                            .unwrap_or_else(|| ball_background_color.unwrap_or(Rgb::BLACK)),
+                        color_at_180: referee_color.unwrap_or(Rgb::BLACK),
```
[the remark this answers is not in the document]
Vivituhh: Done

crates/control/src/led_status.rs
```diff
@@ -47,6 +48,13 @@ pub struct MainOutputs {
     pub leds: MainOutput<Leds>,
 }
 
+struct BallData {
+    at_least_one_ball_data_top: bool,
+    at_least_one_ball_data_bottom: bool,
+    last_ball_data_top_too_old: bool,
+    last_ball_data_bottom_too_old: bool,
+}
```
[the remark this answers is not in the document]
Vivituhh: So you suggest to always throw away the `data` in all the `ball_data` things? Even for the `SystemTime`s? I now only cleaned up the struct.

crates/control/src/behavior/initial.rs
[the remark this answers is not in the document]
schluis: Okay, then let's leave it like this

crates/control/src/led_status.rs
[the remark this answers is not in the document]
schluis: Jep, I think this would be also nice for the `LedStatus` :)

crates/control/src/led_status.rs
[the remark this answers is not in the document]
schluis: Nice 🚀

crates/control/src/referee_pose_detection_filter.rs
schluis: How about:

```
pub majority_vote_is_referee_initial_pose_detected: MainOutput<bool>,
pub is_referee_initial_pose_detected: MainOutput<bool>,
```

crates/control/src/led_status.rs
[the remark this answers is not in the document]
Vivituhh: I changed that. What do you think about `BallSeen` -> `BallPercepts`. Sounds better for me

crates/control/src/referee_pose_detection_filter.rs
Vivituhh: Done

crates/control/src/led_status.rs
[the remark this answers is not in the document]
schluis: Jup, nice, thanks 🚀

schluis (approved):

PR #840, opened 2024-04-15
GC Collision Alert

### LINKED ISSUE #257: Alert when multiple game controllers in network
We should play an audio warning when there is more than one game controller in the network.
The playback should be triggered if there was more than one game controller IP change within a certain time period like 10s.
Perhaps with backoff to reduce annoyance while figuring out where the game controllers are.

knoellle:
## Introduced Changes

Waiting for #806 to prevent potential painful rebases.

Add audio alert when there are multiple active game controllers in the network.
Also adds an AdditionalOutput which contains all observed game controller IPs and the time when they were last seen.

Fixes #257 

## ToDo / Known Issues

- [x] Pass address through robotics domain instead of keeping it inside the endpoint type.

## How to Test

- Start two game controllers
- Hey, listen!


crates/types/src/messages.rs
```diff
@@ -1,12 +1,15 @@
...
+
 use serialize_hierarchy::SerializeHierarchy;
 use spl_network_messages::{
     GameControllerReturnMessage, GameControllerStateMessage, HulkMessage, VisualRefereeMessage,
 };
 
 #[derive(Clone, Debug, Deserialize, Serialize, SerializeHierarchy)]
 pub enum IncomingMessage {
-    GameController(GameControllerStateMessage),
+    GameController(SocketAddr, GameControllerStateMessage),
```
h3ndrk, line 12: Hmmm, my rule of thumb is that I try to use tuple structs only if I have a single wrapped type. You have two. Is it already unclear what the address means s.t. we should introduce proper names for the fields? On the other hand, it would be inconsistent with the remaining enum variants... 🤔

crates/control/src/game_controller_filter.rs
```diff
@@ -71,9 +89,47 @@ impl GameControllerFilter {
                 hulks_team_is_home_after_coin_toss: game_controller_state_message
                     .hulks_team_is_home_after_coin_toss,
             });
+
+            self.last_contact.insert(*source_address, time);
+            let on_cooldown = self
```
h3ndrk, line 94: Naming: Maybe "has cooled down" or also name the WHAT like "last audio message has cooled down" to make it even more clear.

crates/control/src/game_controller_filter.rs
```diff
@@ -71,9 +89,47 @@ impl GameControllerFilter {
...
+                .last_collision_warning
+                .is_some_and(|last_collision_warning| {
+                    time.duration_since(last_collision_warning)
+                        .expect("time ran backwards")
+                        .as_secs_f32()
+                        < 10.0
```
h3ndrk, line 100: Add a Paramalam Ding Dong? https://youtu.be/0HhA0Cghr4k

Same for the second constant below.

crates/control/src/game_controller_filter.rs
```diff
@@ -71,9 +89,47 @@ impl GameControllerFilter {
...
+                });
+            let recent_contacts = self.last_contact.iter().filter(|(_address, last_contact)| {
+                time.duration_since(**last_contact)
+                    .expect("time ran backwards")
+                    .as_secs_f32()
+                    < 5.0
+            });
+            let collisions = recent_contacts.count() > 1;
```
h3ndrk, line 108: Naming: Add a verb. Maybe "at least two game controllers collided" or something else. Gippedy to your success 💪

h3ndrk (changes requested):
Booleans should have verbs

crates/control/src/game_controller_filter.rs
knoellle: I renamed this variable.

crates/types/src/messages.rs
knoellle: I think its fine the way it is.

crates/control/src/game_controller_filter.rs
knoellle: Döne

crates/control/src/game_controller_filter.rs
knoellle: I renamed this variable.

crates/control/src/game_controller_filter.rs
h3ndrk: Nice choice of paramalam name 😂

h3ndrk (dismissed):
Approving but disabling auto merge s.t. Visual Referee may merge first (due to probable conflicts) and a discussion about testing.

h3ndrk (approved):

PR #841, opened 2024-04-16
Replay recording frequency

julianschuler:
## Introduced Changes

This pull request adds allowing to set recording frequencies per cycler.

## ToDo / Known Issues

None.

Consider making `recording_settings` a generated struct instead of the untyped HashMap.

## How to Test

Use `pepsi recording Control=1,VisionTop=30,VisionBottom` followed by `pepsi upload` or `pepsi pregame`, replay the files and observer that the vision cyclers are recorded at a lower frequency (1 per second).


crates/code_generation/src/cyclers.rs
```diff
@@ -455,7 +457,16 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers, mode: Execution) ->
...
         Execution::Run => quote! {
-            let enable_recording = self.enable_recording && self.hardware_interface.should_record();
+            let enable_recording = self.cycles_per_recording != 0 && self.cycles_to_next_recording == 0;
+            if self.cycles_per_recording != 0 {
+                if self.cycles_to_next_recording == 0 {
+                    self.cycles_to_next_recording = self.cycles_per_recording - 1;
+                }
+                else {
+                    self.cycles_to_next_recording -= 1;
+                }
```
oleflb, line 467:
```suggestion
self.cycles_to_next_recording = (self.cycles_to_next_recording + 1) % self.cycles_per_recording
```
Saves one indentation :)

oleflb: Oh and requires renaming of `cycles_to_next_recording` to sth. like `cycles_since_last_recording`

oleflb (commented):
Tested and seems to work as intended :)

crates/code_generation/src/cyclers.rs
julianschuler: I like the idea, but this would also require changing 

```rust
let enable_recording = self.cycles_per_recording != 0 && self.cycles_to_next_recording == 0;
```

to something like this

```rust
let enable_recording = self.cycles_per_recording != 0 && self.cycles_since_last_recording == self.cycles_per_recording - 1;
```
which might be less readable. Otherwise, the order of assignments might be changed, but that would change the meaning of `cycles_since_last_recording`. 
I'm not sure what to think about these options, what is your opinion about this?

tools/pepsi/src/pre_game.rs
```diff
@@ -65,7 +67,7 @@ pub async fn pre_game(arguments: Arguments, repository: &Repository) -> Result<(
 
     recording(
         RecordingArguments {
-            cycler_instances_to_be_recorded: arguments.cycler_instances_to_be_recorded,
+            recording_settings: arguments.recording_settings,
```
h3ndrk, line 70: Naming: Recording frequencies? Then you can also shorten the comment of the command line argument.

crates/code_generation/src/cyclers.rs
h3ndrk: Still an idea: extract the counting into its own struct to implement single responsibility principle?

oleflb: I like this

tools/pepsi/src/pre_game.rs
oleflb: Frequency is something divided by time. I guess this would be a period or interval

julianschuler: I renamed this to `recording_interval` now.

crates/code_generation/src/cyclers.rs
julianschuler: I now factored this out to a `RecordingTrigger` struct, which can also be more easily extended in the future.

docs/tooling/recording_and_replay.md
```diff
@@ -8,11 +8,11 @@ A started communication server during replay can be used to investigate the reco
...
 
 - Manual upload to a robot
-    - Use `./pepsi recording ...` to enable recording for certain cycler instances, e.g., `./pepsi recording Control VisionTop`
-        - This will set the cycler instances in `etc/parameters/framework.json`
+    - Use `./pepsi recording ...` to enable recording at different recording rates, e.g., `./pepsi recording Control=1,VisionTop=30`
+        - This will set the cycler instances and recording rate in `etc/parameters/framework.json`
     - Use `./pepsi upload ...` to upload as usual (this includes the framework configuration with the cycler instances)
 - Pregame
-    - Use `./pepsi pregame --cycler-instances-to-be-recorded ... ...` to enable recording for certain cyclers and upload to the robot in one step, e.g., `./pepsi pregame --cycler-instances-to-be-recorded Control,VisionTop ...`
+    - Use `./pepsi pregame --recording-settings ... ...` to enable recording at different recording rates and upload to the robot in one step, e.g., `./pepsi pregame --recording-settings Control=1,VisionTop=30 ...`
```
h3ndrk, line 15: Also rename the command line arguments in the documentation?

tools/pepsi/src/pre_game.rs
```diff
@@ -37,9 +37,10 @@ pub struct Arguments {
     /// Skip the OS version check
     #[arg(long)]
     pub skip_os_check: bool,
-    /// Cycler instances to record e.g. Control or VisionBottom (don't specify to disable recording)
-    #[arg(long, default_value = "Control")]
-    pub cycler_instances_to_be_recorded: Vec<String>,
+    /// Intervals between cycle recordings, e.g. Control=1,VisionTop=30 to record every cycle in Control
+    /// and one out of every 30 in VisionTop. Set to 0 or don't specify to disable recording for a cycler.
+    #[arg(long, value_delimiter=',', value_parser = parse_key_value::<String, usize>, default_value = "Control=1,VisionTop=30,VisionBottom=30")]
+    pub recording_intervals: Vec<(String, usize)>,
```
h3ndrk, line 43: Is it possible to change the type to `HashMap<String, usize>`? This way you could drop the conversion from `Vec` to `HashMap`?

tools/pepsi/src/recording.rs
```diff
@@ -7,15 +7,28 @@ use repository::Repository;
...
+    T: std::str::FromStr,
+    T::Err: Error + Send + Sync + 'static,
+    U: std::str::FromStr,
+    U::Err: Error + Send + Sync + 'static,
+{
+    let pos = s
+        .find('=')
+        .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{}`", s))?;
+    Ok((s[..pos].parse()?, s[pos + 1..].parse()?))
+}
```
h3ndrk, line 34: Please don't abbreviate words `s` and `pos`.

h3ndrk (changes requested):

crates/code_generation/src/structs.rs
```diff
@@ -114,3 +123,26 @@ fn hierarchy_to_token_stream(
...
+            pub fn cycle_finished(&mut self) {
+                self.counter = (self.counter + 1) % self.recording_interval;
+            }
+
+            pub fn should_record(&self) -> bool {
+                self.recording_interval != 0 && self.counter == 0
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
julianschuler: Thanks, done.

crates/code_generation/src/structs.rs
```diff
@@ -3,7 +3,9 @@ use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
 use source_analyzer::{struct_hierarchy::StructHierarchy, structs::Structs};
 
-pub fn generate_structs(structs: &Structs) -> TokenStream {
+use crate::Execution;
+
+pub fn generate_structs(structs: &Structs, mode: Execution) -> TokenStream {
```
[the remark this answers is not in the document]
julianschuler: Reverted when moving `RecordingTrigger` to the `framework` crate.

docs/tooling/recording_and_replay.md
julianschuler: Good catch, fixed.

tools/pepsi/src/pre_game.rs
julianschuler: `Vec` is natively supported by `clap_derive`, while `HashMap` is unfortunately not. By using a `Vec`, only the tuples themselves instead of the whole collection has to be parsed, otherwise we would also need to add a parser for the `HashMap` itself.

h3ndrk (dismissed):
Jawollo

PR #871, opened 2024-04-17
Plot panel

### LINKED ISSUE #844: Enum Plot Panel: Box Name Right
Place the label of boxes in the enum plot panel on the right border of boxes, so that they don't leave the image to the left in `Follow` mode

CallMeElle:
## Introduced Changes

In twix Plot Panel if 'follow' is active the name of the segment stays at the leftmost visible position. It disappears as soon as the segments visible size is 1/18 of the whole bar (that it doesn't overlap too much with the text of the next segment)

Fixes #844

## ToDo / Known Issues

none

If a segment is smaller then its title, the text overlaps with the title of the next segment.

## How to Test

In twix open Map and Plot Panel, subscribe to e.g. role and change to 'follow'. The change can be observed when a segments left border is not shown on the panel anymore but the title is still there!

### DIFF
--- a/tools/twix/src/panels/enum_plot.rs
+++ b/tools/twix/src/panels/enum_plot.rs
@@ -147,12 +147,9 @@ impl EnumPlotPanel {
         }
 
         plot_ui.text(
-            Text::new(
-                PlotPoint { x: text_x, y: 0.9 },
-                name,
-            )
-            .color(Color32::WHITE)
-            .anchor(eframe::emath::Align2::LEFT_TOP),
+            Text::new(PlotPoint { x: text_x, y: 0.9 }, name)
+                .color(Color32::WHITE)
+                .anchor(eframe::emath::Align2::LEFT_TOP),
         );
     }
 
--- a/tools/twix/src/panels/enum_plot.rs
+++ b/tools/twix/src/panels/enum_plot.rs
@@ -117,6 +117,8 @@ impl EnumPlotPanel {
         const BORDER_WIDTH: f32 = 2.0;
 
         let viewport_left_edge = plot_bounds.min()[0];
+        let viewport_right_edge = plot_bounds.max()[0];
+        let min_boxsize_text = viewport_right_edge - viewport_left_edge;
 
         let name = segment.name();
         let color = color_hash(&name);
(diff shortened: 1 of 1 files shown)


tools/twix/src/panels/enum_plot.rs
```diff
@@ -112,9 +112,14 @@ impl Panel for EnumPlotPanel {
...
 impl EnumPlotPanel {
-    fn plot_segment(plot_ui: &mut PlotUi, segment: &Segment) {
+    fn plot_segment(plot_ui: &mut PlotUi, segment: &Segment, plot_bounds: &PlotBounds) {
         const VERTICAL_MARGIN: f64 = 0.05;
         const BORDER_WIDTH: f32 = 2.0;
+        const MIN_SEGMENT_RELATION_FOR_TEXT: f64 = 1.0 / 18.0;
+
+        let viewport_left_edge = plot_bounds.min()[0];
+        let viewport_right_edge = plot_bounds.max()[0];
+        let min_boxsize_text = viewport_right_edge - viewport_left_edge;
```
Narcha, line 122: Culling based on a fixed relative width is a bit brittle. Let's remove it for now and implement it properly after bumping the egui version.
```suggestion
        let viewport_left_edge = plot_bounds.min()[0];
```

tools/twix/src/panels/enum_plot.rs
```diff
@@ -133,16 +138,18 @@ impl EnumPlotPanel {
             .name(&name)
             .stroke(Stroke::new(BORDER_WIDTH, color)),
         );
+
+        let mut text_x = start + 0.05;
+        if viewport_left_edge > text_x
+            && end > (viewport_left_edge + (min_boxsize_text * MIN_SEGMENT_RELATION_FOR_TEXT))
+        {
+            text_x = viewport_left_edge + 0.05;
+        }
```
Narcha, line 147: Rewrite without `mut`
```suggestion
        let text_x = if start < viewport_left_edge && viewport_left_edge < end {
            viewport_left_edge + 0.05
        } else {
            start + 0.05
        };
```

Narcha (changes requested):
Nice feature, I left a few remarks

Narcha (approved):
LGTM :+1:

PR #872, opened 2024-04-17
Field border detection fix

oleflb:
## Introduced Changes

Field borders were all over the place. Not anymore.
Also `horizon_margin` is removed, in its current implementation it harmed the field border detection.

Fixes #

## ToDo / Known Issues

## How to Test

* Check field border in twix


crates/projection/src/horizon.rs
```diff
@@ -17,6 +17,11 @@ impl Horizon {
         normal: vector![0.0, -1.0],
     };
 
+    pub fn is_below(&self, point: Point2<Pixel>) -> bool {
```
h3ndrk, line 20: Rename function as discussed

h3ndrk (changes requested):

h3ndrk (approved):

PR #881, opened 2024-04-18
look at hypothetical balls when searching

schmidma:
## Introduced Changes

When searching: choose head motion dynamically. When we have any ball candidate, we look at it not checking, whether its validity is high enough.


Narcha (dismissed):
Works as intended

Narcha:
... But the CI is failing.

Narcha (approved):

PR #890, opened 2024-04-18
Deny new searchbehavior to keeper

JonasHolz:
## Introduced Changes

As it says in the title. This prevents the keeper from using the new search behavior

## ToDo / Known Issues

None

None

## How to Test

Check if keeper or defenders go into search behavior


schmidma (approved):
Still not entirely correct and fancy, but works for now

Issue #880, opened 2024-04-18
Post-game task: dynamic goalpost size


Narcha:
@schmidma suggested to change the goalpost obstacle size dynamically based on the game state.

PatrickGoettsch:
Fixed by #911

PatrickGoettsch:
Parameter for goal post radius was not used in code

schluis:
Not needed anymore, since the parameter for the goalpost size was never used, this is fixed now

PR #892, opened 2024-04-18
increases radius for non strikers

LisaSchulz97:
## Introduced Changes

increases radius for non strikers to avoid colliding with the goalpost. The new parameter goalpost_radius_decrease will decrease the goalpost_radius for the striker to maintain the current behaviour. 

currently if the ball is closer than 60 cm the striker ignores obstacles, although goalpost exists. Think about better behaviour. 

## How to Test

First with Gamecontroller. First without the ball. The goalkeeper should make a large radius around the goalpost. Secondly with the ball closed to the robot (> 60cm), it should make a smaller radius (= 20 cm around the goalpost)

### DIFF
--- a/crates/control/src/dribble_path_planner.rs
+++ b/crates/control/src/dribble_path_planner.rs
@@ -48,7 +48,7 @@ pub fn plan(
     } else {
         let mut obstacles = world_state.obstacles.clone();
         for obstacle in &mut obstacles {
-            if obstacle.kind == ObstacleKind::GoalPost {
+            if matches!(obstacle.kind, ObstacleKind::GoalPost) {
                 obstacle.radius_at_hip_height = (obstacle.radius_at_hip_height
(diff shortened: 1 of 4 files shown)


crates/control/src/dribble_path_planner.rs
```diff
@@ -1,6 +1,7 @@
 use framework::AdditionalOutput;
 use spl_network_messages::Team;
 use std::f32::consts::PI;
+use types::obstacles::ObstacleKind;
 use types::{
     filtered_game_controller_state::FilteredGameControllerState, parameters::DribblingParameters,
```
oleflb, line 6: please merge imports

oleflb (changes requested):
Minor remark, otherwise looks great

PatrickGoettsch:
Solves #880 too.

PR #894, opened 2024-04-18
Anakin stop panicking

knoellle:
## Introduced Changes

Upon panic, always prints location regardless of payload.
On main, only the payload of `panic!("hello world")` is printed, but not the one of `panic!("{some_variable}")`.

We only checked for `&str`, but panic with format string has a `String` payload.

Fixes #

## How to Test

Insert panics somehwere and observe how they are printed.

### DIFF
--- a/crates/code_generation/src/execution.rs
+++ b/crates/code_generation/src/execution.rs
@@ -39,10 +39,13 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
                 let keep_running = keep_running.clone();
                 std::panic::set_hook(Box::new(move |panic_info| {
                     keep_running.cancel();
+                    eprint!("panic occurred");
                     if let Some(payload) = panic_info.payload().downcast_ref::<&str>() {
-                        eprintln!("panic occurred: {payload:?} {:?}", panic_info.location());
-                    } else {
-                        eprintln!("panic occurred");
+                        eprint!(": {payload:?}");
+                    }
+                    eprintln!();
+                    if let Some(location) = panic_info.location() {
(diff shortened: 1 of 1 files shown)


crates/code_generation/src/execution.rs
```diff
@@ -39,10 +39,16 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
...
+                    eprint!("panic occurred");
                     if let Some(payload) = panic_info.payload().downcast_ref::<&str>() {
-                        eprintln!("panic occurred: {payload:?} {:?}", panic_info.location());
-                    } else {
-                        eprintln!("panic occurred");
+                        eprint!(": {payload:?}");
+                    }
+                    if let Some(payload) = panic_info.payload().downcast_ref::<String>() {
+                        eprint!(": {payload:?}");
+                    }
```
schmidma, line 48: Can we cast to something like `Box<dyn Debug>`?

crates/code_generation/src/execution.rs
```diff
@@ -39,10 +39,16 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
...
+                    eprintln!();
```
schmidma, line 49: why?

crates/code_generation/src/execution.rs
schmidma: The entire `PanicInfo` seems to implement `Display`?! can we use that?

schmidma (changes requested):

knoellle:
Closing in favor of #895

PR #902, opened 2024-04-18
Fix incorrect parameter path

julianschuler:
## Introduced Changes

Fixes the incorrect parameter path resulting in a crash on main introduced in #840.

## ToDo / Known Issues

None.

None.

## How to Test

Upload to a NAO.


knoellle (approved):
Whoopie

PR #913, opened 2024-04-19
Disable look around when we get the ready signal via gamecontroller

JonasHolz:
## Introduced Changes

As it says in the title

## ToDo / Known Issues

None

None

## How to Test

Check if we still do look around if we get the signal from the gamecontroller

### DIFF
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1,7 +1,7 @@
 {
   "object_detection": {
     "object_detection_top": {
-      "enable": true,
+      "enable": false,
       "intersection_over_union_threshold": 0.45,
       "distance_to_referee_position_threshold": 2.0,
       "keypoint_confidence_threshold": 0.2,


crates/control/src/behavior/look_around.rs
```diff
@@ -1,24 +1,38 @@
...
-                game_phase: GamePhase::PenaltyShootout { .. },
-                ..
+pub fn execute(
+    world_state: &WorldState,
+    game_controller_state: Option<GameControllerState>,
+) -> Option<MotionCommand> {
+    if let Some(game_controller_state) = game_controller_state {
+        match (
+            world_state.filtered_game_controller_state,
+            game_controller_state.game_state,
```
schluis, line 15: Why do you use the `game_controller_state` and not the `filtered_game_controller_state`?

crates/control/src/behavior/look_around.rs
```diff
@@ -1,24 +1,38 @@
...
+                    game_phase: GamePhase::PenaltyShootout { .. },
+                    ..
+                }),
+                _,
+                _,
+            ) => None,
+            (
+                _,
+                GameState::Initial | GameState::Playing,
+                PrimaryState::Ready | PrimaryState::Playing,
```
schluis, line 29: I would assume that it's enough to check whether we're in `Initial` or `Ready {..}` and then return `None`.

crates/control/src/behavior/look_around.rs
```diff
@@ -1,24 +1,38 @@
...
+            ) => Some(MotionCommand::Stand {
+                head: types::motion_command::HeadMotion::LookAround,
             }),
-            _,
-        ) => None,
-        (_, PrimaryState::Ready | PrimaryState::Playing) => Some(MotionCommand::Stand {
```
schluis, line 19: Or to just remove `PrimaryState::Ready` here

schluis (changes requested):
🚀

crates/control/src/behavior/look_around.rs
JonasHolz: Because if we get the signal from seeing the visual referee the filtered_game_controller_state is updated.

crates/control/src/behavior/look_around.rs
JonasHolz: No because we still want to do the intitial lookaround if we get a signal via the visual referee

crates/control/src/behavior/look_around.rs
JonasHolz: see comment above

schluis:
Will be tracked further by #933

PR #923, opened 2024-04-19
Fix tracking lost ground contact

julianschuler:
## Introduced Changes

This PR fixes tracking lost ground contact during penalized.

## ToDo / Known Issues

- [x] Test on a field (currently all occupied)

None.

## How to Test

Upload to a robot and have fun penalizing and unpenalizing while picking up the robot, moving it around, etc.

### DIFF
--- a/crates/control/src/localization.rs
+++ b/crates/control/src/localization.rs
@@ -1,6 +1,5 @@
 use std::{
     f32::consts::{FRAC_PI_2, PI},
-    mem::take,
     time::{Duration, SystemTime},
 };
 
@@ -226,22 +225,8 @@ impl Localization {
                 {
                     if self.is_penalized_with_motion_in_set_or_initial {
                         if self.was_picked_up_while_penalized {
-                            self.hypotheses = take(&mut self.hypotheses_when_entered_playing);
-
-                            let penalized_poses = generate_penalized_poses(
-                                context.field_dimensions,
-                                *context.penalized_distance,
-                            );
-                            self.hypotheses_when_entered_playing = penalized_poses
-                                .into_iter()
-                                .map(|pose| {
-                                    ScoredPose::from_isometry(
-                                        pose,
-                                        *context.penalized_hypothesis_covariance,
-                                        *context.initial_hypothesis_score,
-                                    )
-                                })
-                                .collect();
+                            self.hypotheses
+                                .clone_from(&self.hypotheses_when_entered_playing);
                         }
                         self.is_penalized_with_motion_in_set_or_initial = false;
                         self.was_picked_up_while_penalized = false;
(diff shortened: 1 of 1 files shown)


crates/control/src/localization.rs
```diff
@@ -208,26 +208,44 @@ impl Localization {
...
-                    }) || !context.has_ground_contact) =>
-            {
-                if self.is_penalized_with_motion_in_set_or_initial {
-                    if self.was_picked_up_while_penalized_with_motion_in_set_or_initial {
-                        self.hypotheses = take(&mut self.hypotheses_when_entered_playing);
-
+                    })
+                {
+                    if self.is_penalized_with_motion_in_set_or_initial {
+                        if self.was_picked_up_while_penalized {
```
schluis, line 223: Please restructure this so you don't have to check it twice. I think this is quite confusing to read.

crates/control/src/localization.rs
```diff
@@ -208,26 +208,44 @@ impl Localization {
...
+                                .into_iter()
+                                .map(|pose| {
+                                    ScoredPose::from_isometry(
+                                        pose,
+                                        *context.penalized_hypothesis_covariance,
+                                        *context.initial_hypothesis_score,
+                                    )
+                                })
+                                .collect();
+                        }
```
schluis, line 240: Why do you regenerate this and refill this?  Would it be sufficient to copy the `hypotheses_when_entered_playing`?

crates/control/src/localization.rs
```diff
@@ -208,26 +208,44 @@ impl Localization {
...
-            {
-                if self.is_penalized_with_motion_in_set_or_initial {
-                    if self.was_picked_up_while_penalized_with_motion_in_set_or_initial {
-                        self.hypotheses = take(&mut self.hypotheses_when_entered_playing);
-
+                    })
+                {
+                    if self.is_penalized_with_motion_in_set_or_initial {
+                        if self.was_picked_up_while_penalized {
+                            self.hypotheses = take(&mut self.hypotheses_when_entered_playing);
```
schluis, line 224: I think we should do something more intelligent: sometimes, the robot is picked up by accident during `is_penalized_with_motion_in_set_or_initial` and is placed at a similar location as before. We could keep the old hypothesis, reduce their probability and add the default penalized poses

schluis (changes requested):

crates/control/src/localization.rs
julianschuler: Thanks, done.

crates/control/src/localization.rs
julianschuler: Good catch, done.

crates/control/src/localization.rs
julianschuler: I agree, though this should be done in a separate PR. Feel free to open an issue for that.

crates/control/src/localization.rs
schluis: Doesn't seem like it, even after switching to the correct branch 😉

julianschuler: Could you please elaborate what you mean? The previously included double check of `self.was_picked_up_while_penalized` is not present anymore.

schluis (approved):

PR #937, opened 2024-04-19
look_around

oleflb:
## Introduced Changes

 - Look around max angle ~ 45°
 - Increased time per look direction (not total)
 - Removed halfway directions

Fixes #

## ToDo / Known Issues

## How to Test

* Check penalized -> playing or initial -> ready. Robot now uses other look motion


crates/control/src/motion/look_around.rs
```diff
@@ -92,70 +99,31 @@ impl LookAround {
             return;
```
knoellle, line 99: `look_around` and `quick_search` are identical now. Remove one of them?

PR #940, opened 2024-04-19
paint lines in map panel from vision

schmidma:
instead of painting lines generated by the Control cycler, twix now projects the lines generated by the vision cycler on its own. This improves the user experience while replaying


tools/twix/src/panels/map/layers/lines.rs
```diff
@@ -4,38 +4,73 @@ use color_eyre::Result;
...
+    camera_matrix_bottom: ValueBuffer,
 }
 
-impl Layer<Field> for Lines {
+impl Layer<Ground> for Lines {
     const NAME: &'static str = "Lines";
 
     fn new(nao: Arc<Nao>) -> Self {
-        let lines_in_field = nao.subscribe_output(
-            CyclerOutput::from_str("Control.additional.localization.measured_lines_in_field")
```
h3ndrk, line 24: Yesterday, I changed this panel from using Vision inputs to only using Control inputs because we often record Vision with a very low frequency but, e.g., the `robot_to_field` is still recorded with high frequency. You are now reverting my change. Is this intentional?

knoellle: Yeah, I'm not sure about this change either.
On one hand, its useful to see exactly what vision produced on the map vs. image panel, but when not recording vision with full frequency, it becomes much harder to debug loca issues.

Maybe the real solution is to make our algorithms deterministic and allowing stepping frame-by-frame in one cycler in replay...

schmidma: Further supporting my perspective:

The line data is produced in vision, if I want to visualize line data, I'd expect it to be visualized as it is produced by the node responsible of producing that data.
Using data from control is violating this assumption. For me this sounds like mixing responsibilities and (dirty) fixing the issue that vision is not regularly recorded.
Usually this is annoying when debugging localization. But then I'd expect a totally different map layer like e.g. "line association" as this is the important information to investigate localization issues.

So, use line data from vision if I want to display line data, and use line associations or additional debug information to display localization data.

h3ndrk: I can understand Schmiddies reasoning. On the other hand, I'm also with knoellle that we want to debug localization with localization inputs where I don't care about the origin. One could also argue displaying vision things in the map is also mixing domains, but no hard feelings. For me, it would be fine if we keep the lines in the line association layer, then everybody gets its slice.

oleflb: @h3ndrk do I understand you correctly that you are fine with @schmidma changes? I am not sure what you mean with "line association layer"

oleflb (changes requested):
Looks good codewise. I will await the result of the discussion

tools/twix/src/panels/map/layers/lines.rs
h3ndrk: My proposal is to have two layers:

1. Line and associations seen as input to localization from Control with high frequency for debugging localization 
2. Lines and other features generated by Vision to debug Vision with probably lower frequency

If I read this PR correctly 1. existed before but now only 2. exists. I would vote for having both to make everyone happy.

schmidma: I added the lines from `Control` to the line correspondences layer

h3ndrk: Thanks, resolving

h3ndrk (approved):

PR #948, opened 2024-04-19
Fix segments ending on limbs

h3ndrk:
## Introduced Changes

Previously:

Screenshot from 2024-04-19 14-43-29

With this PR:

Screenshot from 2024-04-19 19-36-14

## ToDo / Known Issues

- We need to tune the limbs because now we have short segments near them

It's perfect

## How to Test

Deploy on a NAO and look through the bottom camera

### DIFF
--- a/crates/geometry/src/line.rs
+++ b/crates/geometry/src/line.rs
@@ -37,8 +37,13 @@ impl<Frame> Line2<Frame> {
         self.0.y() - (self.0.x() * self.slope())
     }
 
+    pub fn project_onto_segment_in_x_axis(&self, point: Point2<Frame>) -> f32 {
+        let rise = (point.x() - self.0.x()) * self.slope();
+        rise + self.0.y()
+    }
+
     pub fn is_above(&self, point: Point2<Frame>) -> bool {
-        self.signed_distance_to_point(point) >= 0.0
+        point.y() >= self.project_onto_segment_in_x_axis(point)
     }
 
     pub fn signed_distance_to_point(&self, point: Point2<Frame>) -> f32 {
--- a/crates/types/src/limb.rs
+++ b/crates/types/src/limb.rs
@@ -4,7 +4,7 @@ use geometry::line::Line;
 use serde::{Deserialize, Serialize};
 
 use coordinate_systems::Pixel;
-use linear_algebra::{point, Point2};
+use linear_algebra::Point2;
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
 
 #[derive(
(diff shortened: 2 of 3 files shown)


knoellle:
What is blocking this PR? I thought this had been merged already. :D

I was looking through the replay from the game against Nao Devils and was confused why this wasn't in the game branch.
Did we forget to merge-squash this into the game branch?

h3ndrk:
> What is blocking this PR?

I have written it in the description:

> We need to tune the limbs because now we have short segments near them

In its current form, it will likely have a large impact on other detection algorithms.

Maybe we should move this back to In Progress? @Irwinlim1097 and @MeiiMoon said they want to continue work on it. Is this still correct?

Irwinlim1097:
Yes me and @MeiiMoon were looking into this code during GO, as we wanted to learn more about the whole Hulks code in general. If this has to be merged into main I could help review it, else it could be moved to In Progress.

h3ndrk:
@Irwinlim1097 @MeiiMoon then feel free to take this over from me, improve the limbs, and bring this into `main`. I will move it back to In Progress and assign you two. I'll be in the lab tomorrow, if you have questions.

knoellle:
What is the state here?
As I recall, this fix has a decent amount of potential to improve our vision but we seem to have forgotten about it.
Can we get this done before robocup?

h3ndrk:
Main problem is the lag between image and camera matrix. Therefore the limbs are lagging and the segmenter produces segments on limbs. One hacky idea was to increase limb size but that will have other drawbacks. The proper fix would be to reduce lag.

@Irwinlim1097 @MeiiMoon What is your state?

knoellle:
Ok, then I would suggest check again with #1107.
From what we can tell, this should completely get rid of the relative delay between camera matrix and image capture.

h3ndrk:
Since my time with a NAO is limited, feel free to test it for me 💪

MeiiMoon:
I tested the respective "limbo" branch, and the delay from the sensors still cause the robots body part to occasionally exceed the projected limbs. Therefore, the delay problem still has to be solved, because the current state would cause problem for the localization.

oleflb:
@MeiiMoon is this a problem? If it improves on the current situation we should merge it

MeiiMoon:
@oleflb When we were in Eindhoven, I had a talk with @h3ndrk and he said there is some issue here that beginner will not be able to solve. Unfortunately I cannot remember what exactly was the problem. So that's why I left this task, but I also did not update it further.

knoellle:
The problem is detailed [here](https://github.com/HULKs/hulk/pull/948#issuecomment-2212327930):
> Main problem is the lag between image and camera matrix. Therefore the limbs are lagging and the segmenter produces segments on limbs. One hacky idea was to increase limb size but that will have other drawbacks. The proper fix would be to reduce lag.

oleflb:
No longer relevant with change to new platform

PR #949, opened 2024-04-19
Dynamic replayer

knoellle:
## Introduced Changes

1. Update replayer outputs when parameters are changed
2. Update replayer periodically (currently 1 Hz) to make up for the fact that we cannot react to new subscriptions from outside of communication.
3. ~~Load replay index after starting the UI~~
  3.1 ~~Show live loading progress~~
  3.2 ~~Allows scrubbing through the timeline even while parts are still being loaded~~
4. ~~Run all index building and~~ replaying in a separate thread from the UI.
  Uses `tokio::watch` to send intended time, resulting in skipped intermediate values if the replayer thread can't keep up with the requests from the UI, but with eventual consistency.

Fixes #

- Also update when new outputs are subscribed. (already soft-fixed by re-replaying at a fixed rate)
- Build recording indices in parallel instead of interleaved.

## How to Test

1. Replay something, change parameters and immediately observe the changes in the output.
2. Subscribe some new key or open image panel. After at most one second, the data should appear without having to jiggle the replayer.
3. Does not apply anymore.
4. UI should look a bit smoother on vegetables.

### DIFF
--- a/crates/framework/src/recording_index.rs
+++ b/crates/framework/src/recording_index.rs
@@ -11,34 +11,34 @@ use color_eyre::eyre::WrapErr;
 #[derive(Debug)]
 pub struct RecordingIndex {
     file: File,
-    frames: Vec<RecordingFrameMetadata>,
+    file_length: u64,
     scan_offset: u64,
+    frames: Vec<RecordingFrameMetadata>,
 }
 
 impl RecordingIndex {
     pub fn read_from(recording_file: impl AsRef<Path>) -> color_eyre::Result<Self> {
-        let file = File::open(&recording_file)
+        let mut file = File::open(&recording_file)
             .wrap_err_with(|| format!("failed to open {}", recording_file.as_ref().display()))?;
+        file.seek(SeekFrom::End(0))
+            .wrap_err("failed to seek to end of file")?;
+        let file_length = file
+            .stream_position()
+            .wrap_err("failed to get stream position of end of file")?;
         Ok(Self {
             file,
-            frames: Vec::new(),
+            file_length,
             scan_offset: 0,
+            frames: Vec::new(),
         })
     }
 
-    pub fn collect_next_frame_metadata(&mut self) -> color_eyre::Result<Option<f32>> {
-        self.file
-            .seek(SeekFrom::End(0))
-            .wrap_err("failed to seek to end of file")?;
-        let file_length = self
-            .file
-            .stream_position()
-            .wrap_err("failed to get stream position of end of file")?;
+    pub fn collect_next_frame_metadata(&mut self) -> color_eyre::Result<Option<()>> {
         self.file
             .seek(SeekFrom::Start(self.scan_offset))
             .wrap_err("failed to rewind file")?;
 
-        if self.scan_offset >= file_length {
+        if self.scan_offset >= self.file_length {
             return Ok(None);
         }
         let Some(timestamp) = end_of_file_error_as_option(deserialize_from(&mut self.file))
@@ -67,7 +67,7 @@ impl RecordingIndex {
         self.file
             .seek(SeekFrom::Current(length as i64))
             .wrap_err("failed to seek to end of data")?;
-        if self.scan_offset + header_length + length as u64 > file_length {
+        if self.scan_offset + header_length + length as u64 > self.file_length {
             eprintln!("unexpected end of file of recording file");
             return Ok(None);
         }
@@ -87,11 +87,21 @@ impl RecordingIndex {
             header_offset: header_length.try_into().unwrap(),
             length,
         });
-        Ok(Some(self.scan_offset as f32 / file_length as f32))
+        Ok(Some(()))
     }
 
     pub fn number_of_frames(&self) -> usize {
         self.frames.len()
+    }
+
+    pub fn scan_state(&self) -> ScanState {
+        if self.scan_offset < self.file_length {
+            ScanState::Loading {
+                progress: self.scan_offset as f32 / self.file_length as f32,
+            }
+        } else {
+            ScanState::Done
+        }
     }
 
     pub fn find_latest_frame_up_to(
@@ -154,6 +164,11 @@ pub struct RecordingFrame {
 pub struct Timing {
     pub timestamp: SystemTime,
     pub duration: Duration,
+}
+
+pub enum ScanState {
+    Loading { progress: f32 },
+    Done,
 }
 
 fn end_of_file_error_as_option<T>(result: Result<T, Error>) -> Result<Option<T>, Error> {
--- a/crates/framework/src/recording_index.rs
+++ b/crates/framework/src/recording_index.rs
@@ -12,78 +12,82 @@ use color_eyre::eyre::WrapErr;
 pub struct RecordingIndex {
     file: File,
     frames: Vec<RecordingFrameMetadata>,
+    scan_offset: u64,
 }
 
 impl RecordingIndex {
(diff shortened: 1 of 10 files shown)


crates/hulk_replayer/src/user_interface.rs
```diff
@@ -17,11 +23,32 @@ pub struct ReplayerApplication {
...
+
+            runtime.block_on(async move {
+                loop {
+                    parameters_updated.notified().await;
+                    println!("requesting re-replaying");
+                    rerun_request_sender.send(()).expect("channel closed");
+                    context.request_repaint();
+                }
+            });
+        });
```
h3ndrk, line 50: Or move the replaying into this thread. The fact that it currently is evaluated synchronously is also smelly. The semantic of replaying is fire-and-forget from the viewpoint of the UI, so that would make sense.

crates/hulk_replayer/src/user_interface.rs
h3ndrk, line 1: I would like to see this rebased onto #971

h3ndrk (changes requested):

crates/hulk_replayer/src/user_interface.rs
knoellle: Done

crates/hulk_replayer/src/user_interface.rs
knoellle: I did this and much more, see updated description.

crates/framework/src/recording_index.rs
h3ndrk, line 1: I think your feature 3 adds a lot of complexity without much benefit. I think we don't need this feature and should therefore keep the changeset simple to keep this tool maintainable. Index creation takes 3-5 seconds on my vegetable. I'm used to such loading times and there is no need for me to already look into partial data. I would propose to keep it simple, remove all the asynchronity for loading etc. and just index once at startup before showing the UI. This also makes error handling during loading much easier btw. 😉

crates/hulk_replayer/src/user_interface.rs
h3ndrk: Thanks

crates/hulk_replayer/src/window.rs
```diff
@@ -92,6 +106,59 @@ fn join_timing(replayer: &Replayer<ReplayerHardwareInterface>) -> FrameRange {
...
+                    }
+                    Ok(None) => {}
+                    Err(error) => eprintln!("{error}"),
+                }
+            }
+            if !frames_found {
+                break;
+            }
+            if time.has_changed().unwrap() {
+                time.mark_unchanged();
```
h3ndrk, line 138: This looks like a lost wake-up? After checking it could change again but afterwards you discard it by marking it as unchanged.

crates/hulk_replayer/src/labels.rs
```diff
@@ -52,6 +51,14 @@ impl<'state> Widget for Labels<'state> {
             {
                 child_ui.label(format!("{} frames", label_content.number_of_frames));
             }
+            if row_height
+                >= (3.0 * ui.style().text_styles.get(&TextStyle::Body).unwrap().size)
+                    + ui.spacing().item_spacing.y
```
h3ndrk, line 56: One spacing unit is missing, right?

crates/hulk_replayer/src/window.rs
```diff
@@ -14,56 +19,65 @@ use crate::{
...
-            .collect::<BTreeMap<_, _>>();
-        for (name, frame) in frames {
-            if let Some(frame) = frame {
-                self.replayer
-                    .replay(&name, frame.timing.timestamp, &frame.data)
-                    .expect("failed to replay frame");
-            }
-        }
+        let frame_range = join_timing(&self.replayer.lock().unwrap());
+        let timestamp = self.position.map_to_absolute_time(&frame_range);
```
h3ndrk, line 60: Hmmm, maybe we need to rethink for a second which party is responsible for holding which state and when we update what. What I would feel as a clean approach would be that, e.g., the replayer thread holds the replayer and then provides the high-level information for the UI and it triggers repaints. The UI would then be very simple and stupid, i.e., doesn't have much state associated with the replayer. Maybe this idea wrongly assumed details but we should still look into checking the intention IMHO.

crates/framework/src/recording_index.rs
h3ndrk: Streaming from slow network shares is a new requirement that I didn't know you wanted to tackle in this PR. But okay, if this is the plan, I agree with your reasoning but have the following remarks:

- Feedback: I disagree, it is obvious and understandable by the user what happens. And the 3-5 seconds are just a kind of relative time. As mentioned before, I'm used to such loading times, so no issue for users like me. For users with more beefy hardware, 3-5 seconds is too long without feedback but that still doesn't occur because they will wait shorter because of their hardware, so also no issue here. Nevertheless, this argument is negligible if we talk about network shares.
- Complexity: Of course this adds a lot of complexity compared to current `main`, this is far from marginal. You need to handle asynchronity, synchronization, information propagation across thread boundaries, update the UI, etc. How many channels are added in this PR? 😉 I think, it is fair to question, whether it is worth the features (with the additional requirement I can agree with it).
- Error Handling: If we cannot replay a frame, this is a hard error Iirc. But since replaying a frame is now detached from the trigger, reporting errors back to the initiator is more difficult.

crates/hulk_replayer/src/window.rs
knoellle: This is a watch channel, so we only care about the last value.
So lost updates are okay as long as we get the last value before a long break.
Note that I check the flag, remove the flag, then inspect the value.
If there is another update between clearing the flag and reading the value, worst case is simulating the same value twice, since the flag is then set again in the next iteration.

h3ndrk: My point was an update between checking and removing the flag. After checking, you have decided that you want to clear the flag regardless if there was another update.

knoellle: That's a lost update, not a lost wakeup, since in that case too we use the newer value.

crates/hulk_replayer/src/window.rs
knoellle: I'm not sure what you mean.
The UI thread owns no state apart from the information about the viewport -> timeline mapping.
Everything else it needs for rendering (current time, cyclers, frame times) it retrieves from the replayer.

crates/hulk_replayer/src/labels.rs
knoellle: I realized both of the if conditions weren't quite right.
Now I'm using `Ui::available_height` to check if another label would fit into the remaining available space.

crates/framework/src/recording_index.rs
knoellle: A large part of the added complexity stems from the fact that we wanted to remove the replayer from the UI thread.
The complexity on top of that which is necessary for the lazy loading is quite acceptable I think.

This also applies to your point about error handling. While I agree that having the replayer in another thread makes proper handling of errors from replaying a frame more difficult, it has nothing to do with the lazy loading.

crates/hulk_replayer/src/window.rs
h3ndrk: That might be true, but my feeling is that the responsibilities are a little cluttered at the moment. My intention is to trigger another evaluation of your design. Try to optimize for minimal interfaces and minimal state. For example where should we calculate timing etc.

knoellle: Applied changes as discussed.
Please have another look.

h3ndrk (approved):

PR #973, opened 2024-04-23
Allow symlinks for log directory

knoellle:
## Introduced Changes

I wanted to share the log directory between my git worktrees, but the `logs/` pattern in the gitignore file does not ignore a symlink name `logs`.

Fixes #

## How to Test

Create a symlink called `logs` in the root of the repo and run `git status`. It should no longer be detected as an untracked file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,4 @@ rust-toolchain.toml
 
 site/
 
-logs/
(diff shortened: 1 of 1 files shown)


julianschuler (approved):
LGTM!

PR #1005, opened 2024-05-08
run checks on merge_group trigger

schmidma:
## Introduced Changes

If I read and understand the documentation correctly, this is the way to go now?!
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group

## ToDo / Known Issues

none

none

## How to Test

check with the documentation, see if you think the same, hope that github also thinks the same, and merge

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,9 +1,8 @@
 name: Rust
 on:
-  push:
-    branches:
-      - gh-readonly-queue/main/**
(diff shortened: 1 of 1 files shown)


knoellle (approved):
LGTM.
I was a bit confused about the "types" condition at first, but it seems the only types are "checks_requested" when a PR is added to the queue and "destroyed" when one is removed.
So, only running on the former event makes sense.

PR #1009, opened 2024-05-09
Add rustdoc

Narcha:
what the title says.


knoellle (approved):
Warning: We test in production

PR #1010, opened 2024-05-09
Bump sdk version aaaaand use `clone_from`

Narcha:
what the title says

### DIFF
--- a/crates/constants/src/lib.rs
+++ b/crates/constants/src/lib.rs
@@ -14,7 +14,7 @@ pub const HULA_DBUS_SERVICE: &str = "org.hulks.hula";
 pub const HULA_SOCKET_PATH: &str = "/tmp/hula";
 pub const OS_RELEASE_PATH: &str = "/etc/os-release";
 pub const OS_VERSION: &str = "7.4.2";
-pub const SDK_VERSION: &str = "7.4.0";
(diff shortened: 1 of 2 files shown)


oleflb:
> what the title says

It does so much more than the title suggested :disappointed:

oleflb (approved):

Narcha:
Fixed it :)

crates/constants/src/lib.rs
```diff
@@ -14,7 +14,7 @@ pub const HULA_DBUS_SERVICE: &str = "org.hulks.hula";
 pub const HULA_SOCKET_PATH: &str = "/tmp/hula";
 pub const OS_RELEASE_PATH: &str = "/etc/os-release";
 pub const OS_VERSION: &str = "7.4.2";
-pub const SDK_VERSION: &str = "7.4.0";
+pub const SDK_VERSION: &str = "7.4.2";
```
schmidma, line 17: Why was this done? This leads to all HULKs downloading the newest patch version when running pepsi?! And using bumping this without also bumping the pepsi version also doesn't make sense... Please explain ;)

Narcha: This was necessary after updating the CI image. Without it, our CI couldn't find the SDK (it was looking for version 7.4.2, the latest meta-nao release). Knoelle knows the details of this.

PR #1015, opened 2024-05-11
add zero moment point calculation

MaikRe:
## Introduced Changes

Add calculation for Zero Moment Point based on my project thesis.
Make gravity a parameter.
~To filter it the low-pass filtered IMU data from the the fall state estimation is used instead of a running average.~
 
This PR only introduces the calculation, but does not use it for anything, future PRs will use the zero moment point.
This one is already in request for review in order to introduce smaller but easier to review changes.

Fixes #

## ToDo / Known Issues

- [x] Filter point location or IMU data
- [X] Output in twix panel

See #1019

## How to Test

Most easily seen using the replayer or with the help of a second person.
Let robot walk and maybe push it occasionally. The twix map panel with the walking overlay will now show a gray point for the ZMP.

### DIFF
--- a/crates/control/src/zero_moment_point_provider.rs
+++ b/crates/control/src/zero_moment_point_provider.rs
@@ -1,0 +1,69 @@
+use color_eyre::Result;
+use context_attribute::context;
+use coordinate_systems::{Ground, Robot};
+use framework::MainOutput;
+use linear_algebra::{point, vector, Point3, Rotation3};
+use serde::{Deserialize, Serialize};
+use types::sensor_data::SensorData;
+
+#[derive(Deserialize, Serialize)]
+pub struct ZeroMomentPointProvider {}
+#[context]
+pub struct CreationContext {}
+
+#[context]
+pub struct CycleContext {
+    center_of_mass: Input<Point3<Robot>, "center_of_mass">,
+    sensor_data: Input<SensorData, "sensor_data">,
+}
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {
+    pub zero_moment_point: MainOutput<Point3<Robot>>,
+}
+
+impl ZeroMomentPointProvider {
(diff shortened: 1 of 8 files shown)


crates/control/src/zero_moment_point_provider.rs
```diff
@@ -0,0 +1,79 @@
...
+        let linear_acceleration = context.robot_to_ground * self.linear_acceleration_filter.state();
+        let x_acceleration_parallel_to_ground = linear_acceleration.x();
+        let y_acceleration_parallel_to_ground = linear_acceleration.y();
+        let x_zero_moment_point_in_robot = ((x_center_of_mass * context.gravity_acceleration)
+            + (x_acceleration_parallel_to_ground * z_center_of_mass))
+            / context.gravity_acceleration;
+        let y_zero_moment_point_in_robot = ((y_center_of_mass * context.gravity_acceleration)
+            + (y_acceleration_parallel_to_ground * z_center_of_mass))
+            / context.gravity_acceleration;
+
```
oleflb, line 69: This would get rid of this lengthy conversion and is much more coordinate safe since you don't work with floats

```suggestion
let zero_moment_point = center_of_mass.xy() +  linear_acceleration.xy() * center_of_mass.z() / context.gravity_acceleration
```
(you may need to pad with an extra 0.0 for the z-component)

oleflb (changes requested):
Looks good, just one simplification for the computation

crates/control/src/zero_moment_point_provider.rs
knoellle: @olefb I think you overlooked the outer parentheses in lines 63 and 66. The original divides the sum, you only divide the right summand.
But I agree with the simplification.

MaikRe: Oh yes this is much simpler. Thank you and has been fixed

oleflb (approved):
👍

PR #1012, opened 2024-05-09
seperate  `kick_target_provider` from `kick_selector`

Vivituhh:
## Introduced Changes

Just seperates `kick_target_provider` from the `kick_selector` to add new kick targets to address #992 in future.

Fixes #

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

Should work as before.

Issue #992, opened 2024-05-04
Indirect Kick Rule


oleflb:
Ideas:
- use non-goalkicking kick decisions
- enforce striker switch after standard situations (kickoff, ...)
- new action: kickoff striker
- kick strength

oleflb:
for now we only implement an indirect kickoff and thats fine. this was implemented in #1018

PR #1018, opened 2024-05-12
Add kick off kick targets

Vivituhh:
## Introduced Changes
Adds kick targets in own kick off phase, to address the indirect kick rule #992 in a first step. The kick targets are currently located at ~1/6 and ~5/6 of the field width.
Because sometimes the newly introduces kick targets are overwritten by the instant kicks, I disabled them for the kick-off phase but they are enabled for all other states. So there should be no different behaviour in the instant kicks except kick-off. 
It also refactors the `kick_decision.rs` so that it is possible to draw kick targets/decisions even though at least one of (instant) kick targets/decisions is null. 

Based on #1012 

## ToDo / Known Issues

- [x] There might be a problem with the instant kicks, which are not needed in the moment of kick off.
- [ ] I would very much appreciate a better naming for `kick_opportunity`/`KickTargetWithKickVariants`

In the future there might also be new kick targets for kick-in, goal kick and maybe corner kick.
Moreover it might be good to step away from the ball after the kick, in very few cases the ball ended up still being inside the center circle.

## How to Test

Deploy NAO with playernumber 7 to have robot that does the kick off. Watch NAO kicking to its right or left and also take a look at the kick targets in the twix map panel (black circles with X inside).

### DIFF
--- a/crates/control/src/kick_target_provider.rs
+++ b/crates/control/src/kick_target_provider.rs
@@ -13,7 +13,7 @@ use types::{
 };
 
 #[derive(Deserialize, Serialize)]
-pub struct KickTargetProvider;
+pub struct KickTargetProvider {}
 
 #[context]
 pub struct CreationContext {}
@@ -34,6 +34,7 @@ pub struct CycleContext {
         Parameter<f32, "kick_target_provider.max_kick_around_obstacle_angle">,
 
     corner_kick_strength: Parameter<f32, "kick_target_provider.corner_kick_strength">,
+    kick_targets: AdditionalOutput<Vec<KickTarget>, "kick_targets">,
 }
 
 #[context]
@@ -48,7 +49,7 @@ impl KickTargetProvider {
         Ok(Self {})
     }
 
-    pub fn cycle(&self, context: CycleContext) -> Result<MainOutputs> {
+    pub fn cycle(&self, mut context: CycleContext) -> Result<MainOutputs> {
         let ball_position = context.ball_state.ball_in_ground;
 
         let obstacle_circles = generate_obstacle_circles(
@@ -65,6 +66,10 @@ impl KickTargetProvider {
             context.find_kick_targets,
             *context.corner_kick_strength,
         );
+
+        context
+            .kick_targets
+            .fill_if_subscribed(|| kick_targets.clone());
 
         Ok(MainOutputs {
             kick_targets: kick_targets.into(),
--- a/crates/control/src/kick_target_provider.rs
+++ b/crates/control/src/kick_target_provider.rs
@@ -19,7 +19,7 @@ use types::{
 };
 
 #[derive(Deserialize, Serialize)]
-pub struct KickTargetProvider {}
+pub struct KickTargetProvider;
 
 #[context]
 pub struct CreationContext {}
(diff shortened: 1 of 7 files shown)


crates/control/src/kick_target_provider.rs
```diff
@@ -8,12 +8,18 @@ use geometry::{circle::Circle, line_segment::LineSegment, two_line_segments::Two
...
+    kick_target::{KickTarget, KickTargetWithKickVariants},
+    motion_command::KickVariant,
+    obstacles::Obstacle,
+    parameters::FindKickTargetsParameters,
+    world_state::BallState,
 };
 
 #[derive(Deserialize, Serialize)]
-pub struct KickTargetProvider;
+pub struct KickTargetProvider {}
```
oleflb, line 22: I think `;` is more idiomatic than `{}`

crates/control/src/kick_target_provider.rs
```diff
@@ -32,15 +40,23 @@ pub struct CycleContext {
         Parameter<FindKickTargetsParameters, "kick_target_provider.find_kick_targets">,
     max_kick_around_obstacle_angle:
         Parameter<f32, "kick_target_provider.max_kick_around_obstacle_angle">,
-
     corner_kick_strength: Parameter<f32, "kick_target_provider.corner_kick_strength">,
+    kick_off_kick_strength: Parameter<f32, "kick_target_provider.kick_off_kick_strength">,
+}
+
+struct CollectKickTargetsParameter<'cycle> {
+    find_kick_targets: &'cycle FindKickTargetsParameters,
```
oleflb, line 48: also add `parameter` to the name

crates/types/src/field_dimensions.rs
```diff
@@ -35,4 +35,10 @@ impl FieldDimensions {
         position.x().abs() > self.length / 2.0 - self.goal_box_area_length
             && position.y().abs() < self.goal_box_area_width / 2.0
     }
+
+    pub fn is_inside_kick_off_target_region(&self, position: Point2<Field>) -> bool {
+        position.x().signum() == position.y().signum()
+            && position.x().abs() < self.width / 2.0
+            && position.x().abs() <= position.y().abs()
+    }
 }
```
oleflb, line 44: Im not sure this should be a method on `FieldDimensions` since its implementation seems very specific for the kickoff behavior. Maybe make this a free function instead

crates/control/src/kick_target_provider.rs
```diff
@@ -95,33 +118,82 @@ fn collect_kick_targets(
...
+        (
+            generate_kick_off_kick_targets(
+                field_dimensions,
+                field_to_ground,
+                collect_kick_targets_parameters.kick_off_kick_strength,
+            )
+            .into_iter()
+            .map(|kick_target| KickTargetWithKickVariants {
+                kick_target,
+                kick_variants: vec![KickVariant::Forward, KickVariant::Side],
```
oleflb, line 150: please make this a parameter. Why don't we want turn kicks here?

crates/control/src/kick_target_provider.rs
```diff
@@ -95,33 +118,82 @@ fn collect_kick_targets(
...
+            generate_corner_kick_targets(
+                collect_kick_targets_parameters.find_kick_targets,
+                field_dimensions,
+                field_to_ground,
+                collect_kick_targets_parameters.corner_kick_strength,
+            )
+            .into_iter()
+            .map(|kick_target| KickTargetWithKickVariants {
+                kick_target,
+                kick_variants: vec![KickVariant::Forward, KickVariant::Side, KickVariant::Turn],
```
oleflb, line 171: please also make this a parameter

crates/control/src/kick_target_provider.rs
```diff
@@ -95,33 +118,82 @@ fn collect_kick_targets(
...
+            true,
         )
     } else {
-        generate_goal_line_kick_targets(field_dimensions, field_to_ground)
+        (
+            generate_goal_line_kick_targets(field_dimensions, field_to_ground)
+                .into_iter()
+                .map(|kick_target| KickTargetWithKickVariants {
+                    kick_target,
+                    kick_variants: vec![KickVariant::Forward, KickVariant::Side, KickVariant::Turn],
```
oleflb, line 182: same here

oleflb (changes requested):

crates/control/src/kick_target_provider.rs
Vivituhh: Ok, done

crates/control/src/kick_target_provider.rs
Vivituhh: Done

crates/types/src/field_dimensions.rs
Vivituhh: Done

crates/control/src/kick_target_provider.rs
Vivituhh: I thought turn kicks are not as precise and as we have enough time at the kick off I think it's fine to joggle a bit but have a kick inside the right region.

crates/control/src/kick_target_provider.rs
Vivituhh: Done

crates/control/src/kick_target_provider.rs
Vivituhh: Done

oleflb (approved):
Noice

PR #1020, opened 2024-05-14
Twix: configurable keybindings

Narcha:
## Introduced Changes

- Twix loads a config file on startup (`.config/twix/config.toml` on Linux)
  - Path is up for debate, other idea: `.config/hulks/twix.toml`
- There is a default config which is compiled into the binary and merged with the user config (`tools/twix/default.toml`)
- Currently, the config only contains a single table `keys`, containing key bindings.
- The format for specifying keybinds is similar to that of Helix or Vim and should feel familiar to many of us. Examples:
  - `C-x`: CTRL+x
  - `C-X` or `C-S-x`: CTRL+SHIFT+x
  - `C-A-ArrowUp`: CTRL+ALT+Arrow key up
  - `C-S-Esc`: CTRL+SHIFT+Escape
- For a list of mappable keys, see [this](https://github.com/emilk/egui/blob/c1eb3f884db8bc4f52dbae4f261619cee651f411/crates/egui/src/data/key.rs#L298-L413)
- Each mappable actions is a simple enum entry, see `KeybindAction`
- In GUI code, we can now use `if context.keybind_pressed(KeybindAction::Whatever) {...}` to do something when a keybind is pressed.

Before merging, we need to get a mac user to confirm that it works on their machine.

## Example user config
```toml
# Remap navigation keys for non-vim users
[keys]
C-h = "no_op"
C-j = "no_op"
C-k = "no_op"
C-l = "no_op"

C-ArrowLeft = "focus_left"
C-ArrowDown = "focus_below"
C-ArrowUp = "focus_above"
C-ArrowRight = "focus_right"
```

## ToDo / Known Issues

It's perfect

- Add more mappable actions
- Config hot-reloading???

## How to Test

By default, all keybinds previously defined should still work the same.
Create a user config file, and play around with it.

### DIFF
--- a/tools/twix/src/configuration.rs
+++ b/tools/twix/src/configuration.rs
@@ -78,16 +78,17 @@ mod tests {
         let mut config_1: Configuration = toml::from_str(
             r#"
                 [keys]
-                "C-a" = "focus_left"
-                "C-S-a" = "reconnect"
+                C-a = "focus_left"
+                C-S-a = "reconnect"
             "#,
         )
         .unwrap();
+
         let config_2: Configuration = toml::from_str(
             r#"
                 [keys]
-                "C-c" = "focus_left"
-                "C-A" = "focus_right"
+                C-c = "focus_left"
+                C-A = "focus_right"
             "#,
         )
         .unwrap();
@@ -99,9 +100,9 @@ mod tests {
             toml::from_str(
                 r#"
                 [keys]
-                "C-a" = "focus_left"
-                "C-A" = "focus_right"
-                "C-c" = "focus_left"
+                C-a = "focus_left"
+                C-A = "focus_right"
+                C-c = "focus_left"
             "#
             )
             .unwrap()
--- a/tools/twix/src/configuration.rs
+++ b/tools/twix/src/configuration.rs
@@ -1,0 +1,24 @@
+use std::path::Path;
+
+use serde::Deserialize;
+
+mod keys;
+
+#[derive(Debug, Deserialize)]
+pub struct Configuration {
+    pub keys: keys::Keybinds,
+}
+
+impl Configuration {
+    pub fn load(path: impl AsRef<Path>) -> Result<Self, ()> {
+        let config_file = r#"
+            [keys]
+            "C-t" = "open_split"
+            "C-t" = "open_split"
+        "#;
+
+        let configuration: Configuration = toml::from_str(config_file).unwrap();
+
+        Ok(configuration)
+    }
+}
--- a/tools/twix/src/configuration.rs
+++ b/tools/twix/src/configuration.rs
@@ -1,9 +1,11 @@
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use serde::Deserialize;
 use thiserror::Error;
 
-mod keys;
+pub mod keys;
+
+const DEFAULT_CONFIG: &str = include_str!("../default.toml");
 
 #[derive(Debug, Error)]
 pub enum Error {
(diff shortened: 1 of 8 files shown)


tools/twix/src/configuration.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct Configuration {
+    pub keys: keys::Keybinds,
+}
+
+impl Configuration {
+    pub fn load() -> Result<Self, Error> {
+        Self::load_at_path(config_path())
+    }
+
+    pub fn load_at_path(path: impl AsRef<Path>) -> Result<Self, Error> {
```
knoellle, line 37:
```suggestion
    pub fn load_from_file(path: impl AsRef<Path>) -> Result<Self, Error> {
```

tools/twix/src/configuration.rs
```diff
@@ -0,0 +1,113 @@
...
+                );
+
+                Ok(Self::load_default())
+            }
+        }
+    }
+
+    fn load_default() -> Self {
+        toml::from_str(DEFAULT_CONFIG).unwrap()
+    }
```
knoellle, line 60: Make this a `Default` impl.

tools/twix/src/configuration/keys.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    pub fn parse(v: &str) -> Result<Self, Error> {
+        let parts = v.split('-').collect::<Vec<_>>();
+
+        let Some((raw_key, raw_modifiers)) = parts.split_last() else {
+            return Err(Error::InvalidKey(v.into()));
+        };
+
+        let is_single_ascii_uppercase_letter =
+            matches!(raw_key.as_bytes(), [letter] if letter.is_ascii_uppercase());
```
knoellle, line 61: Understand what do easy because good name.
Understand how do hard since use black magic.
Unsure how make better and no make code very verbose.

tools/twix/src/configuration/keybind_plugin.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use eframe::egui::{Context, Id, InputState};
+
+use super::keys::{KeybindAction, Keybinds};
+
+type ActionList = Arc<Vec<KeybindAction>>;
+
+pub fn register(ctx: &Context) {
+    ctx.on_begin_frame("keybinds", Arc::new(begin_frame))
+}
```
knoellle, line 11: I very much like this "plugin" way of doing things.
Resolve if read.

tools/twix/src/configuration/keys.rs
```diff
@@ -0,0 +1,233 @@
...
+pub enum KeybindAction {
+    CloseTab,
+    DuplicateTab,
+    FocusAbove,
+    FocusAddress,
+    FocusBelow,
+    FocusLeft,
+    FocusPanel,
+    FocusRight,
+    NoOp,
```
knoellle, line 32: What is this for?

tools/twix/src/configuration/keys.rs
```diff
@@ -0,0 +1,233 @@
...
+        let mut modifiers = Modifiers {
+            shift: is_single_ascii_uppercase_letter,
+            ..Modifiers::NONE
+        };
+
+        for raw_modifier in raw_modifiers {
+            let modifier = KeybindTrigger::parse_modifier(raw_modifier)?;
+
+            if modifiers.contains(modifier) {
+                return Err(Error::DuplicateModifier(String::from(*raw_modifier)));
```
knoellle, line 76: This may be confusing when doing `S-T`.
But may be fine for now unless you have a great idea how to handle it better.

knoellle (changes requested):

tools/twix/src/configuration/keys.rs
Narcha: Unbinding keys defined in the default config. Could be removed if we decide not to merge keybinds with the defaults

tools/twix/src/configuration/keybind_plugin.rs
```diff
@@ -0,0 +1,61 @@
...
+        let actions = ctx.input_mut(|input| consume_actions(keybinds, input));
+
+        ctx.data_mut(|data| data.insert_temp::<ActionList>(Id::NULL, Arc::new(actions)))
+    }
+}
+
+fn consume_actions(keybinds: Arc<Keybinds>, input: &mut InputState) -> Vec<KeybindAction> {
+    let mut actions = Vec::new();
+
+    input.events.retain(|event| {
```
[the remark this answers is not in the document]
Narcha: Oh, wow... Thanks for figuring this out. I'd be in favor of just hard-erroring when trying to bind these, I suspect most people would not want to bind them anyway

tools/twix/src/configuration/keys.rs
Narcha: This was intentional, I tried to match helix' behavior. I'm open for a debate on what we want in this case.

tools/twix/src/configuration.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self::load_at_path(config_path())
+    }
+
+    pub fn load_at_path(path: impl AsRef<Path>) -> Result<Self, Error> {
+        match std::fs::read_to_string(&path) {
+            Ok(config_file) => {
+                let mut configuration = Self::load_default();
+                let user_configuration: Configuration = toml::from_str(&config_file)?;
+
+                configuration.merge(user_configuration);
```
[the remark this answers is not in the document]
Narcha: I tried matching helix' behavior here. This way, user configs can be relatively small if only a few keys are remapped, and you don't have to copy the entire default config (and keep it up-to-date with new features!) in order to remap single keys. I'm open for a debate here, it might be worth asking other team members what they think about this.

tools/twix/src/configuration/keys.rs
Narcha: What would [grug](https://grugbrain.dev/) do in this situation? :thinking:

tools/twix/src/configuration.rs
Narcha: fixed in 722b5c343

tools/twix/src/configuration.rs
Narcha: fixed in f8a27e0d6

tools/twix/src/configuration/keybind_plugin.rs
```diff
@@ -0,0 +1,61 @@
...
+        if let eframe::egui::Event::Key {
```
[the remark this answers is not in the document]
Narcha: fixed in 6c74daf74

tools/twix/src/main.rs
```diff
@@ -93,13 +97,15 @@ fn main() -> Result<(), eframe::Error> {
             .unwrap();
     }
 
+    let configuration = Configuration::load().expect("failed to load configuration");
```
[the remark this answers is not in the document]
Narcha: fixed in 6e44c1142

tools/twix/src/configuration.rs
```diff
@@ -0,0 +1,113 @@
...
+pub enum Error {
+    #[error("IO error: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("TOML parsing error: {0}")]
+    Parsing(#[from] toml::de::Error),
+}
+
+fn config_path() -> PathBuf {
+    let mut result = dirs::config_dir().unwrap();
+    result.extend(["twix", "config.toml"]);
```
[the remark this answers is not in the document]
Narcha: fixed in fe57b13f1

tools/twix/src/configuration/keybind_plugin.rs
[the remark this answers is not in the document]
Narcha: fixed in 911073517

tools/twix/src/configuration.rs
```diff
@@ -0,0 +1,113 @@
...
+        assert_eq!(
+            config_1,
+            toml::from_str(
+                r#"
+                [keys]
+                C-a = "focus_left"
+                C-A = "focus_right"
+                C-c = "focus_left"
+            "#
+            )
```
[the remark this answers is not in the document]
Narcha: fixed in dc9b3500a

tools/twix/src/configuration.rs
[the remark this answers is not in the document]
knoellle: Alright, @schmidma agrees with the current approach.
Gekauft.

tools/twix/src/configuration/keys.rs
```diff
@@ -0,0 +1,259 @@
...
+use serde::{
+    de::{self, Deserializer},
+    Deserialize,
+};
+use thiserror::Error;
+
+#[cfg_attr(test, derive(PartialEq))]
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Duplicate modifier `{0}`")]
```
schmidma, line 13: @Narcha For future PRs: please keep all error messages lower-case ;)

PR #1022, opened 2024-05-15
Reduce visural referee false positives during setup

### LINKED ISSUE #999: Deny False Positives in Visual Referee Detection
- [ ] use horizon check
- [ ] use height check
- [x] use keypoint confidence
- ~maybe check "redness" at hand keypoints (remember the referee is wearing gloves)~

ThagonDuarte:
## Why? What?

During setup of NAOs, i.e. when carrying them to the field and temporary stiffing them, a lot of false positives in the visual referee pose detection occurred. This PR resets the detection times, which are used for the decision to go to Ready.
Previous changes of this PR have been moved to #1093.

Fixes #999 partially 

## How to Test

- Setup a NAO for visual referee
- In twix subscribe `detection_times` and see if it is reset to default when not in Standby

### DIFF
--- a/crates/control/src/behavior/initial.rs
+++ b/crates/control/src/behavior/initial.rs
@@ -36,11 +36,11 @@ fn look_at_referee(
     expected_referee_position: Option<Point2<Field>>,
     world_state: WorldState,
 ) -> Option<MotionCommand> {
-    let ground_to_field = world_state.robot.ground_to_field?;
-    let expected_referee_position = expected_referee_position?;
     if world_state.filtered_game_controller_state?.game_state != FilteredGameState::Standby {
         return None;
     }
+    let ground_to_field = world_state.robot.ground_to_field?;
+    let expected_referee_position = expected_referee_position?;
 
     let position = ground_to_field.as_pose().position();
 
--- a/crates/control/src/motion/look_at.rs
+++ b/crates/control/src/motion/look_at.rs
@@ -29,6 +29,8 @@ pub struct CreationContext {}
 
 #[context]
 pub struct CycleContext {
+    last_motion_command: CyclerState<MotionCommand, "last_motion_command">,
+
     camera_matrices: Input<Option<CameraMatrices>, "camera_matrices?">,
     cycle_time: Input<CycleTime, "cycle_time">,
     ground_to_robot: Input<Option<Isometry3<Ground, Robot>>, "ground_to_robot?">,
@@ -57,6 +59,8 @@ impl LookAt {
     }
 
     pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+        *context.last_motion_command = context.motion_command.clone();
+
         let cycle_start_time = context.cycle_time.start_time;
         let measured_head_angles = context.sensor_data.positions.head;
         let default_output = Ok(MainOutputs {
--- a/crates/object_detection/src/pose_detection.rs
+++ b/crates/object_detection/src/pose_detection.rs
@@ -22,6 +22,7 @@ use serde::{Deserialize, Serialize};
 use types::{
     bounding_box::BoundingBox,
     color::Rgb,
+    cycle_time::CycleTime,
     motion_command::MotionCommand,
     pose_detection::{HumanPose, Keypoints},
     ycbcr422_image::YCbCr422Image,
@@ -54,14 +55,14 @@ pub struct CycleContext {
     inference_duration: AdditionalOutput<Duration, "inference_duration">,
     postprocess_duration: AdditionalOutput<Duration, "postprocess_duration">,
 
+    cycle_time: Input<CycleTime, "cycle_time">,
     image: Input<YCbCr422Image, "image">,
     motion_command: Input<MotionCommand, "Control", "motion_command">,
 
     intersection_over_union_threshold:
-        Parameter<f32, "object_detection.$cycler_instance.intersection_over_union_threshold">,
-    keypoint_confidence_threshold:
-        Parameter<f32, "object_detection.$cycler_instance.keypoint_confidence_threshold">,
-    enable: Parameter<bool, "object_detection.$cycler_instance.enable">,
+        Parameter<f32, "pose_detection.intersection_over_union_threshold">,
+    keypoint_confidence_threshold: Parameter<f32, "pose_detection.keypoint_confidence_threshold">,
+    enable: Parameter<bool, "pose_detection.enable">,
 }
 
 #[context]
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1,6 +1,5 @@
 {
-  "object_detection": {
-    "object_detection_top": {
+  "pose_detection": {
       "enable": false,
       "intersection_over_union_threshold": 0.45,
       "distance_to_referee_position_threshold": 2.0,
@@ -9,8 +8,7 @@
       "foot_z_offset": 0.05,
       "referee_pose_queue_length": 8,
       "minimum_number_poses_before_message": 5
-    }
-  },
+    },
   "feet_detection": {
     "vision_top": {
       "enable": true,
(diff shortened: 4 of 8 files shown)


crates/control/src/behavior/initial.rs
```diff
@@ -34,10 +36,13 @@ fn look_at_referee(
     ground_to_field: Option<Isometry2<Ground, Field>>,
     filtered_game_controller_state: Option<FilteredGameControllerState>,
     expected_referee_position: Option<Point2<Field>>,
+    enable_pose_detection: bool,
 ) -> Option<MotionCommand> {
     let ground_to_field = ground_to_field?;
     let expected_referee_position = expected_referee_position?;
-    if filtered_game_controller_state?.game_state != FilteredGameState::Initial {
+    if !enable_pose_detection
+        && filtered_game_controller_state?.game_state != FilteredGameState::Initial
```
schmidma, line 44: Shouldn't this be 
```suggestion
        || filtered_game_controller_state?.game_state != FilteredGameState::Initial
```
???

crates/control/src/behavior/node.rs
```diff
@@ -70,6 +70,7 @@ pub struct CycleContext {
     intercept_ball_parameters: Parameter<InterceptBallParameters, "behavior.intercept_ball">,
     maximum_step_size: Parameter<Step, "step_planner.max_step_size">,
     striker_set_position: Parameter<Point2<Field>, "behavior.role_positions.striker_set_position">,
+    enable_pose_detection: Parameter<bool, "object_detection.object_detection_top.enable">,
```
schmidma, line 73: Suggestion for a more fitting name?
```suggestion
    enable_pose_detection: Parameter<bool, "pose_detection.enable">,
```

crates/control/src/motion/look_at.rs
```diff
@@ -29,6 +29,8 @@ pub struct CreationContext {}
 
 #[context]
 pub struct CycleContext {
+    last_motion_command: CyclerState<MotionCommand, "last_motion_command">,
```
schmidma, line 32: Can you please elaborate on the circular dependency, you are trying to solve here?
I am not sure whether this is actually the best way to solve this... Nevertheless, this field should definitely not be produced in `look_at`.

crates/object_detection/src/pose_detection.rs
```diff
@@ -222,6 +222,36 @@ impl PoseDetection {
             human_poses: poses.into(),
         })
     }
+
+    fn should_look_for_referee<T: TimeInterface>(
+        &mut self,
+        hardware_interface: &Arc<T>,
```
schmidma, line 228: you are only using the hardware interface to get the time. Do that once and pass the actual `now` to the function. You should maybe look into using `CycleInfo` instead anyway.

crates/object_detection/src/pose_detection.rs
```diff
@@ -222,6 +222,36 @@ impl PoseDetection {
...
+        motion_command: &MotionCommand,
+        pose_detection_inactive_duration: &Duration,
+    ) -> bool {
+        let is_in_initial_with_pose_detection = matches!(
```
schmidma, line 232:
```suggestion
        let should_look_for_referee = matches!(
```
or
```suggestion
        let behavior_requests_pose_detection = matches!(
```

crates/object_detection/src/pose_detection.rs
```diff
@@ -222,6 +222,36 @@ impl PoseDetection {
...
+                ..
+            }
+        );
+
+        if is_in_initial_with_pose_detection && !self.pose_detection_was_started {
+            self.start_time = hardware_interface.get_now();
+            self.pose_detection_was_started = true;
+        } else if !is_in_initial_with_pose_detection {
+            self.start_time = hardware_interface.get_now();
+        }
```
schmidma, line 245: Why do you have to distinguish these cases?

crates/object_detection/src/pose_detection.rs
```diff
@@ -222,6 +222,36 @@ impl PoseDetection {
...
+
+        let inactive_period_as_elapsed = hardware_interface
+            .get_now()
+            .duration_since(self.start_time)
+            .ok()
+            > Some(*pose_detection_inactive_duration);
```
schmidma, line 251: huh?! this is possible? You are ordering options here? What happens if it is `None`?

crates/object_detection/src/pose_detection.rs
```diff
@@ -222,6 +222,36 @@ impl PoseDetection {
...
+        );
+
+        if is_in_initial_with_pose_detection && !self.pose_detection_was_started {
+            self.start_time = hardware_interface.get_now();
+            self.pose_detection_was_started = true;
+        } else if !is_in_initial_with_pose_detection {
+            self.start_time = hardware_interface.get_now();
+        }
+
+        let inactive_period_as_elapsed = hardware_interface
```
schmidma, line 247: Did you mean
```suggestion
        let inactive_period_is_elapsed = hardware_interface
```

schmidma (changes requested):
some questions and remarks...

knoellle:
Is this PR still relevant since #1058 was merged?

MaikRe:
Further superseded by #1086

crates/control/src/behavior/initial.rs
ThagonDuarte: You are absolutely right. Done, but this change as well as the move of the pose detection toggle is now tracked in #1093.

crates/control/src/behavior/node.rs
ThagonDuarte: Done

crates/control/src/motion/look_at.rs
ThagonDuarte: As it is currently implemented, the production of the `MotionCommand::Initial` with the field `should_look_for_referee` is used as the behavioral trigger to do pose detection. I want to use this field to check whether or not pose detection is enabled in behavior. The motion command is produced in motion which is dependent on behavior, hence the circular dependency. 
Now that I think about it, this decision of whether or not pose detection should run should be done in behavior, which would resolve this issue.

crates/object_detection/src/pose_detection.rs
ThagonDuarte: Done

crates/object_detection/src/pose_detection.rs
ThagonDuarte: Done. But is again now tracked in #1093.

crates/object_detection/src/pose_detection.rs
ThagonDuarte: This feature has been removed, as the `Standby` state solves this.

crates/object_detection/src/pose_detection.rs
ThagonDuarte: It is. A `None` is always smaller than any other value. But is also removed, as mentioned above.

crates/object_detection/src/pose_detection.rs
ThagonDuarte: Same as before, removed...

etc/parameters/default.json
```diff
@@ -6,7 +6,11 @@
       "distance_to_referee_position_threshold": 2.0,
       "keypoint_confidence_threshold": 0.2,
       "shoulder_angle_threshold": 0.2,
-      "foot_z_offset": 0.05
+      "foot_z_offset": 0.05,
+      "pose_detection_inactive_duration": {
+        "nanos": 0,
+        "secs": 7
```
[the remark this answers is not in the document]
ThagonDuarte: Here as well, removed...

ThagonDuarte:
I have now removed any unnecessary features from this PR that have been solved by the Standby state or are better tracked in another PR. The remaining feature of resetting detection when not in Standby is however still relevant in my opinion.

ThagonDuarte:
Nevermind, perhaps we don't need the resetting of the detection times either, because of the Standby state.

PR #1027, opened 2024-05-19
Do not include SDK in CI image

### LINKED ISSUE #1025: Pin `SDK_VERSION` for the CI image action
The CI image action automatically downloads the *latest* release. We should pin that version.

This might have implications on our updating mechanisms, but we can discuss that

### LINKED ISSUE #1026: Evaluate publishing a container every CI run
maybe use something like

```
buildah build --layers --build-arg SDK_VERSION=7.4.2 --tag hulks-ci .
```

maybe even using `--cache-from` and `--cache-to`

schmidma:
## Why? What?

- Removes the SDK from the CI container image.

The SDK is properly installed via pepsi in the CI run, but cached in the runners cache. This allows for a much smoother update procedure, when upgrading the CI. Each PR runs with the SDK version specified in the constants crate.

- This PR also removes all the old stuff of former ci runners...

- Upgrade to `checkout@v4`

Closes  #1026
Closes  #1025

## ToDo / Known Issues

- ~~publish a proper `latest` ci image on the registry, as soon as this PR is about to be merged.~~

none

## How to Test

look at the code, and reason whether this should work :D The CI should be green, but this is a requirement anyway


tools/ci/github-runners/v3/Dockerfile
```diff
@@ -1,40 +0,0 @@
...
-ARG SDK_VERSION
-RUN wget --no-verbose https://github.com/HULKs/meta-nao/releases/download/${SDK_VERSION}/HULKs-OS-toolchain-${SDK_VERSION}.sh && \
-  chmod +x HULKs-OS-toolchain-${SDK_VERSION}.sh && \
-  ./HULKs-OS-toolchain-${SDK_VERSION}.sh -y -d /ci/.naosdk/${SDK_VERSION} && \
-  rm HULKs-OS-toolchain-${SDK_VERSION}.sh
-
-COPY cargo.sh.patch /ci
-RUN cd /ci/.naosdk/${SDK_VERSION}/sysroots/x86_64-naosdk-linux/environment-setup.d && \
-  patch -i /ci/cargo.sh.patch && \
-  rm /ci/cargo.sh.patch
```
h3ndrk, line 40: Is the `cargo.sh.patch` still being used? It looks like it is dead code currently. :thinking:

tools/ci/github-runners/README.md
h3ndrk, line 1: What was the reasoning for dropping the version? Did you also change the runner labels? I introduced that once because we wanted to gradually upgrade the CI and keep the old runners running for old PRs.

knoellle: Do we need differently tagged runners if we can just use differently tagged images?
That way all runners can run jobs for old & new PRs, though the tag from this PR should probably not be called `no-sdk`, but `v4` or something.

Also, I'm not convinced we need versioning within our file tree. That's what git is for.

tools/ci/github-runners/v3/Dockerfile
schmidma: no it is not. Removed it...

tools/ci/github-runners/README.md
schmidma: I agree with konrad about this. If we have versioning, we should not keep them in our worktree in parallel, but migrate to the next version, and rolling back is done with git.

Regarding the versioning, we should consider that there are two different things, we have to version. One is the runner environment, that is currently tagged with v3. That is the way the runner is setup, etc. For example if it is a windows or linux runner, or how it is configured locally. We tag that with `vX` to distinguish which jobs are compatible with what type of environment.

The other thing we should version is the CI image, that is used for all our jobs and provides rust and other dependencies. This is currently not versioned, and we use latest. We may want to introduce proper versioning here, but I doubt, that we actually upgrade that image very often, as it is not including the SDK anymore...

tools/ci/github-runners/v3/Dockerfile
h3ndrk: Did you check whether it was still used or did you check what it did and decided that it is obsolete? Vote for an intentional instead of coincidental decision. If the latter, feel free to resolve.

tools/ci/github-runners/README.md
h3ndrk: Since we manually deploy our runners, there may be times where the deployment is out of sync with our repository. Only having a rolling release version in our repository might not help. But since these situations became rare, we can be pragmatic.

Regarding runner tagging: I think, starting with this PR we will loose the connection between runner tag and repository version. It will not be clear, at which point we for example "released" `v4`. But also here we can be pragmatic. Just wanting to point out potential future issues.

Same pragmatism with the CI image version probably.

tools/ci/github-runners/v3/Dockerfile
schmidma: This is intentional: This patch previously moved the cargo home to something outside the container to persist and cache the registry. As the entire SDK now lives outside the container, also the default cargo home at `$OECORE_TARGET_SYSROOT/home/cargo` is outside the container.

tools/ci/github-runners/README.md
schmidma: I conclude:

- Remove runner tagging and versioning of runner environments for now. We can easily add a `runs-on` entry to the ci actions when we want to run in a potential future runner environment.
- introduce tagged CI images and version them to pin the version in the individual actions and not have them implicitly rely on `latest`. This makes the upgrade workflow of ci images much easier, as we can have multiple different ci images used in parallel

Agreed?

tools/ci/github-runners/v3/Dockerfile
schmidma: I even think,  it is more correct now, as the cargo homes are not shared between different SDK versions.

tools/ci/github-runners/README.md
schmidma: The main reason to upgrade the CI image after this PR is merged, is to upgrade the Rust version, we can think of maybe reflecting this in the tagging of our CI images?

knoellle: > - Remove runner tagging and versioning of runner environments for now. We can easily add a `runs-on` entry to the ci actions when we want to run in a potential future runner environment.

Can we? If we add a new runner with a different environment and a new tag, does that prevent PRs that don't have this tag in their workflow yaml yet from running on that new runner?

If I read the [docs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow) correctly, the routing requires label presences, not absence of other labels.

So when introducing a new runner/environment, that would have to not have one of the labels that all the reuglar jobs on the main branch want to run on (exept if it also supports the old jobs).

h3ndrk (approved):

PR #1029, opened 2024-05-22
Smaller Locking Scopes of HuLA interfacing

schmidma:
## Why? What?

the `.get_now()` accidentally created a barrier between all cyclers...
We moved the Mutex scopes to a lower level, locking the respective parts for sending, receiving, or getting time of the HuLA wrapper.

## ToDo / Known Issues

none

none

## How to Test

Record with replay, look at the timestamps in the pretty replayer tooling, the timestamps are now no longer aligned to the control cycle (compare to `main`, which has all cycle starts aligned to control`...)

### DIFF
--- a/crates/hulk_nao/src/hardware_interface.rs
+++ b/crates/hulk_nao/src/hardware_interface.rs
@@ -53,7 +53,7 @@ pub struct Parameters {
 }
 
 pub struct HardwareInterface {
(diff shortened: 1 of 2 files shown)


h3ndrk (approved):
Werkelt wie im Katalog beworben

PR #1042, opened 2024-05-30
RNG Determinism

knoellle:
## Why? What?

What it says.

Why? To make replays more useful.
Currently nodes like line detection and field border detection run use non-deterministic randomness to generate their output.
Thus, when replaying the same vision frame twice, we will likely get two different sets of detected lines.
The random generator state is now stored inside the nodes, making them fully replayable.

Also makes replayer error messages more useful, instead of "failed to replay frame", it will now print the actual root cause.

Fixes #

## How to Test

Record data, replay it. 
Look at lines in the image panel, they should no longer jump around when replaying the same vision frame.

### DIFF
--- a/crates/ransac/src/lib.rs
+++ b/crates/ransac/src/lib.rs
@@ -1,8 +1,7 @@
 use geometry::line::{Line, Line2};
 use linear_algebra::Point2;
 use ordered_float::NotNan;
-use rand::seq::SliceRandom;
-use rand_chacha::ChaChaRng;
+use rand::{seq::SliceRandom, Rng};
 
 #[derive(Default, Debug, PartialEq)]
 pub struct RansacResult<Frame> {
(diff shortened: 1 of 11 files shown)


crates/ransac/src/lib.rs
```diff
@@ -11,22 +12,18 @@ pub struct RansacResult<Frame> {
...
-        }
+    pub fn new(unused_points: Vec<Point2<Frame>>) -> Ransac<Frame> {
+        Ransac { unused_points }
     }
 }
 
 impl<Frame> Ransac<Frame> {
     pub fn next_line(
         &mut self,
+        random_number_generator: &mut ChaChaRng,
```
schmidma, line 26:
```suggestion
        random_number_generator: &mut impl rand::Rng,
```

schmidma (changes requested):
one minor thing...

crates/ransac/src/lib.rs
knoellle: Döne

schmidma (approved):

PR #1053, opened 2024-06-10
Robustify SDK installation

### LINKED ISSUE #1052: Interrupting SDK install step in CI breaks  CI
When a CI step is interruped while it is installing a SDK, it may result in a partial SDK.
This is fixed by deleting the broken SDK and ensuring that the SDK install step runs through successfully.
Maybe we could add a seperate sdk install step in the CI that is not interruptable. Alternatively, we could disable `fail-fast` entirely for the build step

knoellle:
## Why? What?

SDK installation leads to an inconsistent state when interrupted, see linked issue.
This PR solves the issue by creating a marker file before installing the sdk and only removing the marker again when installation finishes successfully. Thus, a failed or aborted installation will not be mistaken for a successful one because the "is already installed check" fails if the marker file is present.

Fixes #1052 

## How to Test

0. Make sure you don't have the most recent sdk installed (remove or rename the 7.4.2 directory)
1. Begin sdk installation `./pepsi sdk install`
2. Abort sdk installation during extraction
3. Try to build for nao `./pepsi build --target nao`

On main this should fail due to the sdk installation being incomplete.
With this change, the command in step 3 will prompt for sdk installation again.

### DIFF
--- a/crates/repository/src/lib.rs
+++ b/crates/repository/src/lib.rs
@@ -339,6 +339,7 @@ impl Repository {
                     .wrap_err("failed to download SDK")?;
             }
             if sdk.exists() {
+                println!("Removing incomplete SDK directory...");
                 remove_dir_all(&sdk)
                     .await
                     .wrap_err("failed to remove old SDK directory")?;
--- a/crates/repository/src/lib.rs
+++ b/crates/repository/src/lib.rs
@@ -327,7 +327,9 @@ impl Repository {
             directory
         };
         let sdk = installation_directory.join(version);
-        if !sdk.exists() {
+        let incomplete_marker = installation_directory.join(format!("{version}.incomplete"));
+        let need_to_install = !sdk.exists() || incomplete_marker.exists();
+        if need_to_install {
             let downloads_directory = installation_directory.join("downloads");
             let installer_name = format!("HULKs-OS-toolchain-{version}.sh");
             let installer_path = downloads_directory.join(&installer_name);
@@ -336,9 +338,15 @@ impl Repository {
                     .await
                     .wrap_err("failed to download SDK")?;
             }
+            File::create(&incomplete_marker)
+                .await
+                .wrap_err("failed to create marker")?;
             install_sdk(installer_path, &sdk)
                 .await
                 .wrap_err("failed to install SDK")?;
+            remove_file(&incomplete_marker)
+                .await
+                .wrap_err("failed to remove marker")?;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


schmidma:
I think we should tell the user that the SDK at the install location is incomplete. Otherwise the user is prompted with `The directory "/home/maximilian/worktree/hulks/hulk/naosdk/7.4.2" already contains a SDK for this architecture.`, and this does not sound like an error.

Furthermore, are we sure that just installing an SDK over an existing one actually fixes the problem? A cleaner approach would be to purge the entire incomplete SDK before trying another installation

knoellle:
Done, removing old sdk directory before attempting reinstallation.

schmidma:
Do you think it is fine to silently remove it? or should we inform the user or even wait for an acknowledgement? What do you think?

knoellle:
Yes, I think its fine.
If installation fails because of an error, pepsi shows an error message anyway.
If there was a risk of repeatedly trying to install a new sdk on every pepsi invocation, I agree that we'd need an additional message.
But I don't see how that could happen.

schmidma:
Hmm... I am not entirely sure. Can we please just add an informational message like

```
removing an incomplete SDK installation for version VERSION
```

or something similar.

crates/repository/src/lib.rs
```diff
@@ -336,9 +338,22 @@ impl Repository {
                     .await
                     .wrap_err("failed to download SDK")?;
             }
+            if sdk.exists() {
+                println!("Removing incomplete SDK directory...");
+                remove_dir_all(&sdk)
+                    .await
+                    .wrap_err("failed to remove old SDK directory")?;
+            }
```
schmidma, line 346: Can we move this block outside the `if need_to_install` and guard it with `incomplete_marker.exists()`? If we then remove the directory first, the "old" check for `needs_to_install` by checking whether there is an SDK suffices, as we removed it...

knoellle: Good point, this seems cleaner.

crates/repository/src/lib.rs
```diff
@@ -327,9 +327,16 @@ impl Repository {
             directory
         };
         let sdk = installation_directory.join(version);
+
         let incomplete_marker = installation_directory.join(format!("{version}.incomplete"));
-        let need_to_install = !sdk.exists() || incomplete_marker.exists();
-        if need_to_install {
+        if sdk.exists() && incomplete_marker.exists() {
```
schmidma, line 332:
```suggestion
        if incomplete_marker.exists() {
```
is this sufficient?

schmidma (approved):

PR #1054, opened 2024-06-12
Add support for building for the NAO with docker

julianschuler:
## Why? What?

This PR adds support for building for the NAO within a docker container. This is allows for building on non-linux systems like macOS.

## ToDo / Known Issues

None.

None.

## How to Test

To test this under Linux, you additionally have to change `OS_IS_NOT_LINUX` to true in `crates/repository/src/lib.rs`.

Afterwards, you can build for the NAO:

```
./pepsi build --target=nao
```


schmidma:
As discussed with @larskna I disagree with the proposed solution. We should only do the bare minimum in the linux container. Everything except the cross-compiling works natively on MacOS, so we should run it on the native machine

julianschuler:
I changed the implementation now to keep the Dockerfile to the bare minimum and only execute the SDK installation and all commands requiring the SDK within the docker container. Currently the feature is enabled at compile time for all non-linux operating systems. Feel free to discuss whether this should be an explicit flag (maybe even an explicit flag with a OS-dependent default?).

crates/repository/src/lib.rs
```diff
@@ -39,6 +39,7 @@ use tokio::{
 use spl_network_messages::PlayerNumber;
 
 const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const USE_DOCKER: bool = !cfg!(target_os = "linux");
```
schmidma, line 42:
```suggestion
const OS_IS_NOT_LINUX: bool = !cfg!(target_os = "linux");
```

crates/repository/src/lib.rs
```diff
@@ -139,26 +140,18 @@ impl Repository {
...
-        }
-
         let cargo_command = format!("cargo {action} ")
             + format!("--profile {profile} ").as_str()
             + if workspace {
                 "--workspace --all-features --all-targets ".to_string()
             } else {
                 let manifest = format!("crates/hulk_{target}/Cargo.toml");
-                format!("--manifest-path={} ", self.root.join(manifest).display())
+                let root = if target == "nao" && USE_DOCKER {
```
schmidma, line 149:
```suggestion
                let root = if use_docker {
```

and introduce the `use_docker` variable accordingly

crates/repository/src/lib.rs
```diff
@@ -326,6 +341,11 @@ impl Repository {
             create_symlink(&directory, &symlink).await?;
             directory
         };
+
+        if USE_DOCKER {
+            return Ok(());
+        }
```
schmidma, line 347: I'd rather skip the invocation of this function instead of skipping parts of the function. But this may have to be adjusted as soon as we have a proper naosdk container image

schmidma (changes requested):
I'd suggest to move this back to "in progress" as long as the respective meta-nao is not merged

crates/repository/src/lib.rs
julianschuler: Thanks, done.

crates/repository/src/lib.rs
julianschuler: Thanks, done.

crates/repository/src/lib.rs
julianschuler: We still need to create the symlink, maybe factoring out the symlink creation to a separate funtion instead? FYI: There is already a container image published with the same semantic as the final one, so this is already testable.

crates/repository/src/lib.rs
```diff
@@ -170,9 +164,32 @@ impl Repository {
 
         println!("Running: {cargo_command}");
 
+        let shell_command = if target == "nao" {
+            if OS_IS_NOT_LINUX {
```
schmidma, line 168: move the `use_docker` variable to its parent scope and use it here. This branch should depend on whether we want to use docker, and not whether the OS is not linux. I imagine actively selecting for the docker build in some future feature

crates/repository/src/lib.rs
```diff
@@ -139,26 +140,19 @@ impl Repository {
...
             } else {
+                let use_docker = target == "nao" && OS_IS_NOT_LINUX;
                 let manifest = format!("crates/hulk_{target}/Cargo.toml");
-                format!("--manifest-path={} ", self.root.join(manifest).display())
+                let root = if use_docker {
+                    Path::new("/hulk")
+                } else {
+                    &self.root
+                };
+                format!("--manifest-path={} ", root.join(manifest).display())
```
schmidma, line 155: The entire control flow here is now very confusing and complex. In some future iteration, I can think of a fancy builder syntax using type-states leading to compiletime-typesafe construction of the respective commands...

Resolve this conversation if read, maybe open an issue :shrug:

crates/repository/src/lib.rs
julianschuler: Thanks, done.

crates/repository/src/lib.rs
julianschuler: I split this now into two separate functions.

crates/repository/src/lib.rs
```diff
@@ -304,13 +320,28 @@ impl Repository {
...
         version: Option<&str>,
         installation_directory: Option<&Path>,
     ) -> Result<()> {
+        let installation_directory = self
+            .link_sdk_home(installation_directory)
+            .await
+            .wrap_err("failed to link SDK home")?;
+
+        let use_docker = OS_IS_NOT_LINUX;
+        if !use_docker {
```
schmidma, line 334: Please move this to the call site. The caller should decide what to do. That is, the "normal" install caller now calls two functions, the docker way is only calling the link function. Currently, this function hides a lot of semantic not visible to the caller in any way.

julianschuler: I decided against this to avoide the inherent code duplication, should I still inline this?

schmidma (approved):

PR #1055, opened 2024-06-13
Add support for aach64 SDK

julianschuler:
## Why? What?

This PR adds support for using aarch64 SDKs.

## ToDo / Known Issues

We currently don't have any linux aarch64 users, so testing this might be difficult.

None

## How to Test

```sh
./pepsi build --target nao
```

### DIFF
--- a/crates/repository/src/lib.rs
+++ b/crates/repository/src/lib.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
-    env::{self, current_dir},
+    env::{self, consts::ARCH, current_dir},
     ffi::OsStr,
     fmt::Display,
     fs::Permissions,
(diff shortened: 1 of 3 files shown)


crates/repository/src/lib.rs
```diff
@@ -40,6 +40,16 @@ use spl_network_messages::PlayerNumber;
 
 const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 
+#[cfg(not(target_os = "linux"))]
+const USE_DOCKER: bool = true;
```
oleflb, line 44: What do you think of using cfg! directly here?

crates/repository/src/lib.rs
```diff
@@ -40,6 +40,16 @@ use spl_network_messages::PlayerNumber;
...
+#[cfg(target_os = "linux")]
+const USE_DOCKER: bool = false;
+
+#[cfg(target_arch = "x86_64")]
+const TARGET_ARCH: &str = "x86_64";
+#[cfg(target_arch = "aarch64")]
+const TARGET_ARCH: &str = "aarch64";
```
oleflb, line 51: Use std::consts::env::ARCH directly?

oleflb (changes requested):
Not an extensive review yet.

crates/repository/src/lib.rs
julianschuler: This PR now no longer depends on #1054, this is now removed.

crates/repository/src/lib.rs
julianschuler: Good catch, thanks!

oleflb (approved):
This surely looks much easier :D

PR #1056, opened 2024-06-14
make catching steps toggleable

MaikRe:
## Why? What?

#887 Disabled catching steps by means of commenting out code and modifying parameters to adjust for a earlier needed fall protection.
This PR allows for catching steps to be toggled on and off based on a single parameter.
Additionally it introduces the old fall angle threshold to be used when catching steps are enabled.

This is done both for code cleanliness and useability of catching steps, as well as a prerequisite for #1019 so that these new catching steps can just as easily be disabled and enabled when desired.

Fixes #

## How to Test

Change the new ```use_catching_steps``` parameter between true and false and observe the presence or absence of catching steps


schluis (approved):
Nice

PR #1058, opened 2024-06-17
Standby phase

### LINKED ISSUE #980: GameState Setup
>Dear Teams, a small heads up: due to feedback from the German Open we will be introducing a new GameState in the rules >and GameController before the 2024 World Championship.
>
>The new GameStatse will likely be called "Setup" and happen before the first Initial of a match. It gives teams time to set up ?>on the side on the field. It will be the same as Initial (thus, easy to implement), but the robots will not be penalized for the >"Motion in Initial" penalty.
>
>We are changing the rules soon and will publish a new GameController (thanks to AH). We will notify as soon as the details >are set in stone.

### LINKED ISSUE #983: Adapt Visual Referee to Setup Game State
What the title says.

MaikRe:
## Why? What?

Implementation of the new standby phase according to the GameControllerv3 version 18.
We need this in order to compete at RC24.

Also renames the ```is_referee_initial_pose_detected``` variable to ```is_referee_ready_pose_detected``` to more accurately describe what this variable does

Fixes #980
Fixes #983  

## How to Test

Using the newest game controller, do a normal test game.
During the initial phase the robots should not look for the referee nor respond to visual referee signals.
During the standby phase the robots should look towards the referee T-intersection and respond to visual referee signals.


oleflb (approved):
Tested in testgame 2024-06-19, works like a charm. Thanks for the fast implementation

PR #1059, opened 2024-06-17
clear error message when NAO with hardware id doesn't exist

### LINKED ISSUE #719: Panic on pepsi playernumber with invalid player numbers
Using `pepsi playernumber` with an invalid player number results in a panic.

Example:

```sh
./pepsi playernumber 42:1
```

There should be a descriptive error message instead.

MaikRe:
## Why? What?

Gives a more clear error message when a hardware ID that doesn't exist is used.

Fixes #719

## ToDo / Known Issues

## How to Test

Previously:
```
pepsi playernumber 88:1
[80:1] ⠙ Setting player number...                                                                                                                                                                                                                                                     The application panicked (crashed).
Message:  no entry found for key
Location: tools/pepsi/src/player_number.rs:48
```
Now:
```
pepsi playernumber 80:1 81:2
[80:1] ✗
   0: NAO with Hardware ID 80 does not exist

Location:
   tools/pepsi/src/player_number.rs:51
[81:2] ✗
   0: NAO with Hardware ID 81 does not exist

Location:
   tools/pepsi/src/player_number.rs:51
```

### DIFF
--- a/tools/pepsi/src/player_number.rs
+++ b/tools/pepsi/src/player_number.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 
 use clap::Args;
 use color_eyre::{
-    eyre::{bail, WrapErr},
+    eyre::{bail, eyre, WrapErr},
     Result,
 };
 
@@ -40,26 +40,19 @@ pub async fn player_number(arguments: Arguments, repository: &Repository) -> Res
     ) {
         bail!("Duplication in NAO to player number assignments")
     }
-
+    let hardware_ids = &hardware_ids;
     ProgressIndicator::map_tasks(
         arguments.assignments,
         "Setting player number...",
-        |assignment, _progress_bar| {
-            let hardware_id = &hardware_ids
-                .get(&assignment.nao_number.number)
-                .unwrap_or_else(|| {
-                    panic!(
-                        "NAO with Hardware ID {} does not exist",
-                        &assignment.nao_number.number
-                    )
-                });
-            let head_id = &hardware_id.head_id;
-            async move {
-                repository
-                    .set_player_number(head_id, assignment.player_number)
-                    .await
-                    .wrap_err_with(|| format!("failed to set player number for {assignment:?}"))
-            }
+        |assignment, _progress_bar| async move {
+            let number = assignment.nao_number.number;
+            let hardware_id = hardware_ids
+                .get(&number)
+                .ok_or_else(|| eyre!("NAO with Hardware ID {number} does not exist",))?;
+            repository
+                .set_player_number(&hardware_id.head_id, assignment.player_number)
+                .await
+                .wrap_err_with(|| format!("failed to set player number for {assignment}"))
         },
     )
     .await;
--- a/tools/pepsi/src/player_number.rs
+++ b/tools/pepsi/src/player_number.rs
@@ -45,7 +45,7 @@ pub async fn player_number(arguments: Arguments, repository: &Repository) -> Res
         arguments.assignments,
         "Setting player number...",
         |assignment, _progress_bar| {
-            let hardware = &hardware_ids
+            let hardware_id = &hardware_ids
                 .get(&assignment.nao_number.number)
                 .unwrap_or_else(|| {
                     panic!(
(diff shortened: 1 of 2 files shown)


tools/pepsi/src/player_number.rs
```diff
@@ -45,7 +45,15 @@ pub async fn player_number(arguments: Arguments, repository: &Repository) -> Res
         arguments.assignments,
         "Setting player number...",
         |assignment, _progress_bar| {
-            let head_id = &hardware_ids[&assignment.nao_number.number].head_id;
+            let hardware = &hardware_ids
```
schluis, line 48: Renaming suggestion: `hardware` -> `hardware_id`

schluis (changes requested):
Nice

tools/pepsi/src/player_number.rs
```diff
@@ -45,7 +45,15 @@ pub async fn player_number(arguments: Arguments, repository: &Repository) -> Res
...
+                .get(&assignment.nao_number.number)
+                .unwrap_or_else(|| {
+                    panic!(
+                        "NAO with Hardware ID {} does not exist",
+                        &assignment.nao_number.number
+                    )
```
schmidma, line 54: Please return a proper error instead of panic here

schluis: What's the advantage?

tools/pepsi/src/player_number.rs
MaikRe: Done

tools/pepsi/src/player_number.rs
MaikRe: Same question from me. But additionally if I did want to return a proper error here, how would I do this since the Result<_> does not implement IntoFuture and I therefore cannot return the Error out pf map_tasks back into the player_number function?

schmidma: This indeed needs some adjustment to the `map_tasks` function, as this falsely required the closure to be `Copy`. See below...

```
pub async fn map_tasks<T, F, M>(
    items: impl IntoIterator<Item = T>,
    message: &'static str,
    task: impl Fn(T, ProgressBar) -> F,
) where
    T: ToString,
    F: Future<Output = Result<M>>,
    M: Into<TaskMessage>,
{
    let multi_progress = Self::new();
    items
        .into_iter()
        .map(|item| (multi_progress.task(item.to_string()), item))
        .map(|(progress, item)| {
            progress.enable_steady_tick();
            progress.set_message(message);
            let future = task(item, progress.progress.clone());
            async move { progress.finish_with(future.await) }
        })
        .collect::<FuturesUnordered<_>>()
        .collect::<Vec<_>>()
        .await;
}
```

```
let hardware_ids = &hardware_ids;
ProgressIndicator::map_tasks(
    arguments.assignments,
    "Setting player number...",
    |assignment, _progress_bar| async move {
        let number = assignment.nao_number.number;
        let hardware_id = hardware_ids
            .get(&number)
            .ok_or_else(|| eyre!("NAO with Hardware ID {number} does not exist",))?;
        repository
            .set_player_number(&hardware_id.head_id, assignment.player_number)
            .await
            .wrap_err_with(|| format!("failed to set player number for {assignment}"))
    },
)
.await;
```

MaikRe: Thank you! This indeed much prettier. I have implemented the changes as recommended

MaikRe: Done

tools/pepsi/src/player_number.rs
```diff
@@ -40,18 +40,19 @@ pub async fn player_number(arguments: Arguments, repository: &Repository) -> Res
...
-                repository
-                    .set_player_number(head_id, assignment.player_number)
-                    .await
-                    .wrap_err_with(|| format!("failed to set player number for {assignment:?}"))
-            }
+        |assignment, _progress_bar| async move {
+            let number = assignment.nao_number.number;
+            let hardware_id = hardware_ids
+                .get(&number)
+                .ok_or_else(|| eyre!("NAO with Hardware ID {number} does not exist",))?;
```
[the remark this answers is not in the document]
MaikRe: Overlooked this, fixed

schluis (approved):

PR #1062, opened 2024-06-19
Remove unused bool fallen in HulkMessage

MaikRe:
## Why? What?

Removes unused ```fallen``` boolean that was a remnant of the time to reach implementation but is entirely unused

## How to Test

If CI is green then this open pr closed should be seen


Vivituhh (approved):
Schaut gut zu mir aus

PR #1064, opened 2024-06-19
increase referee pose detection parameter

Vivituhh:
## Why? What?
Increases `minimum_above_head_arms_detections` again to value before #831. 
Also increases `referee_pose_queue_length` and `minimum_number_poses_before_message` to be more robust.

Fixes #

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

*Describe how to test your changes. (For the reviewer)*

PR #1065, opened 2024-06-19
fix removal of voting for VR

Vivituhh:
## Why? What?
It fixes the issue that robots do not transition from `Initial` -> `Ready` with only 1 robot voting for it. 
Also it renames the booleans which represent the initial referee pose detection to hopefully prevent this mistake in the future. If there are renaming suggestions I'm open to rename again.

Fixes #

## ToDo / Known Issues

## How to Test

Test with only one robot the `Initial` -> `Ready` transition and compare results to main. It should not start walking with this fix as 2 independent detections are needed for the transition. (Assumes that #1064 is already merged, otherwise increase `minimum_above_head_arms_detections` >= 2.


schmidma (approved):
lgtm

PR #1066, opened 2024-06-19
Eye color for referee percept 

Vivituhh:
## Why? What?

As wished for this introduces eye color for each above arms pose percept being in the given radius around the referee at the T-junction (inside `distance_to_referee_position_threshold`). The chosen color is purple as it is better visible when flickering. 
Moreover the eye color is changed to yellow, when the robot is voting for referee pose detected, meaning its queue has enough above arms detected. 
As a result for each single percept the left eye flickers purple and when raising the arms long enough it changes to a continuous yellow color. 

Naming depends on #1065 

## ToDo / Known Issues

## How to Test
Deploy NAO as 4 or 7, start GameController, place NAO at its initial position, raise arms at T-junction and watch led colors (firstly purple flickering, after `minimum_number_poses_before_message` successive percepts change to continuous yellow). When only testing with one NAO it should not start walking, therefore it is possible to test a bit. Also test raising and take down again after one or two percepts and see only purple flickering.

### DIFF
--- a/crates/control/src/led_status.rs
+++ b/crates/control/src/led_status.rs
@@ -35,6 +35,7 @@ pub struct CycleContext {
     filtered_whistle: Input<FilteredWhistle, "filtered_whistle">,
     role: Input<Role, "role">,
     is_own_referee_ready_pose_detected: Input<bool, "is_referee_ready_pose_detected">,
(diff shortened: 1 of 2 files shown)


oleflb (approved):
Hübbisch

Issue #1047, opened 2024-06-03
Webots camera does not support parallel access


knoellle:
Even when disabling the openvino stuff, webots does not work properly.
The [camera implementation](https://github.com/HULKs/hulk/blob/main/crates/hulk_webots/src/camera.rs) for `hulk_webots` does not support the parallel access feature implemented in #564.

This results in the setup node crashing and returning an error.
But because of a bug in the communication server, the symptom is actually the simulation not advancing and communication using 100% of one CPU core, see #1046,

PR #1068, opened 2024-06-20
Add more descriptive OpenVino error message + installation information.

tuxbotix:
## Why? What?

- There is no mention of OpenVino requirements for Webots usage.
- OpenVino load errors are not explanatory enough in `ObjectDetection`, I spent quite some time to figure out why I got this "General Error" -> needed to install some additional components.

### Related issues
- #1047
- #1037

Fixes #

## How to Test

Launching for Webots will end up with the current status (#1047) *without* an OpenVino error.

```bash
./pepsi run
...
 Finished `incremental` profile [optimized] target(s) in 7m 21s
     Running `target/incremental/hulk_webots`
The Webots simulation world is not yet ready, pending until loading is done...
Warning: Webots [R2023b] and libController [] versions are not the same for Robot 'webots2'! Different versions can lead to undefined behavior.
2024-06-20 14:31:10  openvino_finder      INFO  Attempting to find library: libopenvino_c.so
2024-06-20 14:31:10  openvino_finder     DEBUG  Searching in: ......./libopenvino_c.so
...
...
2024-06-20 14:31:10  openvino_finder     DEBUG  Searching in: /lib/libopenvino_c.so
2024-06-20 14:31:10  openvino_finder     DEBUG  Searching in: /usr/lib64/libopenvino_c.so
2024-06-20 14:31:10  openvino_finder      INFO  Found library at path: /usr/lib64/libopenvino_c.so
cycler thread VisionTop returned error: 
   0: failed to execute cycle of cycler `VisionTop`
   1: failed to execute cycle of `ImageReceiver`
   2: termination requested

Location:
   crates/hulk_webots/src/hardware_interface.rs:297
```

### DIFF
--- a/crates/object_detection/src/pose_detection.rs
+++ b/crates/object_detection/src/pose_detection.rs
@@ -15,7 +15,9 @@ use hardware::{PathsInterface, TimeInterface};
 use itertools::Itertools;
 use linear_algebra::{point, vector};
 use ndarray::{s, ArrayView};
-use openvino::{CompiledModel, Core, DeviceType, ElementType, Tensor};
+use openvino::{
+    CompiledModel, Core, DeviceType, ElementType, InferenceError::GeneralError, Tensor,
+};
 use serde::{Deserialize, Serialize};
 use types::{
     bounding_box::BoundingBox,
@@ -89,9 +91,10 @@ impl PoseDetection {
                 weights_path
                     .to_str()
                     .wrap_err("failed to get detection weights path")?,
-            ).map_err(|error| match error{
-                openvino::InferenceError::GeneralError => eyre!("General Error: Please ensure that you have a complete & working OpenVino installation."),
-                _=> eyre!("failed to create detection network: {error}")
+            )
+            .map_err(|error| match error {
+                GeneralError => eyre!("{error}: possible incomplete OpenVino installation"),
+                _ => eyre!("{error}: failed to create detection network"),
             })?;
 
         let number_of_inputs = network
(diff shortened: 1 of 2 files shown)


crates/object_detection/src/pose_detection.rs
```diff
@@ -89,8 +89,10 @@ impl PoseDetection {
                 weights_path
                     .to_str()
                     .wrap_err("failed to get detection weights path")?,
-            )
-            .wrap_err("failed to create detection network")?;
+            ).map_err(|error| match error{
+                openvino::InferenceError::GeneralError => eyre!("General Error: Please ensure that you have a complete & working OpenVino installation."),
```
oleflb, line 93: Please replace qualified path with use. This error message is different than our usual error messages, don't you think the error message should just state the error, not propose solutions? Please also make the error message lowercase

oleflb (changes requested):
Thank you for improving error messages, just a small remark

crates/object_detection/src/pose_detection.rs
tuxbotix: Yes you are spot on about `use`, didn't notice it at first.
I wasn't aware of the lower case convention, will change.

As for the error message, they use GeneralError in a rather ambiguous way, could be reformulated to say "possible problem with missing OpenVino components" as this is one of the reasons.

As for errors with suggestions: when they are directly shown to the user or closer to the user, I feel it might make more sense to give a hint than just say we failed.

oleflb (approved):

Issue #1034, opened 2024-05-29
camera interface webots broken


pejotejo:
If you run webots (with locally installed openvino), you get the following error message, because of the multiple image cyclers:

```
Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
cycler thread ObjectDetectionTop returned error:
   0: failed to execute cycle of cycler `ObjectDetectionTop`
   1: failed to execute cycle of `ImageReceiver`
   2: failed to read from top camera
   3: no updated image found

Location:
   crates/hulk_webots/src/camera.rs:71
```

Probably you have to fix the webots equivalent of the hardware-interface.

schluis:
solved by #1069

PR #1069, opened 2024-06-20
Parallel/ multiple read() access for Webots Camera

### LINKED ISSUE #1047: Webots camera does not support parallel access

tuxbotix:
## Why? What?

As described in #1047 , the issue boils down to the way how the webots cameras are read from.

Compared to the `hulks_nao` camera, there isn't a queue of images, but a `Option<Vec<u8>>`. When `read()` is called, this option is emptied by calling `take()` resulting in the next `read()` call triggering an error (due to empty option) that happens *before*  `update_image()` is called.

Possible Solutions:
1. Clone the buffer and take -> no need to change the condition variable's announcing mechanism. Also the second cycler will get the same image (i.e. object detection)
2. Switch to `tokio` based approach done in #564 but I think this is overkill for the job.
3. Change how the condition variable is notified, use `notify_once` instead of `notify_all` -> two different cycles calling `read()` on same camera will get images from other times as things are now deferred. Additionally, the second cycler will wait for the next image.

This PR took the option 1, the simplest. IMO the possible drawbacks are lesser than the additional complexities we introduce from other methods.

Fixes #1047, #1034

## How to Test

Things should run on Webots fine, can test with Twix, etc.
Users with OpenVino errors should refer to #1068 and ensure their installation is done properly.

### DIFF
--- a/crates/hulk_webots/src/camera.rs
+++ b/crates/hulk_webots/src/camera.rs
@@ -6,7 +6,7 @@ use std::arch::x86_64::{
 };
 
 use color_eyre::{
-    eyre::{eyre, WrapErr},
+    eyre::{OptionExt, WrapErr},
     Result,
 };
 use parking_lot::{Condvar, Mutex};
@@ -67,8 +67,9 @@ impl Camera {
             let mut bgra_buffer = self.buffer.lock();
             self.buffer_updated.wait(&mut bgra_buffer);
             bgra_buffer
+                .clone()
                 .take()
-                .ok_or_else(|| eyre!("no updated image found"))?
+                .ok_or_eyre("no updated image found")?
         };
         assert_eq!(bgra_buffer.len(), 4 * 640 * 480);
         let mut ycbcr_buffer = vec![


crates/hulk_webots/src/camera.rs
```diff
@@ -67,8 +67,9 @@ impl Camera {
             let mut bgra_buffer = self.buffer.lock();
             self.buffer_updated.wait(&mut bgra_buffer);
             bgra_buffer
+                .clone()
```
oleflb, line 70: This behavior is different compared to the robotics implementation. I'd be fine merging this anyway and opening an issue that we should make the implementations similar. However the fix in this PR would allow people to use webots again. What do you think of this @schmidma @knoellle

oleflb (commented):
Waiting for feedback

crates/hulk_webots/src/camera.rs
tuxbotix: I was tempted to use the tokio sender/ receiver pattern, but I wanted to get it working in the first place to work on another PR.

schmidma: Please open an issue with a proper description of the problem, then I'm totally fine with this quick fix to get webots running again

oleflb:
Someone with webots please test, opened issue in #1073 to describe the issue with this PR. If webots works with this PR, we can merge it

crates/hulk_webots/src/camera.rs
oleflb: #1073

tuxbotix:
I'm already using webots with this :wink: , of course, if secondary verification is needed, we need to wait for a volunteer ^_^ 
image

PR #1075, opened 2024-06-23
ball detected far from kick off depending on seen location

### LINKED ISSUE #861: Post-game task: Investigate center circle incursions
No. 6:41 and 7:38 illegally entered the center circle towards the end of the first half of our first game today (and then both didn't kick the ball). Investigate.

### LINKED ISSUE #1043: New method for ball is moved
We need a new method for detecting whether the ball has moved for situations like the kickoff #1035 #861 which is not based on absolute change in position but rather maybe a movement over time, requiring some velocity. If the camera matrix of the striker is improperly calibrated he might think the ball has moved and illegally walk into the center circle.

```rust
    127 let ball_detected_far_from_kick_off_point = ball_position
    128   .map(|ball| {
    129       let absolute_ball_position = ground_to_field * ball.position;
    130        distance(absolute_ball_position, Point2::origin())
    131           > config.distance_to_consider_ball_moved_in_kick_off
    132  })
    133    .unwrap_or(false);
```
in ```game_controller_state_filter.rs```

MaikRe:
## Why? What?

We have often had an issue where camera matrix miscalibrations lead to the robot seeing the ball at a location that was not the midpoint of the center circle, thus ```ball_detected_far_from_kick_off_point``` see #861.
This should fix that.

The cycle we hear the whistle we now save the location of the ball. We check for whether the ball has moved according to this seen location, if thats not an option we use our old method.

Additionally we don't use active vision during kickoff when the ball is not free. This means we look at the ball until either the gamecontroller says it is free or we see it move and it is free. This if the camera matrix is off we don't get random ball drifts due to head movements.

Fixes #861
Fixes #1043

Implementing something similar for the penalty shot direction provider

## How to Test

Take a robot and purposely set its camera matrix wrong until it always violates the center circle during kickoff using the old software. Now use the new software and see the difference.

### DIFF
--- a/crates/control/src/active_vision.rs
+++ b/crates/control/src/active_vision.rs
@@ -11,6 +11,8 @@ use linear_algebra::{point, Isometry2, Point2, Vector2};
 use types::{
     cycle_time::CycleTime,
     field_dimensions::FieldDimensions,
+    filtered_game_controller_state::FilteredGameControllerState,
+    filtered_game_state::FilteredGameState,
     obstacles::{Obstacle, ObstacleKind},
     parameters::LookActionParameters,
     point_of_interest::PointOfInterest,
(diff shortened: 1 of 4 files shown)


oleflb (approved):
Zänk ju vor sis greyt imprufmänt

PR #1077, opened 2024-06-24
Fix typo in cargo::rerun-if-changed

julianschuler:
## Why? What?

This PR changes `cargo:rerun-if-changed` to the new `cargo::rerun-if-changed` notation.

See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script.

## ToDo / Known Issues

This requires at least rust 1.77, so this is blocked until we have updated everything to at least 1.77.

None.

## How to Test

Build it.


oleflb (approved):
Approving but we need a newer rust version in the SDK in order for the CI to green.

Issue #699, opened 2024-01-24
Better Field Color Detection


schluis:
We should improve our Field Color Detection so that is more robust against bad lighting conditions and shadows.

PR #1078, opened 2024-06-24
Use hsv color space

massimiliano-mantione:
## Why? What?

The current segment detection is not accurate enough.

Using the HSV color space it is easier to define thresholds for "similarly looking " colors, so this PR adds support for that.

Related to #699 

## ToDo / Known Issues

The configuration should be reworked.

None

## How to Test

Select the `Hsv` function in the vision tuner panel.
Look in the image segments panel if the field recognition improves.
In general the hue for the field green is between 70 and 125, and the saturation is way above 100.
It also makes sense to force the luminance ("value" in VSV) above 50 or so, to be sure that the saturation is meaningful.

### DIFF
--- a/crates/types/src/color.rs
+++ b/crates/types/src/color.rs
@@ -197,6 +197,59 @@ impl From<YCbCr444> for Rgb {
     }
 }
 
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Deserialize,
+    Eq,
+    PartialEq,
+    Serialize,
+    PathSerialize,
+    PathDeserialize,
+    PathIntrospect,
+)]
+pub struct Hsv {
+    pub h: u16,
+    pub s: u8,
+    pub v: u8,
+}
+
+impl From<Rgb> for Hsv {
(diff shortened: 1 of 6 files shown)


schluis (approved):
Hey Massimiliano, thanks for your contribution! I reviewed the code, everything seems very nice! 🚀
- Apparently, you missed to update a few test-cases in `crates/vision/src/image_segmenter.rs`, where the new struct fields in `FieldColor` are missing
- Feel free to turn on auto-merge (if you're allowed to do so, we're currently unsure 😅)

schluis:
A few things depend on this, so we decided to just merge this and fix the test cases afterwards -> #1105 

Thanks again! :)

massimiliano-mantione:
Thanks for merging!
Last week I was traveling (vacation) and I could not follow it...
But thanks, this has been fun!

h3ndrk:
Hey @massimiliano-mantione, just wanted to give you an update live from RoboCup 2024. We've deployed the HSV color space by default now and it works amazing. Thank you for your contribution!! :100:

PR #1080, opened 2024-06-24
Rule Balls in Heatmap

### LINKED ISSUE #514: Use `SubState` in ball search
If information about possible ball positions can be extracted from the current game state, we should use this information for ball search.
Some ideas/examples of where to search when:
- corners during `SubState::CornerKick`
- goalkick positions during `SubState::GoalKick`
- field boundaries close to the last known ball position during `SubState::KickIn`
- penalty spot(s) during `SubState::PenaltyKick`

### LINKED ISSUE #994: Ruleballs in Heatmap
Add search hypotheses for corner kick, ...

knoellle:
## Why? What?

Adds "heat" to the ball search heat map in positions where a ball might be according to the rules.
Currently covers kick off, penalty kicks, goal kicks and corner kicks.
Not handled are kick-in and pushing free kick.
Kick-in because it can happen on any point on the side lines.
Pushing free kick does **not** need special handling because the ball doesn't (or isn't supposed to) move.

N.b. I haven't tested this with a game controller yet, so I'm not sure if the conditions are quite correct.

Fixes #994 
Fixes #514

Handle kick-in somehow, maybe by projecting the existing heat to the nearest side line.

## How to Test

Check heatmap in the relevant situations.

### DIFF
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -7,10 +7,13 @@ use framework::{AdditionalOutput, MainOutput};
 use linear_algebra::{point, Isometry2, Point2};
 use nalgebra::{clamp, DMatrix};
 use serde::{Deserialize, Serialize};
+use spl_network_messages::{SubState, Team};
 use types::{
     ball_position::{BallPosition, HypotheticalBallPosition},
-    field_dimensions::FieldDimensions,
+    field_dimensions::{FieldDimensions, Half, Side},
+    filtered_game_controller_state::FilteredGameControllerState,
     parameters::SearchSuggestorParameters,
+    primary_state::PrimaryState,
 };
 
 #[derive(Deserialize, Serialize)]
@@ -27,10 +30,16 @@ pub struct CreationContext {
 #[context]
 pub struct CycleContext {
     search_suggestor_configuration: Parameter<SearchSuggestorParameters, "search_suggestor">,
+    field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
+
     ball_position: Input<Option<BallPosition<Ground>>, "ball_position?">,
     hypothetical_ball_positions:
         Input<Vec<HypotheticalBallPosition<Ground>>, "hypothetical_ball_positions">,
     ground_to_field: Input<Option<Isometry2<Ground, Field>>, "ground_to_field?">,
+    primary_state: Input<PrimaryState, "primary_state">,
+    filtered_game_controller_state:
+        Input<Option<FilteredGameControllerState>, "filtered_game_controller_state?">,
+
     heatmap: AdditionalOutput<DMatrix<f32>, "ball_search_heatmap">,
 }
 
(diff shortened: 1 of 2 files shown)


crates/control/src/search_suggestor.rs
```diff
@@ -151,3 +161,45 @@ impl IndexMut<Point2<Field>> for Heatmap {
...
+            vec![
+                field_dimensions.corner(kick_half, Side::Left),
+                field_dimensions.corner(kick_half, Side::Right),
+            ]
+        }
+        (PrimaryState::Playing, Some(SubState::GoalKick)) => {
+            let kick_half = kicking_team_half;
+            vec![
+                field_dimensions.corner(kick_half, Side::Left),
+                field_dimensions.corner(kick_half, Side::Right),
```
MaikRe, line 191:
```suggestion
        (PrimaryState::Playing, Some(SubState::GoalKick)) => {
            let kick_half = kicking_team_half;
            vec![
                field_dimensions.goal_box_corner(kick_half, Side::Left),
                field_dimensions.goal_box_corner(kick_half, Side::Right),
```

crates/control/src/search_suggestor.rs
```diff
@@ -151,3 +161,45 @@ impl IndexMut<Point2<Field>> for Heatmap {
...
+fn get_rule_hypotheses(
+    primary_state: PrimaryState,
+    filtered_game_controller_state: FilteredGameControllerState,
+    field_dimensions: FieldDimensions,
+) -> Vec<Point2<Field>> {
+    let kicking_team_half =
+        conservative_kicking_team_half(filtered_game_controller_state.kicking_team);
+
+    match (primary_state, filtered_game_controller_state.sub_state) {
+        (PrimaryState::Ready, Some(SubState::PenaltyKick)) => {
```
MaikRe, line 174: Should this also include the primary state Set?

crates/control/src/search_suggestor.rs
```diff
@@ -151,3 +161,45 @@ impl IndexMut<Point2<Field>> for Heatmap {
...
+            let kick_half = kicking_team_half.mirror();
+            vec![field_dimensions.penalty_spot(kick_half)]
+        }
+        // Kick-off
+        (PrimaryState::Ready, None) => vec![field_dimensions.center()],
```
MaikRe, line 179: This both seems like it overlaps/duplicates a bit with the ```rule_ball``` in the ```ball_state_composer```. 
I would suggest maybe using the rule ball as an input for the search suggestor, what do you think?

MaikRe (changes requested):

crates/control/src/search_suggestor.rs
MaikRe: Fair enough

crates/control/src/search_suggestor.rs
knoellle: I'm not sure about the exact ball search mechanics, but during set no one should be doing ball search anyways.

@JonasHolz do you think it makes sense to apply rule ball positions during set also?

crates/control/src/search_suggestor.rs
knoellle: Done

MaikRe:
Does this also fix #514?

crates/control/src/search_suggestor.rs
MaikRe: macht zu mir sinn

MaikRe (approved):
Arbeitet wie in der Werbung

PR #1084, opened 2024-06-29
Use new API calls for OpenVino

### LINKED ISSUE #1083: openvino-rs 0.7.1 is yanked, new version has breaking changes
I found this weird issue after `cargo clean`, turns out 0.7.1 which we use is yanked (taken out altogether?).

The next version has a different API call to create Tensors, therefore `object_detection` is broken as of now (at least for non-nao builds).

- Release notes: https://github.com/intel/openvino-rs/releases/tag/v0.7.2
- The API change: https://github.com/intel/openvino-rs/commit/b6eacef9b5ab2cc964abbf02be3c74b0d7a948b1

```
error[E0599]: no function or associated item named `new_from_host_ptr` found for struct `Tensor` in the current scope
   --> crates/object_detection/src/pose_detection.rs:155:30
    |
155 |         let tensor = Tensor::new_from_host_ptr(
    |                              ^^^^^^^^^^^^^^^^^ function or associated item not found in `Tensor`
    |
note: if you're trying to build a new `Tensor` consider using one of the following associated functions:
      Tensor::new
      Tensor::from_ptr
   --> /home/darshana/.cargo/registry/src/index.crates.io-6f17d22bba15001f/openvino-0.7.2/src/tensor.rs:21:5
    |
21  |     pub fn new(element_type: ElementType, shape: &Shape) -> Result<Self> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
33  |     pub(crate) fn from_ptr(ptr: *mut ov_tensor_t) -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tuxbotix:
## Why? What?

Use the changed API for creating a tensor. How-To taken from their commit: https://github.com/intel/openvino-rs/commit/b6eacef9b5ab2cc964abbf02be3c74b0d7a948b1

Fixes #1083

- Add a 'smoke test' for the Object Detector to catch OpenVino setup, etc issues (Hopefully, the CI will also catch these?)

## How to Test

`./pepsi run` should build for webots and launch properly.

### DIFF
--- a/crates/object_detection/src/pose_detection.rs
+++ b/crates/object_detection/src/pose_detection.rs
@@ -254,17 +254,3 @@ fn non_maximum_suppression(
 
     poses
 }
-
-// Allowing dead code for future use of `scratchpad` once it is restored.
-#[allow(dead_code)]
-trait AsBytes {
-    fn as_bytes(&self) -> &[u8];
-}
-
-impl AsBytes for [f32] {
-    fn as_bytes(&self) -> &[u8] {
-        unsafe {
-            std::slice::from_raw_parts(self.as_ptr() as *const u8, std::mem::size_of_val(self))
-        }
-    }
-}
--- a/crates/object_detection/src/pose_detection.rs
+++ b/crates/object_detection/src/pose_detection.rs
@@ -152,11 +152,11 @@ impl PoseDetection {
         }
 
         let mut infer_request = self.network.create_infer_request()?;
-        let tensor = Tensor::new_from_host_ptr(
-            ElementType::F32,
-            &self.network.get_input()?.get_shape()?,
-            self.scratchpad.as_bytes(),
-        )?;
+        let mut tensor = Tensor::new(ElementType::F32, &self.network.get_input()?.get_shape()?)?;
+        tensor
+            .buffer_mut()?
+            .copy_from_slice(self.scratchpad.as_bytes());
+
         infer_request.set_input_tensor(&tensor)?;
 
         {
--- a/crates/object_detection/src/pose_detection.rs
+++ b/crates/object_detection/src/pose_detection.rs
@@ -38,8 +38,6 @@ const STRIDE: usize = DETECTION_IMAGE_HEIGHT * DETECTION_IMAGE_WIDTH;
 #[derive(Deserialize, Serialize)]
 pub struct PoseDetection {
     #[serde(skip, default = "deserialize_not_implemented")]
-    scratchpad: Box<[f32]>,
-    #[serde(skip, default = "deserialize_not_implemented")]
     network: CompiledModel,
 }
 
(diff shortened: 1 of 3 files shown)


oleflb:
unfortunately this requires cloning..., opened some issues over at openvino
- [133](https://github.com/intel/openvino-rs/issues/133)
- [134](https://github.com/intel/openvino-rs/issues/134)

tuxbotix:
I think the reason to yank is the `0.7.1`, `0.7.0` releases have memory leaks and segfaults caused by the API that got replaced in addition to some others. 2-3 of these fixes are related to leaks and other memory issues -> https://github.com/intel/openvino-rs/releases/tag/v0.7.2 . While I'm not a fan of yanking a release, I can't blame them for doing this as the listed issues are quite serious.

The cost of copying is also discussed in the change of API (second link in my issue). IMHO a slightly less performant library is better than a library that might segfault.

Additionally, we can consider changing the cycler's architecture to keep the tensor as part of the state and load the input image directly into it instead of the scratchpad.

oleflb:
No, not possible since the Tensor does not impl Send :(

crates/object_detection/src/pose_detection.rs
```diff
@@ -152,11 +152,12 @@ impl PoseDetection {
         }
 
         let mut infer_request = self.network.create_infer_request()?;
-        let tensor = Tensor::new_from_host_ptr(
-            ElementType::F32,
-            &self.network.get_input()?.get_shape()?,
-            self.scratchpad.as_bytes(),
-        )?;
+        let mut tensor = Tensor::new(ElementType::F32, &self.network.get_input()?.get_shape()?)?;
```
oleflb, line 155: remove `scratchpad` from node state and pass the tensor buffer to the `load_into_scratchpad` directly. This way we save ourselves a whole copy of the image

oleflb (changes requested):

tuxbotix:
Ouch! 😬 That's bad 😞.
I'll try to setup a benchmark today if I can to see the impact of the copy and some alternative we could try.

Based on the notes by the authors, they would need to touch a lot of code to get the lifetime information right to create a tensor from a pointer and ensure safety.

oleflb:
For now we can just remove the `scratchpad` and create the `Tensor` in the cycle function. you can use `get_data_mut` to create a mutable f32 slice that you can pass to the load_into_scratchpad function.

crates/object_detection/src/pose_detection.rs
tuxbotix: Done, let me know if I should also remove the `AsBytes` trait as well.
I didn't previously notice why you used the scratchpad, it all makes sense :) .

In summary, we saved the copy but still have the memory allocation in each cycle.

oleflb:
Yes, feel free to remove the trait. I'd say we merge this state with the allocation.

crates/object_detection/src/pose_detection.rs
```diff
@@ -268,6 +255,8 @@ fn non_maximum_suppression(
     poses
 }
 
+// Allowing dead code for future use of `scratchpad` once it is restored.
+#[allow(dead_code)]
```
oleflb, line 259: please remove instead. We can add this back (if needed)

oleflb (changes requested):

crates/object_detection/src/pose_detection.rs
tuxbotix: Done :)

oleflb (approved):
thank you for fixing :)

tuxbotix:
And thank you for integrating all of this and catching the things I missed!!

PR #1087, opened 2024-07-01
Look at all previous players to decide if replacement keeper is best role

### LINKED ISSUE #1082: Fix Two Replacement Keeper again
Current code only checks the previous player for penalties when deciding when not to be ```ReplacementKeeper```
If ```PlayerNumber::Two``` is permanently picked up or otherwise out of the game the ```Three``` will remain ```ReplacementKeeper``` even if the ```One``` has returned to be ```Keeper```.

```rust
if self.role == Role::ReplacementKeeper {
            let previous_player = match context.player_number {
                PlayerNumber::Three => PlayerNumber::Two,
                PlayerNumber::Four => PlayerNumber::Three,
                PlayerNumber::Five => PlayerNumber::Four,
                PlayerNumber::Six => PlayerNumber::Five,
                PlayerNumber::Seven => PlayerNumber::Six,
                _ => PlayerNumber::One,
            };
            if let Some(last_time_player_penalized) =
                self.last_time_player_was_penalized[previous_player]
            {
                let deny_replacement_keeper_switch = cycle_start_time
                    .duration_since(last_time_player_penalized)
                    .expect("Keeper/Replacmentkeeper was penalized in the future")
                    < *context.keeper_replacementkeeper_switch_time;
                if !send_spl_striker_message && deny_replacement_keeper_switch {
                    new_role = Role::ReplacementKeeper;
                }
            }
        }
```

MaikRe:
## Why? What?

There was a bug that theoretically could have allowed several replacement keepers as only the previous player was considered when deciding if to stay replacement keeper, i,.e. player three didn't care if player one was unpenalized and back in the game.

This fixes that by checking all players up until the own player number to see if they are back so that they become replacement keeper and the player itself can choose a new role.

The lua scenario file has been edited to reflect the scenario in which this bug was noticed

Fixes #1082 

## How to Test

Penalize player 2, penalize the goalkeeper, unpenalize the goalkeeper. Before this would lead to player 3 remaining replacement keeper, this is no longer the case.


schmidma:
I had a brief look at it...
First, the naming has to be improved drastically. It is extremely hard to follow what is going on. Here is some suggestion:

```rust
let mut other_players_with_lower_number = self
    .last_time_player_was_penalized
    .iter()
    .filter(|(player_number, _)| player_number < context.player_number);
let is_lowest_number_without_penalty =
    other_players_with_lower_number.all(|(_, penalized_time)| {
        penalized_time
            .map(|system_time| {
                let since_last_penalized = cycle_start_time
                    .duration_since(system_time)
                    .expect("penalty time to be in the past");
                since_last_penalized < *context.keeper_replacementkeeper_switch_time
            })
            .unwrap_or(false)
    });
```
Something like this. Additionally I did not understand what this `keeper_replacementkeeper_switch_time` is about? Maybe we can discuss this in the lab today.

MaikRe:
The ```keeper_replacementkeeper_switch_time``` is the time we give the keeper to come back into the game before the replacement keeper leaves the goal. Originally named by @JonasHolz and I would be happy to rename it too but I couldn't think of anything better. Happy to discuss

schmidma (approved):

Issue #997, opened 2024-05-04
Penalty Shootout Keeper


oleflb:
Our Keeper jumps directly after going down, likely because the ball gets velocity due to the sensor lag.
Ideas:
~- Just go down earlier (anyways good idea)~
~- ball only gets velocity in Play~
~- use whistle detection?~
- [ ] Ask the Devils if they would provide their Keeper motion since it looks soooo much better than ours
- [x] Implement the wide stance from the wide stance into the current keeper behavior

MaikRe:
1. I don't know/think we can go down much earlier. The goal keeper is placed while the game state is set, and this only goes into playing through the whistle. Assuming the robot hears the whistle, it immediately does the action ```PrepareJump```.
3. We use the whistle detection, thats how we know when we are allow to prepare the jump
4. Technically speaking the Devil's motion could eventually be illegal as it includes a T-Pose. If the attacking robot waits 10 seconds to do the kick the goalkeeper either needs to move its arms back in, or it is called for wide stance and a goal is automatically counted.

MaikRe:
As for the ball velocity, we only do penalty direction estimation in playing, it may be good to wait until the sit down motion is finished for this to happen?
```rust
(PrimaryState::Playing, GamePhase::PenaltyShootout { .. }, ..)
            | (PrimaryState::Playing, _, Some(SubState::PenaltyKick), Team::Opponent) => {
                if let PenaltyShotDirection::NotMoving = self.last_shot_direction {
                    if (context.ball_position.position.x()
                        - context.field_dimensions.penalty_marker_distance)
                        .abs()
                        > *context.moving_distance_threshold
```
but I have no idea how that would be best implemented.
The other issue sort of relates to #1043, in that we do the estimation of movement purely based on the location of the penalty marker. i.e. if the camera matrix is off (which can easily happen if the robot has already jumped once) then this estimation will likely immediately trigger a penalty shot estimation and the robot jumps.

Therefore we once again need an estimation for this situation which does a relative movement of the ball or we need some other method which properly gives the ball velocity

MaikRe:
I am modifying the description to match the current state of what could be done, or should be implemented

MaikRe:
[...]

Issue #546, opened 2023-09-06
Penalty Shootout


okiwi6:
Our shootout performance at the robocup 2023 competition  ... leaves room for improvement.
The focus here is the defending capabilities of the robot, especially jumping in the right direction at the right time.

Estimated time: 4 weeks

MaikRe:
```node.rs```
```rust
  Role::Striker => match filtered_game_state {
                None
                | Some(FilteredGameState::Playing {
                    ball_is_free: true, ..
                }) => {
                    actions.push(Action::Dribble);
                }
```
It may be worth considering having a different/specific action other than dribbling for the striker, so that we don't just walk through the ball/it doesn't modify/interfere with out current dribbling behavior.

MaikRe:
After more clear reading of the the rules it seems that the widestance penalty also applies to the goal keeper during the penalty shootout. Therefore it is of interest of us:
- if the striker hears the whistle, wait for 10-15 seconds in hopes the opponent goalkeeper is penalized for wide stance
- If we do not hear the whistle then we should strike as normal

Also the new kick targets have made kicking better, but we end up with double contacts because the reached thresholds are wrong

oleflb:
fixed in #1168

PR #1088, opened 2024-07-01
Penalty shootout keeper improvements

### LINKED ISSUE #477: Penalty shootout behaviour is buggy/ weird

MaikRe:
## Why? What?

- Correctly set location during set phase of penalty shootout to prevent robot randomely jerking head because localization keeps generating alternating field sides
- Turn off active vision during penalty shootout to prevent keeper looking up and down between ball and 1m in front of self
- Use whistle in set ball position to determine penalty shot direction to counter for poor camera matrix
- Use a y distance threshold other than zero, specifically require a distance to be greater than one where the linear projection of the ball would lead to it rolling past the robots feet. If the threshold is not met it probably makes more sense to remain standing

Mitigates #997 and #546
Fixes #477 

The y direction threshold can be further adapted to accommodate for a wide stance once implemented

## How to Test

Do penalty shootout I guess. A slightly off camera matrix should no longer lead to immediate jumping. And if the ball rolls straight at the robot he should remain standing.

### DIFF
--- a/crates/control/src/penalty_shot_direction_estimation.rs
+++ b/crates/control/src/penalty_shot_direction_estimation.rs
@@ -1,9 +1,9 @@
 use ball_filter::BallPosition;
 use color_eyre::Result;
 use context_attribute::context;
-use coordinate_systems::Ground;
+use coordinate_systems::{Field, Ground};
 use framework::MainOutput;
-use linear_algebra::{point, Point2};
+use linear_algebra::{Isometry2, Point2};
 use serde::{Deserialize, Serialize};
 use spl_network_messages::{GamePhase, SubState, Team};
 use types::{
@@ -32,6 +32,7 @@ pub struct CycleContext {
     filtered_game_controller_state:
         RequiredInput<Option<FilteredGameControllerState>, "filtered_game_controller_state?">,
     primary_state: Input<PrimaryState, "primary_state">,
+    ground_to_field: RequiredInput<Option<Isometry2<Ground, Field>>, "ground_to_field?">,
 }
 
 #[context]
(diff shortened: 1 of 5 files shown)


MaikRe:
@knoellle  I could not reproduce the weird drift issue we had yesterday, even when using the same robot and following the same game states that we had use, including undos. 
Would need to be looked at more, for now I suggest merging this PR

crates/control/src/penalty_shot_direction_estimation.rs
```diff
@@ -53,28 +59,46 @@ impl PenaltyShotDirectionEstimation {
...
                 self.last_shot_direction = PenaltyShotDirection::NotMoving;
+                self.placed_ball_position = Some(context.ball_position.position);
                 Ok(MainOutputs::default())
             }
             (PrimaryState::Playing, GamePhase::PenaltyShootout { .. }, ..)
             | (PrimaryState::Playing, _, Some(SubState::PenaltyKick), Team::Opponent) => {
+                let penalty_marker_position_in_ground = context.ground_to_field.inverse()
+                    * FieldDimensions::penalty_spot(
+                        context.field_dimensions,
+                        types::field_dimensions::Half::Own,
```
knoellle, line 70: replace qualified path with use

knoellle (approved):
Held good to me

PR #1089, opened 2024-07-01
Restore Sanity in Vision Parameters

schmidma:
## Why? What?

ich kann nicht mehr...

## How to Test

open twix, use the vision tuner panel to tune...


schluis:
I strongly disagree with merging this. This removes a feature, which was needed in the past and which might be needed again. There currently doesn't exist an alternative.

schmidma:
Superseded by #1203

Issue #1095, opened 2024-07-03
Expose features flag to pepsi


ThagonDuarte:
Pepsi does not yet expose the `--features` flag of `cargo`. This flag is needed to run the replayer with the `ObjectDetectionTop` cycler running.

Issue #778, opened 2024-03-27
Improve recording triggers


h3ndrk:
https://github.com/HULKs/hulk/pull/517#issuecomment-1685232445 and https://github.com/HULKs/hulk/pull/517#issuecomment-1685235621

PR #1097, opened 2024-07-03
Enable visual referee recording

ThagonDuarte:
## Why? What?

Enables visual referee by default and sets the `Standby` state to be recorded.

## How to Test

Record a visual referee scenario and run the replayer with the `--features "with_object_detection"`. Requires openvino to be installed locally. 
This may change as is tracked in #1095.

### DIFF
--- a/etc/parameters/framework.json
+++ b/etc/parameters/framework.json
@@ -3,7 +3,6 @@
   "hardware_parameters": "etc/parameters/hardware.json",
   "parameters_directory": "etc/parameters",
   "recording_intervals": {
-    "Control": 1,
-    "ObjectDetectionTop": 1
+    "Control": 1
   }
 }
--- a/etc/parameters/framework.json
+++ b/etc/parameters/framework.json
@@ -3,6 +3,7 @@
   "hardware_parameters": "etc/parameters/hardware.json",
   "parameters_directory": "etc/parameters",
   "recording_intervals": {
-    "Control": 1
+    "Control": 1,
+    "ObjectDetectionTop": 1
   }
 }
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1,7 +1,7 @@
 {
   "object_detection": {
     "object_detection_top": {
-      "enable": false,
+      "enable": true,
       "intersection_over_union_threshold": 0.45,
       "distance_to_referee_position_threshold": 2.0,
       "keypoint_confidence_threshold": 0.2,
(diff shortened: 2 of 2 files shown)


etc/parameters/framework.json
```diff
@@ -3,6 +3,7 @@
   "hardware_parameters": "etc/parameters/hardware.json",
   "parameters_directory": "etc/parameters",
   "recording_intervals": {
-    "Control": 1
+    "Control": 1,
+    "ObjectDetectionTop": 1
```
schmidma, line 7: Wait a second... This enables recording of `ObjectDetectionTop` for every cycle. And every cycle for the entire game means, we record every image of the vision top cam?

@h3ndrk @knoellle

h3ndrk: If the object detection cycler runs with lower frequency, it will skip images, so not every image. I don't know, how much slower it actually runs.

Maybe this should instead be added to pepsi pregame instead of the default `framework.json`?

schmidma: But it only runs slower in Initial. In all the other gamestates, it completes instantly, as it is skipping computation in the nodes. This leads to the object detection cycler recording every image. Doesn't it?

ThagonDuarte: I didn't realize this until now tbh, but you are correct. Is there a way to completely disable a cycler at runtime?

schmidma: no

schmidma: unless you block its execution in the setup node... but that is also not a good idea

h3ndrk: My long term idea was to also add a way from reporting from robotics domain whether and what and how often you want to record. This way, we could disable general recording in Initial but the object detection cycler could override that if it actually runs. See https://github.com/HULKs/hulk/issues/778

h3ndrk: I think the general problem or too simple current state is that we are not able to express the full decision matrix currently for all orthogonal inputs/reasons/causes. These inputs are (hopefully I didn't forget any):

- Which cycler
- Which frequency or maybe a burst for a short time
- Which game state
- Whether static framework.json, Communication, or robotics domain requested it

ThagonDuarte: Then I would suggest for now only enabling the recording during standby but not recording the `ObjectDetectionTop` cycler by default or by use of pregame. If we really need to record it, we can still use the `pepsi pregame --recording-intervals`.

schmidma (approved):

PR #1101, opened 2024-07-04
kickoff position with rotation

MaikRe:
## Why? What?

Make kickoff in the direction of the supporting striker easier by positioning differently in ready

Fixes #

Kick off to left supporter if supporting striker is penalized? I.e. rotate 180 degrees about midpoint Z axis

## How to Test

Behavior sim or test game. Observe Striker position when HULKs are kicking team

### DIFF
--- a/crates/types/src/parameters.rs
+++ b/crates/types/src/parameters.rs
@@ -2,7 +2,7 @@ use std::ops::{Index, Range};
 use std::{path::PathBuf, time::Duration};
 
 use coordinate_systems::{Field, Ground, NormalizedPixel};
-use linear_algebra::{Point2, Vector2};
+use linear_algebra::{Point2, Pose2, Vector2};
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
 use serde::{Deserialize, Serialize};
 
@@ -87,8 +87,7 @@ pub struct RolePositionsParameters {
     pub striker_supporter_minimum_x: f32,
     pub keeper_x_offset: f32,
     pub striker_distance_to_non_free_center_circle: f32,
-    pub striker_kickoff_position: Point2<Field>,
-    pub striker_kickoff_orientation: f32,
+    pub striker_kickoff_pose: Pose2<Field>,
 }
 
 #[derive(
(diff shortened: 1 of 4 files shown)


crates/types/src/parameters.rs
```diff
@@ -87,7 +87,8 @@ pub struct RolePositionsParameters {
     pub striker_supporter_minimum_x: f32,
     pub keeper_x_offset: f32,
     pub striker_distance_to_non_free_center_circle: f32,
-    pub striker_set_position: Point2<Field>,
+    pub striker_kickoff_position: Point2<Field>,
+    pub striker_kickoff_orientation: f32,
```
schmidma, line 91: I'd suggest to just make it a `Pose2<Field>`. This is exactly the type to encode position and rotation

schmidma (changes requested):

crates/types/src/parameters.rs
MaikRe: Because of the way the `Orientation2` is deserialized the ```Pose2``` in the ```default.json``` would have look like this in order to ensure correct rotation
```json
"striker_kickoff_pose": {
        "translation": [0.0, 0.3],
        "rotation": [0.0007962742820382118, -0.9999997019767761]
      }
```
Thoughts?

schmidma: I think this is okay for now, as this is another issue. There may even be an issue already opened for deserializing degree values as unit complexs...

etc/parameters/default.json
```diff
@@ -1111,7 +1111,10 @@
       "striker_supporter_minimum_x": 2.0,
       "keeper_x_offset": 0.1,
       "striker_distance_to_non_free_center_circle": 0.4,
-      "striker_set_position": [-0.3, 0.0]
+      "striker_kickoff_pose": {
+        "translation": [0.0, 0.3],
+        "rotation": [0.0007962742820382118, -0.9999997019767761]
```
schmidma, line 1116: Do you intend to have it turned by 90 degrees? Then better put a clear `0.0` and `-1.0`

MaikRe: Good point, I serialized it to see what the proper format was and then didn't think about rounding errors. Fixed now

schmidma (approved):

PR #1102, opened 2024-07-04
Expose features flag to pepsi

### LINKED ISSUE #1095: Expose features flag to pepsi

ThagonDuarte:
## Why? What?

Exposes the `--features <features>` flag of cargo to pepsi. This flag is needed to enable the `ObjectDetectionCycler` when running the replayer.   

Fixes #1095

## How to Test

Run the command:
`$ pepsi run --target replayer --features with_object_detection -- <path to replay log>`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4511,7 +4511,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pepsi"
-version = "3.12.0"
+version = "3.13.0"
 dependencies = [
  "aliveness",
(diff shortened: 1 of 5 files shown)


oleflb:
Does not work when no features are supplied ;)

oleflb (changes requested):
What do you think of making the argument `Option<&str>` instead? I think that's more verbose compared to checking the string length

knoellle:
"features" sounds suspiciously plural though. How about a vec or a set? 😉

oleflb (approved):
Noice #2

PR #1108, opened 2024-07-05
Avoid singularity in testcase

knoellle:
## Why? What?

[Why?](https://github.com/HULKs/hulk/actions/runs/9812469696/job/27096599499?pr=1107#step:4:211)
What? There's a singularity at PI and -PI that apparently causes the test to fail when we unintentionally bump the nalgebra version.

Fixes #

## How to Test

'ave a propaganda at the CI


MaikRe (approved):
Oy govna! Migh'y fine job ye done theya!

PR #1110, opened 2024-07-06
Make URLs in docs hyperlinks

knoellle:
## Why? What?

Should fix the warnings seen in [this](https://github.com/HULKs/hulk/actions/runs/9820147790/job/27114417595#step:4:1007) CD run.

Fixes #

Check docs in CI instead of ignoring the warnings in CD

## How to Test

`cargo doc --workspace --no-deps` should no longer show the warnings.


oleflb (approved):
Noise

PR #1111, opened 2024-07-07
Higher ball candidates

knoellle:
## Why? What?

We noticed with the new kick striker positioning that it has a really hard time seeing the ball.
The ball was at the top of the bottom image, partially cut off.
However, there were no perspective grid candidates there because the generator stops as soon as the center of the circles would be above the upper image border.

With this change, the generator stops once the candidate row is completely outside the image.

Fixes #

## How to Test

Read above


oleflb (approved):
Gekauft

PR #1112, opened 2024-07-07
Fix Sacrificial Nao

### LINKED ISSUE #877: Opponent penalties for visual referee
When detecting the visual referee hand signal for Initial -> Ready, delay a few seconds before deciding.
If an opponent robot is penalized during this time (we get this information from the game controller), ignore the signal.

This assumes that whatever false positive we saw was picked up by the opposing team as well.

Vivituhh:
## Why? What?

The sacrificial lamb also changed to `Ready` when the opponent was penalized. This is fixed now.
Also removes unused variable `detection_times`.

Fixes #877 

## ToDo / Known Issues

## How to Test
Check normal behaviour: correct detection, sacrificial lamb starts after 5s, rest of the team after another 15s.
Check penalizing opponent in first 5s after detection: no nao should move.
Check penalizing of sacrificial lamb after it starts moving: rest of the team should not start moving.

### DIFF
--- a/crates/control/src/referee_pose_detection_filter.rs
+++ b/crates/control/src/referee_pose_detection_filter.rs
@@ -78,7 +78,6 @@ impl RefereePoseDetectionFilter {
             detected_above_arm_poses_queue: VecDeque::with_capacity(
                 *context.referee_pose_queue_length,
             ),
-            motion_in_standby_count: 0,
         })
     }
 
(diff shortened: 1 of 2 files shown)


MaikRe:
I think `last_majority_vote` and `majority_vote_detected_now` should be renamed to be a bit more similar and clear that they are both boolean values.
Maybe `last_majority_vote_result` and `current_majority_vote_result`?
Otherwise this works as expected!

MaikRe (approved):
Sacrifice! Sacrifice! Sacrifice!

PR #1113, opened 2024-07-07
Separate Messages

oleflb:
## Why? What?

Many of the fields for the VisualRefereeMessage are not used in StrikerMessages and vice versa. This PR separates the messages into two distinct types.

Note to reviewer: The player number check in role_assignment was removed since the message filter already takes care of that.

Fixes #

## ToDo / Known Issues

## How to Test

* Everything should work as before

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -13,7 +13,8 @@ use framework::{AdditionalOutput, MainOutput, PerceptionInput};
 use hardware::NetworkInterface;
 use linear_algebra::{Isometry2, Point2, Vector};
 use spl_network_messages::{
-    GameControllerReturnMessage, GamePhase, HulkMessage, Penalty, PlayerNumber, SubState, Team,
+    GameControllerReturnMessage, GamePhase, HulkMessage, Penalty, PlayerNumber, StrikerMessage,
+    SubState, Team,
 };
 use types::{
     cycle_time::CycleTime,
@@ -238,8 +239,8 @@ impl RoleAssignment {
             .into_values()
             .flatten()
             .filter_map(|message| match message {
-                Some(IncomingMessage::GameController(..)) | None => None,
-                Some(IncomingMessage::Spl(message)) => Some(message),
+                Some(IncomingMessage::Spl(HulkMessage::Striker(message))) => Some(message),
+                _ => None,
             })
             .peekable();
         if spl_messages.peek().is_none() {
@@ -327,15 +328,14 @@ impl RoleAssignment {
                             cycle_start_time,
                         )
                     };
-                    context
-                        .hardware
-                        .write_to_network(OutgoingMessage::Spl(HulkMessage {
+                    context.hardware.write_to_network(OutgoingMessage::Spl(
+                        HulkMessage::Striker(StrikerMessage {
                             player_number: *context.player_number,
                             pose: ground_to_field.as_pose(),
-                            is_referee_ready_signal_detected: false,
                             ball_position,
                             time_to_reach_kick_position: Some(*context.time_to_reach_kick_position),
-                        }))?;
+                        }),
+                    ))?;
                 }
             }
         }
@@ -374,7 +374,7 @@ fn process_role_state_machine(
     current_pose: Isometry2<Ground, Field>,
     detected_own_ball: Option<&BallPosition<Ground>>,
     primary_state: PrimaryState,
-    incoming_message: Option<&HulkMessage>,
+    incoming_message: Option<&StrikerMessage>,
     time_to_reach_kick_position: Option<Duration>,
     send_spl_striker_message: bool,
     team_ball: Option<BallPosition<Field>>,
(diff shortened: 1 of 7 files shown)


crates/control/src/role_assignment.rs
```diff
@@ -412,12 +412,6 @@ fn process_role_state_machine(
         }
     }
 
-    if let Some(message) = incoming_message {
-        if message.player_number == player_number {
-            return (current_role, false, team_ball);
-        }
-    }
```
schmidma, line 419: What is the reason to remove that? Is this no longer necessary?

crates/control/src/visual_referee_filter.rs
```diff
@@ -97,12 +99,13 @@ impl VisualRefereeFilter {
 
             // Initially a random visual referee decision
             let gesture =
-                VisualRefereeDecision::from_u32(self.random_state.gen_range(1..=13)).unwrap();
+                GestureVisualRefereeDecision::from_u32(self.random_state.gen_range(1..=13))
+                    .unwrap();
```
schmidma, line 103: Do we still need to send these gesture referee messages?

crates/control/src/role_assignment.rs
schmidma: ah, is this duplicated to the check in the message filter?

crates/spl_network/src/message_filter.rs
```diff
@@ -32,9 +32,10 @@ impl MessageFilter {
             IncomingMessage::GameController(source_address, message) => Some(
                 IncomingMessage::GameController(*source_address, message.clone()),
             ),
-            IncomingMessage::Spl(message) if message.player_number != *context.player_number => {
-                Some(IncomingMessage::Spl(*message))
-            }
+            IncomingMessage::Spl(
+                message @ HulkMessage::Striker(StrikerMessage { player_number, .. })
+                | message @ HulkMessage::VisualReferee(VisualRefereeMessage { player_number, .. }),
```
schmidma, line 37: Is there a syntactic way to deduplicate the `message @` tag?

crates/spl_network_messages/src/lib.rs
```diff
@@ -18,17 +18,34 @@ pub use game_controller_state_message::{
...
+pub enum HulkMessage {
+    Striker(StrikerMessage),
+    VisualReferee(VisualRefereeMessage),
+}
+
+impl Default for HulkMessage {
+    fn default() -> Self {
+        HulkMessage::Striker(StrikerMessage::default())
+    }
+}
```
schmidma, line 33: Why do we need `Default`?

crates/spl_network_messages/src/lib.rs
```diff
@@ -18,17 +18,34 @@ pub use game_controller_state_message::{
...
     pub pose: Pose2<Field>,
-    pub is_referee_ready_signal_detected: bool,
     pub ball_position: Option<BallPosition<Field>>,
     pub time_to_reach_kick_position: Option<Duration>,
 }
 
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+pub struct VisualRefereeMessage {
+    pub player_number: PlayerNumber,
+    pub is_referee_ready_signal_detected: bool,
```
schmidma, line 46: Isn't the mere existence of the message enough? Do we need an additional bool?

schmidma (changes requested):

crates/control/src/role_assignment.rs
oleflb: yes

crates/spl_network_messages/src/lib.rs
oleflb: Hmm, `IncomingMessage` and `Outgoing` message are outputs and therefore require `Default`. I could remove the default implementations of `HulkMessage`, ... but then there would need to be some hardcoded default in the `Default` implementation of `Incoming/OutgoingMessage`

crates/spl_network/src/message_filter.rs
oleflb: `message @ (...)` works to, done

crates/spl_network_messages/src/lib.rs
oleflb: you are absolutely right, removed

crates/control/src/visual_referee_filter.rs
oleflb: Are you sure about this? We expect the rules to include more visual referee stuff in the future right? We can remove it in a separate PR however, so that we could easily revert it in theory.

crates/spl_network_messages/src/lib.rs
schmidma: This may have to be improved in some next iteration of the framework. There are several `Default`s flying around that should not be there. Ok for now, but we should try to not implement traits for types that are not designed to implement these...

resolving for now

schmidma (approved):

PR #1117, opened 2024-07-08
print full backtrace if available

MaikRe:
## Why? What?

If something crashes due to an except we have no idea where this happened.
This prints this full backtrace if the `RUST_BACKTRACE` environment variable is set

Fixes #

## How to Test

Cause an exception or panic, `hulk.err` should show where it happened

### DIFF
--- a/crates/code_generation/src/execution.rs
+++ b/crates/code_generation/src/execution.rs
@@ -34,6 +34,7 @@ pub fn generate_run_function(cyclers: &Cyclers) -> TokenStream {
             {
                 let keep_running = keep_running.clone();
                 std::panic::set_hook(Box::new(move |panic_info| {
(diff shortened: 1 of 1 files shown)


oleflb:
This is still of by default, right (RUST_BACKTRACE not set)?
No numbers but I found this in the rust docs: 
"Capturing a backtrace can be a quite expensive runtime operation, so the environment variables allow either forcibly disabling this runtime performance hit or allow selectively enabling it in some programs."

MaikRe:
@oleflb see [#15](https://github.com/HULKs/meta-nao/pull/15) for the environment variable setting.
This is only intended for crashing of hulk service so performance shouldn't be an issue

schmidma (approved):

PR #1126, opened 2024-07-09
Use correct clock for foot bumper filter

knoellle:
## Why? What?

What it says. Was using SystemTime which is very very incorrect in replay.

Bidde immer alles richdich machen

Fixes #


oleflb (approved):
:+1:

PR #1131, opened 2024-07-10
ignore failing notification in `buffered_watch::Sender`

### LINKED ISSUE #1127: Buffered watch panics on sole writer
Notifying readers when dropping a sender guard panics if there are no more readers.

Failing testcase:
```rs
#[test]
fn sender_guard_dropped_without_receivers() {
    let (mut writer, reader) = channel(0);
    drop(reader);
    writer.borrow_mut();
}
```

schmidma:
## Why? What?

Failing to notify receivers if there are none anymore, it is not an error.
This ignores the error.

Fixes #1127

## ToDo / Known Issues

none

none

## How to Test

Not used in this fashion in `main` at the moment. But you can create a test case as described in #1127

### DIFF
--- a/crates/buffered_watch/src/sender.rs
+++ b/crates/buffered_watch/src/sender.rs
@@ -90,6 +90,6 @@ impl<'lock, T> Drop for SenderGuard<'lock, T> {
                 }
             }
         }
-        self.notifier.send(()).unwrap();
+        let _ = self.notifier.send(());
     }
 }


oleflb (changes requested):
What do you think of adding the testcase from #1127?

schmidma:
> What do you think of adding the testcase from #1127?

This is a good idea, does not hurt to have it. Even if it only checks that it doesn't panic.

... Done

oleflb (approved):

PR #1136, opened 2024-07-10
Warn if cycle takes too long

knoellle:
## Why? What?

Plays a sound if a cycler's node execution took too long.
Should help diagnose issues like we saw in yesterday's testgame.

Fixes #

## ToDo / Known Issues

Use bound closer to the actual cycle bound of each cycle.

## How to Test

### DIFF
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -530,7 +530,8 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers, mode: CyclerMode) -
                 #after_remaining_nodes
                 let recording_duration = recording_timestamp.elapsed().expect("time ran backwards");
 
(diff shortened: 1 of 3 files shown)


crates/code_generation/src/cyclers.rs
```diff
@@ -530,6 +530,14 @@ fn generate_cycle_method(cycler: &Cycler, cyclers: &Cyclers, mode: CyclerMode) -
                 #after_remaining_nodes
                 let recording_duration = recording_timestamp.elapsed().expect("time ran backwards");
 
+                if recording_duration.as_secs_f32() > 0.4 {
```
julianschuler, line 533: Where does this value come from? Consider factoring this out into a constant.

julianschuler (changes requested):
Not a full review yet, I still have to test this on the field.

julianschuler (approved):

PR #1143, opened 2024-07-10
Re-Initialize Microphone Device on failed read

### LINKED ISSUE #1125: Crash caused by microphone disconnect
During our testgame on 09-07-2024 the 37 fell and then crashed with following error code
```bash
Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
cycler thread Audio returned error: 
   0: failed to execute cycle of cycler `Audio`
   1: failed to execute cycle of `MicrophoneRecorder`
   2: failed to read from microphones
   3: failed to read audio data
   4: ALSA function 'snd_pcm_readi' failed with error 'EPIPE: Broken pipe'
   5: EPIPE: Broken pipe

Location:
   crates/hulk_nao/src/microphones.rs:54
```
Most likely the fall disconnected the microphone briefly

schmidma:
## Why? What?

Retries the read from microphones for `number_of_retries` times when a read fails. This hopefully fixes #1125.

## ToDo / Known Issues

none

none

## How to Test

I am actually not sure, we have to somehow trigger the microphone disconnect...


crates/hulk_nao/src/microphones.rs
```diff
@@ -17,32 +21,28 @@ pub struct Microphones {
...
-                .wrap_err("failed to set hardware parameters")?;
-        }
-        device.prepare().wrap_err("failed to prepare device")?;
+        let device = create_device(&parameters).wrap_err("failed to create device")?;
         Ok(Self { device, parameters })
     }
 
-    pub fn read_from_microphones(&self) -> Result<Samples> {
+    pub fn retrying_read(&mut self) -> Result<Samples> {
+        let number_of_retries = 5;
```
knoellle, line 29: You defined this in hardware parameters but are not using it.

crates/hulk_nao/src/microphones.rs
```diff
@@ -17,32 +21,28 @@ pub struct Microphones {
...
+        for _ in 0..number_of_retries {
+            match self.read() {
+                Ok(samples) => return Ok(samples),
+                Err(error) => {
+                    warn!("failed to read from microphones: {error:#?}");
+                    self.device =
+                        create_device(&self.parameters).wrap_err("failed to create device")?;
+                }
+            }
+        }
```
knoellle, line 39: Might be fine, not sure, but if this fails repeatedly in the future, maybe add a sleep in the error branch here before creating the new device.

Also, you `?` return the error if device creation fails, I think its unlikely that after a disconnect, reading fails but opening the device anew does not.
You probably want to `continue` instead if device creation fails.

schmidma: Yes, good idea, added a sleep duration and only warn on fail to open device

crates/hulk_nao/src/microphones.rs
schmidma: oooooops :sweat_smile: :rocket: 

Fixed

knoellle (approved):
LGTM

PR #1156, opened 2024-07-11
Prevent zero measurement noise

### LINKED ISSUE #1039: Covariance not invertible
... a panic that occured during the test game on 2024-05-29

knoellle:
## Why? What?

See issue for details, adding epsilon to the measurement distance in all obstacle filter updates to prevent crashes.

Fixes #1039 

## How to Test

#1113 fixes #1039 symtomatically, making it hard to reproduce this bug.


oleflb (approved):
good catch

Issue #1115, opened 2024-07-07
What to do with the output from calibration?


tuxbotix:
With #798 being merged, an `AdditionalOutput` of `CalibrationController` streams the latest correction values.

The question is how to save these values into the parameters.json?

* Can we directly write to parameters?
* If not, perhaps a twix panel can be built to view the corrections and apply to the nao and save to disk (i.e. modified version of manual extrinsic panel)
* What does other teams do?

PR #1159, opened 2024-07-11
add toggle between old calibration mode and new calibration mode

MaikRe:
## Why? What?

I think completely removing the old calibration mode this close to an event could prove fatal if the new one isn't full functional #1115 or if and when people want to do manual adjustments, either during an event, or in the future to manually break camera matrices on purpose as we have done before for testing.
This adds a simple boolean toggle between the old and new calibration.

Feel free to suggest better naming

Fixes #

## How to Test

Setting the bool to false uses the new calibration which does (something) I guess?
Setting to true should make robot behave as before, i.e. stand up straight with the head stiffness at 0.

### DIFF
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1530,6 +1530,7 @@
     "look_at_dispatch_delay": {
       "nanos": 0,
       "secs": 1
-    }
+    },
+    "use_stand_head_unstiff_calibration": false
   }
 }
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1531,6 +1531,6 @@
       "nanos": 0,
       "secs": 1
     },
-    "use_stand_head_unstiff_calibration": false
+    "use_stand_head_unstiff_calibration": true
   }
 }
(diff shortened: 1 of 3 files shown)


etc/parameters/default.json
```diff
@@ -1530,6 +1530,7 @@
     "look_at_dispatch_delay": {
       "nanos": 0,
       "secs": 1
-    }
+    },
+    "use_stand_head_unstiff_calibration": false
```
oleflb, line 1534: please make true default for now until we make the switch to automatic calibration

oleflb (changes requested):

etc/parameters/default.json
MaikRe: done

oleflb (approved):
nice :fire:

tuxbotix:
Thanks for adding this feature, I fully agree with your reasoning. Manual mode was unintentionally disabled when I introduced the new controller.
And the controller still works fine when the parameter is toggled.

PR #1161, opened 2024-07-12
Penalty keeper CenterJump

### LINKED ISSUE #997: Penalty Shootout Keeper

larskna:
## Why? What?

We added the WideStance to the penalty keeper für Center balls for to win better

Fixes #997 

Squat Motion improvments

## How to Test

Shot the Ball in the Penalty in the Center of the Goal


crates/control/src/penalty_shot_direction_estimation.rs
```diff
@@ -25,8 +26,8 @@ pub struct CreationContext {}
 #[context]
 pub struct CycleContext {
     field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
-    moving_distance_threshold:
-        Parameter<f32, "penalty_shot_direction_estimation.moving_distance_threshold">,
+    penalty_shot_paramters:
```
oleflb, line 29: typo

etc/parameters/default.json
```diff
@@ -1042,7 +1042,9 @@
     }
   },
   "penalty_shot_direction_estimation": {
-    "moving_distance_threshold": 0.2
+    "moving_distance_threshold": 0.2,
+    "minimum_velocity": -0.5,
```
oleflb, line 1046: This disables the velocity check, right?

oleflb (changes requested):
Can't test, just code review

etc/parameters/default.json
larskna: We dont have a velocity check in the main right now, this creates one

crates/control/src/penalty_shot_direction_estimation.rs
larskna: fixed

oleflb (approved):

PR #1167, opened 2024-07-12
arms up for goalkeeper during set for penalty kicks

### LINKED ISSUE #1164: Precise rule adhesion for penalty keeper
In interest of adhering to the penalty shootout rules following things are discovered:
- When doing an ingame penalty kick the transition is from `Ready -> Set` followed by `Playing`, when doing a penalty shootout the transition is from 'Penalized -> Set'
- In the `Penalized` state we **must** be in the normal stand position
- In `Set` we are **not** allowed to "move [our] legs", i.e. sitting down in this moment would be considered illegal. However we can freely move our arms

Current idea is that in the set phase we already move our arms upwards so they are pointing up above the head.
Then when we hear the whistle for `Playing` we only need to sit down, which should be relatively stable since the arms no longer need to swing in any direction.

Keeping the arms above the head in a straight line prevents us from being penalized for `WideStance`

MaikRe:
## Why? What?

The arms up part of `ArmsUpSquat` either takes a long time or is unstable. We can take advantage of the permission to move our arms during set in order to move the arms up then and only have to sit down during the actual playing phase, increasing speed and stability.
Additionally or current `ArmsUpSquat` technically counts as `WideStance` and results in a standard removal penalty after 10 seconds.
The new ArmsUpStand and ArmsUpSquat pull the arms in closer and are separate motions.

Fixes #1164 

## ToDo / Known Issues

Depends on #1161 so that the motion can be properly adjusted for the wide stance

## How to Test

For both the penalty kick and penalty shootout the goalkeeper should now raise its arms when the game state transitions into set.
For the penalty kick this happens from ready to set, during the penalty shootout this happens when the player is selected as the goalkeeper.
When the whistle is blown the robot should now only need to sit down


crates/control/src/behavior/stand.rs
```diff
@@ -11,13 +11,16 @@ use types::{
...
     field_dimensions: &FieldDimensions,
+    role: &Role,
 ) -> Option<MotionCommand> {
     match world_state.robot.primary_state {
         PrimaryState::Initial => Some(MotionCommand::Stand {
             head: HeadMotion::ZeroAngles,
         }),
         PrimaryState::Set => {
             let ground_to_field = world_state.robot.ground_to_field?;
+            let mut kicking_team_in_set = Team::Uncertain;
```
oleflb, line 22: I would prefer this to be a separate match on the filtered_game_controller_state only checking the required variant instead of having mutable variables.

crates/control/src/motion/arms_up_stand.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Default)]
+pub struct MainOutputs {
+    pub arms_up_stand_joints_command: MainOutput<MotorCommands<Joints<f32>>>,
+}
+
+impl ArmsUpstand {
+    pub fn new(context: CreationContext<impl PathsInterface>) -> Result<Self> {
+        let paths = context.hardware_interface.get_paths();
+        Ok(Self {
+            interpolator: MotionFile::from_path(paths.motions.join("arms_up_stand.json"))?
```
oleflb, line 42: is it enough to use dispatching interpolator instead?

crates/control/src/motion/motion_selector.rs
```diff
@@ -114,6 +115,7 @@ fn transition_motion(
         (MotionType::StandUpSitting, _, MotionType::FallProtection, _) => {
             MotionType::StandUpSitting
         }
+        (MotionType::ArmsUpStand, _, MotionType::FallProtection, _) => MotionType::ArmsUpStand,
```
oleflb, line 118: u sure about that?

crates/control/src/motion/motion_selector.rs
```diff
@@ -122,6 +124,7 @@ fn transition_motion(
         (_, _, MotionType::FallProtection, _) => MotionType::FallProtection,
         (_, _, MotionType::WideStance, _) => MotionType::WideStance,
         (_, _, MotionType::CenterJump, _) => MotionType::CenterJump,
+        (MotionType::ArmsUpStand, _, _, false) => MotionType::ArmsUpStand,
```
oleflb, line 127: Why is this required? Isnt this a default transition?

oleflb (changes requested):

crates/control/src/motion/motion_selector.rs
MaikRe: If the robot is lifted up by the referee during arms up stand he goes back to stand which we do not want

crates/control/src/behavior/stand.rs
MaikRe: Done

crates/control/src/behavior/stand.rs
```diff
@@ -50,13 +51,32 @@ pub fn execute(
...
+                    },
+                }),
+                _ => Some(MotionCommand::Stand {
+                    head: HeadMotion::LookAt {
+                        target,
+                        image_region_target: Default::default(),
+                        camera: None,
+                    },
+                }),
+            }
```
oleflb, line 79: so doch net... stick to the variables you had, but don't make them mutable and assign them using a match statement at creation

MaikRe: Doughne

oleflb (approved):
Second attempt genehmigt

PR #1168, opened 2024-07-13
Disbale obstacles in penalty shootout

Vivituhh:
## Why? What?

Sometimes the robot falsely sees obstacles right in front of it during the penalty shootout phase and then kicks the ball around them (and not in the goal). Therefore they are disabled now during the penalty shoothot, but still drawn in twix.

One fix to improve penalty shootout #546.

Fixes #

## ToDo / Known Issues

- [x] disable also for Penalty Kick substate

## How to Test

Test normal behavior in normal game phases and also test behavior in penalty shootout with obstacles. Should kick through obstacles.


oleflb (approved):
Nice

PR #1176, opened 2024-07-14
Add parameters for walk speed of actions

### LINKED ISSUE #1152: Configure walking speed per role / behaviour in default.json
Move the hard-coded speed selections from #1150 into the `default.json`

Narcha:
## Why? What?

Fixes #1152

## How to test

- read the code
- use twix to change the walk speed of the robot's current action, check if the robot accelerates/slows down


oleflb (approved):
LGTM

PR #1177, opened 2024-07-14
disallow instant kicks in penalty shootout and penalty kick

vivi-enne:
## Why? What?

Tests for penalty shootout looked better without instant kicks. In my opinion a little jockeling is fine in this phase.
More small changes to addresses #546.

Fixes #

## ToDo / Known Issues

- [ ] should be tested again on the field

## How to Test
Use GameController in Penalty Shootout with a HULKs striker and compare main against this.

### DIFF
--- a/crates/control/src/kick_target_provider.rs
+++ b/crates/control/src/kick_target_provider.rs
@@ -7,7 +7,7 @@ use framework::MainOutput;
 use geometry::{circle::Circle, line_segment::LineSegment, two_line_segments::TwoLineSegments};
 use linear_algebra::{distance, point, Isometry2, Point2};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


crates/control/src/kick_target_provider.rs
```diff
@@ -154,6 +154,20 @@ fn collect_kick_targets(
         })
     );
 
+    let is_penalty_shoot = matches!(
```
oleflb, line 157: Prefer matching both patterns simultaneously using | and only one matches!

oleflb (changes requested):
Minor thing, also still requires testing

oleflb:
@Vivituhh do you think we still want this?

Vivituhh:
If our kicks are better now, we can close this. One could implement it again if needed

Issue #998, opened 2024-05-04
Check ideas for Line Detection


oleflb:
Lets check what other teams do:
- Check what BHuman/HTWK/... do
- Read again Justus BA

h3ndrk:
Current ideas:

- Readd horizontal segments (we had them once, look into Git and port them back to current `main`)
    - Reduce maximum line segment length since we don't need to accept extremely diagonal lines anymore then
- Segmentation: Instead of finding the edge (maximum change in difference) in the stride pixels, we may want to find it in real pixels to increase segment boundary accuracy
- Segmentation: First calculate field color on the sample points for the segment color and then do some majority thesholding whether something is field color (currently we are calculating a median color and then calculate a single field color)
- Segmentation: Calculate median of Y and then take the whole pixel color instead of also calculating the median for the other channels
- Segmentation: Place horizontal and vertical segments over each other to detect higher-level regions
- Segmentation: Respect locality of segments: Currently all segments are evaluated individually for field color, but it may make sense to cluster neighboring regions. B-Human also does this.
- Implement custom white-balance
- Field Color: Use a different color space for field color, for example rotate the intensity vector (currently it is only pointing how green something is but maybe we need to measure whether something is green/blueish...)
- Use amount of line segments in localization scoring
- Dilation, open closing (I forgot what that means)
- Train a Support Vector Machine for tuning vision parameters

tuxbotix:
http://www.ipol.im/pub/art/2012/gjmr-lsd/  This was also an interesting approach, available on OpenCV (so we could test it via the Rust bindings).

Performance-wise, not super fast if you feed the whole image, but you don't need most of the preprocessing needed for an approach like Canny. I was testing this out to see if I can do the calibration with Line detection.

oleflb:
@julianschuler and I tested LSD already, it also failed to find lines under kassel lighting conditions

schmidma:
What is the plan for this until RC24? Are you able to work on it? Can the issue be split up into smaller parts? This would allow multiple people to work on improving line detection

h3ndrk:
[...]

PR #1179, opened 2024-07-14
Sample more pixels for color calculation of segments

h3ndrk:
## Why? What?

We are currently only sampling up to 3 pixels for each segment in the image segmenter. My assumption is that if we sample more pixels, we should get a more stable color for segments. Therefore, I now sample every pixel with a stride in the segment instead of a fixed amount of pixels.

In addition, previously we calculated a median in all color channels individually. This PR only selects the pixel with the median Y value and uses it as a whole.

Related to #998 

## ToDo / Known Issues

#998

#998

## How to Test

Deploy to a NAO or replay through a recording and observe that the segment colors are still sensible.

### DIFF
--- a/crates/vision/src/image_segmenter.rs
+++ b/crates/vision/src/image_segmenter.rs
@@ -1,4 +1,7 @@
-use std::time::{Duration, Instant};
+use std::{
+    ops::{Add, Range},
+    time::{Duration, Instant},
+};
 
 use color_eyre::Result;
 use projection::{camera_matrix::CameraMatrix, horizon::Horizon};
@@ -236,10 +239,8 @@ fn new_vertical_scan_line(
                 fix_previous_edge_type(&mut segments);
                 break;
             }
-            segments.push(set_color_in_vertical_segment(
-                segment,
-                image,
-                position,
+            segments.push(set_field_color_in_vertical_segment(
+                set_color_in_vertical_segment(segment, image, position),
                 field_color,
             ));
         }
@@ -254,10 +255,8 @@ fn new_vertical_scan_line(
         field_color: Intensity::Low,
     };
     if !segment_is_below_limbs(position as u16, &last_segment, projected_limbs) {
-        segments.push(set_color_in_vertical_segment(
-            last_segment,
-            image,
-            position,
+        segments.push(set_field_color_in_vertical_segment(
+            set_color_in_vertical_segment(last_segment, image, position),
             field_color,
         ));
     }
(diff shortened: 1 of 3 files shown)


crates/vision/src/image_segmenter.rs
```diff
@@ -315,94 +314,62 @@ fn pixel_to_edge_detection_value(
...
     segment.field_color = field_color.get_intensity(segment.color);
     segment
 }
 
+fn sample_image_pixels_vertically(
+    image: &YCbCr422Image,
+    x: u32,
+    y: Range<u32>,
+    stride: u32,
+) -> Vec<YCbCr444> {
```
schmidma, line 350: This is doing a heap allocation in a hot loop. We should sanity check performance. Maybe using the replay after the test game

crates/vision/src/image_segmenter.rs
```diff
@@ -315,94 +314,62 @@ fn pixel_to_edge_detection_value(
...
+    segment.color = *median_pixel;
+    segment
+}
+
+fn set_field_color_in_vertical_segment(mut segment: Segment, field_color: &FieldColor) -> Segment {
     segment.field_color = field_color.get_intensity(segment.color);
     segment
 }
 
+fn sample_image_pixels_vertically(
```
schmidma, line 345: What do you think about making this a function of the `Image`? To hide a lot of implementation detail, and not making the `coordinates_to_buffer_index` function `pub`.
Maybe consider replacing "vertically" with something like column?

schluis:
I'm not sure if the improvement of this justifies the computational overhead 🤔 
I think a median of max. 5 should be sufficient to reduce noise.

knoellle:
> I'm not sure if the improvement of this justifies the computational overhead 🤔 I think a median of max. 5 should be sufficient to reduce noise.

Coarse analysis, but I just had a look at a replay I had sitting around from Kassel, one from last week, and one from the test game today, which AFAIK included this branch.
There doesn't seem to be any discernible difference in vision cycler execution times between them, and so I think we should disregard computational cost and just focus on the results.
Disclaimer: I have not looked at how good the results are compared to main.

schluis:
> > I'm not sure if the improvement of this justifies the computational overhead 🤔 I think a median of max. 5 should be sufficient to reduce noise.
> 
> Coarse analysis, but I just had a look at a replay I had sitting around from Kassel, one from last week, and one from the test game today, which AFAIK included this branch. There doesn't seem to be any discernible difference in vision cycler execution times between them, and so I think we should disregard computational cost and just focus on the results. Disclaimer: I have not looked at how good the results are compared to main.

Ah interesting. My comment relates more to the effectiveness of a larger median for noise reduction, which I consider to be quite low. I did some tests with my camera a few years ago, layering multiple images and computing the median between them, which drastically increased computational time without adding much benefit for an increasing number of images. Maybe we could repeat the same with uncompressed pictures from the Nao to check this.

crates/vision/src/image_segmenter.rs
h3ndrk: No allocations anymore because we now calculate averages instead of medians.

crates/vision/src/image_segmenter.rs
h3ndrk: - `YCbCr422Image::buffer()` is now private
- `YCbCr422Image::coordinates_to_buffer_index()` is now private
- The Image Color Select panel uses a new public method `YCbCr422Image::iter_pixels()` which returns the 444 pixels
- Iteration in Image Segmenter is now done in coordinate-domain instead of buffer-index-domain

tuxbotix:
Is it absolutely nessesary to make the buffer private?
If so, is there an alternative way to access the values to say create an image with `imageproc` crate?

The calibration with center circle uses `buffer()` to create the nessesary types to perform edge detection.

schmidma (approved):

PR #1182, opened 2024-07-15
Kick-Decision anticipation

schmidma:
## Why? What?

Walking plans and acts in the matter of "steps". That means, new decisions are discrete and a new step can only start as soon as the current step ends. The next step is then planned in the idea of starting from zero. That means, if we request a zero step, and the current (just ended) step was not a zero step, then the ground frame travels at least the distance to the zero state.

What does that mean for deciding for kicks:
If we want to now whether we can kick the ball or even decide on a kick, we have to anticipate, that in the instant of deciding, the ground frame is still behind the support foot (assuming forward travel) and is going to travel some distance anyway. That is, we have to decide on the kick relative to the support foot.

To fix this issue, step planning uses the formerly called `walk_return_offset` which describes the distance ground will travel up until the zero step. I refactored this to a properly tagged framed transform called `Isometry2<Ground, UpcomingSupport>`. Step planning and also all kick decisions now use this transform to decide for kicks.

Additionally, I think there was a bug of subtracting the `walk_return_offset` in the step planner, as a mere component-wise subtraction does not care for rotating to the `From` frame correctly.

## ToDo / Known Issues

none

none

## How to Test

Look at step planning and kick decision output in a replay of a robot dribbling the ball. There currently is no other way than just debugging it manually... ...

### DIFF
--- a/crates/control/src/behavior/dribble.rs
+++ b/crates/control/src/behavior/dribble.rs
@@ -1,4 +1,4 @@
-use coordinate_systems::Ground;
+use coordinate_systems::{Ground, UpcomingSupport};
 use geometry::look_at::LookAt;
 use linear_algebra::{Isometry2, Point, Pose2};
 use types::{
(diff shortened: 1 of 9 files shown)


oleflb (commented):
code lgtm, will test now

oleflb (dismissed):
Schaut mir gut

oleflb (approved):

PR #1192, opened 2024-07-15
Deaf ref

ThagonDuarte:
## Why? What? 

Removes the poses ear keypoints from confidence filtering and demotes the eyes to the lower confidence threshold. Also adds a color differentiation between the importance of the points in the twix pose detection overlay, based on the confidence threshold of that point.
The ears are sometimes on the side of the refs head that is not facing the nao and thus gets really poor confidence values.
As this keypoint is not needed for the pose interpretation, it should be removed. The eyes have been demoted to the overall confidences, as it is also not needed for the pose interpretation.  

## How to Test

- Setup visual referee, check the confidence values and whether or not the pose is filtered out or not in twix.


knoellle (changes requested):

knoellle (approved):
LGTM

PR #1204, opened 2024-07-15
only Forward kicks as penalty kicks

Vivituhh:
## Why? What?

@pejotejo and me decided it is best to only allow Forward kicks in the penalty situation, as they are most precise and also strike a goal. 
Also increases the kick strength for penalty kicks to 2.

More improvement for #546.

Fixes #

## ToDo / Known Issues

## How to Test

Test in Penalty Shootout or Penalty Kick, and see only Forward kicks. Can also be tested with other kicks if want clearer difference.

### DIFF
--- a/crates/control/src/kick_target_provider.rs
+++ b/crates/control/src/kick_target_provider.rs
@@ -219,28 +219,21 @@ fn collect_kick_targets(
                 .collect(),
                 true,
             )
-        } else if is_penalty_shot {
-            (
-                generate_goal_line_kick_targets(field_dimensions, field_to_ground)
-                    .into_iter()
-                    .map(|kick_target| KickTargetWithKickVariants {
-                        kick_target,
-                        kick_variants: collect_kick_targets_parameters
-                            .penalty_kick_kick_variants
-                            .clone(),
-                    })
-                    .collect(),
-                true,
-            )
         } else {
             (
                 generate_goal_line_kick_targets(field_dimensions, field_to_ground)
                     .into_iter()
                     .map(|kick_target| KickTargetWithKickVariants {
                         kick_target,
-                        kick_variants: collect_kick_targets_parameters
-                            .goal_line_kick_variants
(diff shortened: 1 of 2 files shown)


crates/control/src/kick_target_provider.rs
```diff
@@ -163,6 +167,16 @@ fn collect_kick_targets(
             ..
         })
     );
+    let is_penalty_shoot = matches!(
+        filtered_game_controller_state,
+        Some(FilteredGameControllerState {
+            game_phase: GamePhase::PenaltyShootout { .. },
+            ..
+        }) | Some(FilteredGameControllerState {
+            sub_state: Some(SubState::PenaltyKick),
```
schluis: Also check `KickingTeam`?

crates/control/src/kick_target_provider.rs
```diff
@@ -203,6 +219,19 @@ fn collect_kick_targets(
...
+                    .into_iter()
+                    .map(|kick_target| KickTargetWithKickVariants {
+                        kick_target,
+                        kick_variants: collect_kick_targets_parameters
+                            .penalty_kick_kick_variants
+                            .clone(),
+                    })
+                    .collect(),
+                true,
+            )
```
schluis, line 234: Suggestion: Merge this with the else below like this:
```
        } else {
            (
                generate_goal_line_kick_targets(field_dimensions, field_to_ground)
                    .into_iter()
                    .map(|kick_target| KickTargetWithKickVariants {
                        kick_target,
                        kick_variants: if is_penalty_shot {
                            collect_kick_targets_parameters
                                .penalty_kick_kick_variants
                                .clone()
                        } else {
                            collect_kick_targets_parameters
                                .goal_line_kick_variants
                                .clone()
                        },
                    })
                    .collect(),
                true,
            )
        };

```

schluis (changes requested):
Nice

PR #1207, opened 2024-07-15
Ball Filter - Throw out by validity

oleflb:
## Why? What?

In the refactor of the ball filter I forgot to throw out hypotheses by validity criterium. This PR reintroduces the check.

Fixes #

## ToDo / Known Issues

## How to Test

* Check hypotheses are removed from the filter if their validity drops low


knoellle (approved):
LGTM

PR #1208, opened 2024-07-15
WideStance Left and right 

larskna:
## Why? What?

We want to jump to the right and left whith the WideStance, so we can better catch the blall 
Its only for the keeper
Fixes #

## ToDo / Known Issues

The Motienfile is in this moment indentical with the WideStance. That is only the .json file 

## How to Test

You have to kick the ball to the Keeper (playernumber 1) and look


crates/control/src/behavior/defend.rs
```diff
@@ -58,24 +58,38 @@ impl<'cycle> Defend<'cycle> {
...
+            (true, _, _) => Some(MotionCommand::WideStance {
+                direction: JumpDirection::Center,
+            }),
+            (false, true, _) => Some(MotionCommand::WideStance {
+                direction: JumpDirection::Left,
+            }),
+            (false, _, true) => Some(MotionCommand::WideStance {
+                direction: JumpDirection::Right,
+            }),
+            (false, false, false) => None,
```
schluis, line 92: I think a simple if-else statement is completely sufficient. This is basically a glorified if-else. Other opinions? @h3ndrk

h3ndrk: Looks like it, but not sure about the `_` in line 89 :thinking:

schluis: I'm sure. All cases are mutually exclusive

oleflb:
@larskna whats the state here?

larskna:
It can be close. 
it was the first iteration of the keeper jump. I will do a new PR for that

PR #1229, opened 2024-07-15
Fix saving camera matrix

knoellle:
## Why? What?

The save button previously wrapped the calibration parameters in a spurious "parameters" object when writing to disk. 
Now it doesn't.

Fixes #

## How to Test

- Move sliders in manual calibration panel
- Hit save
- ?
- Profit


PatrickGoettsch (approved):
LGTM

PR #1232, opened 2024-07-16
Reduce visual referee threshold

ThagonDuarte:
## Why? What?
Reduce `minimum_visual_referee_keypoint_confidence` to improve visual referee performance.


JackyBloxx (approved):
look good

PR #1236, opened 2024-07-16
Allow to disable instant kicks

schmidma:
## Why? What?

Only retain the enabled kick variants. This was done for "normal" kick decisions but the instant ones were not filtered.

## ToDo / Known Issues

none

none

## How to Test

disable one variant of a kick (e.g. the turn kick) via the parameters and check that the robot is not doing this kick anymore. Or that no instant kick variants of this type are produced.


schluis (approved):
lgtm

PR #1237, opened 2024-07-16
Reduce liquidity

knoellle:
## Why? What?

Remove superfluous parameters and add a check.

Fixes #

## How to Test

Add an unused field anywhere in the default.json or even a robot/location specific config file and upload.
Should error if the modified config file is actually used by the robot.


crates/parameters/src/directory.rs
```diff
@@ -48,6 +51,7 @@ pub enum SerializationError {
 pub fn deserialize<Parameters>(
     parameters_root_path: impl AsRef<Path>,
     hardware_ids: &Ids,
+    allow_superfluous_fields: bool,
```
schmidma, line 54: What do you think about replacing the bool with an enum. To also make this argument readable at call-site?

schmidma (changes requested):
minor idea

schmidma (approved):

PR #1241, opened 2024-07-16
refactor `kick_target_provider`, disable obstacle detection for kick off

Vivituhh:
## Why? What?

In last test game the kick off targets were not placed as supposed, because they tried to avoid obstacles. We decided to ignore obstacles in kick off, as no opponent robot is allowed in the center circle. Only own team members are allowed until after the kick and this might also help with the indirect kick rule.

Fixes #

## ToDo / Known Issues

Issue from testgame 16.July.2024

none

## How to Test

With game controller test kick off situations and watch kick targets to be placed on the center line and no other kick targets should be changed.

### DIFF
--- a/crates/control/src/kick_target_provider.rs
+++ b/crates/control/src/kick_target_provider.rs
@@ -12,12 +12,26 @@ use types::{
     field_dimensions::FieldDimensions,
     filtered_game_controller_state::FilteredGameControllerState,
     filtered_game_state::FilteredGameState,
-    kick_target::{KickTarget, KickTargetWithKickVariants},
+    kick_target::{KickTarget, KickTargetWithKickVariants, PlayingSituation},
     motion_command::KickVariant,
     obstacles::Obstacle,
     parameters::FindKickTargetsParameters,
     world_state::BallState,
 };
+
+trait WithVariants {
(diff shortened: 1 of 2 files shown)


crates/control/src/kick_target_provider.rs
```diff
@@ -146,188 +160,79 @@ fn collect_kick_targets(
...
-            sub_state: Some(SubState::PenaltyKick),
-            kicking_team: Team::Hulks,
-            ..
-        })
-    );
-
-    let (kick_opportunities, allow_instant_kicks): (Vec<_>, _) =
-        if is_own_kick_off && is_not_free_for_opponent {
-            (
+    let (mut kick_opportunities, allow_instant_kicks, respect_obstacles): (Vec<_>, _, _) =
```
schmidma, line 169:
```suggestion
    let (mut kick_opportunities, allow_instant_kicks, respect_obstacles) =
```

schmidma (changes requested):

crates/control/src/kick_target_provider.rs
Vivituhh: Done

schmidma (approved):

PR #1247, opened 2024-07-16
Better defender corner kick positioning

ThagonDuarte:
## Why? What?

Improves the defender positioning for opponent corner kicks. Previously, the defender would be clamped to the goal line. This positioning overlaps with the Striker and is thus not desirable. This offsets the defender x position in these situations.   

Depends on #1244 

## How to Test

- Serve the behavior simulator with the `corner_kicks.lua` test and confirm the defender position for the opponent corner kicks.


oleflb (approved):
Nice

PR #1248, opened 2024-07-16
slow walking in penalty shootout

JackyBloxx:
## Why? What?

slower walking in penalty shootout for mor precision on the shoot
Fixes #

## How to Test

look if thte penalty shootout is better


knoellle (approved):
Merge and pray

PR #1249, opened 2024-07-16
New Calib data

PatrickGoettsch:
## Why? What?

*Describe here why you created this PR and what it does*

Fixes #

## ToDo / Known Issues

## How to Test


etc/parameters/rc24a/head.P0000074A05S98U00019.json
```diff
@@ -0,0 +1,20 @@
+{
+  "parameters": {
```
schmidma, line 2: this has to be removed

oleflb (changes requested):
This looks like all jsons got formatted, but the calibration data is the same compared to main. Also there are several line endings missing at end of files

PR #1252, opened 2024-07-16
Tune outdoor field

h3ndrk:
- Parameters for outdoor field
- Now we only have an indoor and outdoor location


julianschuler (approved):
LGTM

PR #1253, opened 2024-07-16
Ball filter - Not so Smart Update

### LINKED ISSUE #789: Ball Validity
Replace the validity computation by a measure that directly compares measurements to the ball model. In return we can spawn new hypotheses on every measurement and prune hypotheses which don't have a high enough correlation with measurements.
This would make the filter model much more reactive to balls that move non-linear, i.e. when kicked.

### LINKED ISSUE #993: Forget about ball modes
We believe too long in ball models in the ball filter, this makes it hard to track balls that start/stop moving (under/overshoot). We need to forget about balls in that state and spawn new candidates in order to handle the non-linear movement.

oleflb:
## Why? What?

Changes the mode implementation of the ball filter to be more reactive to changing ball models.
Since many hypotheses are used in the filter now, a more sophisticated matching algorithm is used to match percepts to hypotheses.

Fixes #789 Fixes #993

## ToDo / Known Issues

## How to Test


crates/control/src/ball_filter.rs
```diff
@@ -164,19 +189,26 @@ impl BallFilter {
...
-            |hypothesis1: &BallHypothesis, hypothesis2: &BallHypothesis| {
-                let ball1 =
-                    hypothesis1.choose_ball(filter_parameters.resting_ball_velocity_threshold);
-                let ball2 =
-                    hypothesis2.choose_ball(filter_parameters.resting_ball_velocity_threshold);
-
-                distance(ball1.position, ball2.position)
-                    < filter_parameters.hypothesis_merge_distance
+            |_hypothesis1: &BallHypothesis, _hypothesis2: &BallHypothesis| {
+                // distance(
```
ThagonDuarte: The problem of merging moving and resting hypotheses could be addressed by simply matching the `BallMode`s of the Hypotheses. In the case of a two resting hypotheses, the positions can simply be compared. In the case of moving hypotheses one could compare the position of the ball hypothesis and the direction as well as magnitude of the velocity.

crates/control/src/ball_filter.rs
```diff
@@ -125,34 +132,52 @@ impl BallFilter {
...
-                        detection_time,
-                        mean_position,
-                        Matrix4::from_diagonal(&filter_parameters.noise.initial_covariance),
-                        Matrix2::from_diagonal(&filter_parameters.noise.initial_covariance.xy()),
-                    )
-                }
             }
         }
 
         let is_hypothesis_valid = |hypothesis: &BallHypothesis| {
```
ThagonDuarte, line 179: For me this inplace definition of the filtering criterion was kinda confusing at first. Maybe we could refactor these into separate functions?  Same with `should_merge_hypothesis`

ThagonDuarte (changes requested):
Initial code review...

crates/ball_filter/src/hypothesis.rs
```diff
@@ -17,104 +20,132 @@ use types::{
...
-    pub fn new(
-        moving_hypothesis: MultivariateNormalDistribution<4>,
-        resting_hypothesis: MultivariateNormalDistribution<2>,
-        last_seen: SystemTime,
-    ) -> Self {
+    pub fn new(hypothesis: MultivariateNormalDistribution<4>, last_seen: SystemTime) -> Self {
         Self {
-            moving: moving_hypothesis,
-            resting: resting_hypothesis,
+            mode: BallMode::Moving(hypothesis),
```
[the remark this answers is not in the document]
oleflb: because only the state transition moving to resting is implemented using the kalman filter (on purpose, since the transition ready to moving is not continuous in the real world)

crates/control/src/ball_filter.rs
oleflb: discussed, not needed

crates/control/src/ball_filter.rs
oleflb: decided to do this only for resting since all moving hypotheses eventually become resting

crates/control/src/ball_filter.rs
```diff
@@ -246,33 +275,29 @@ impl BallFilter {
             .filter(|hypothesis| {
                 hypothesis.validity >= context.ball_filter_configuration.validity_output_threshold
             })
-            .map(|hypothesis| hypothesis.choose_ball(velocity_threshold).position)
+            .map(|hypothesis| hypothesis.position().position)
             .collect::<Vec<_>>();
 
         Ok(MainOutputs {
             ball_position: filtered_ball.into(),
             removed_ball_positions: removed_ball_positions.into(),
```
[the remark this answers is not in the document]
oleflb: you are right. removed now

crates/control/src/ball_filter.rs
```diff
@@ -125,34 +132,52 @@ impl BallFilter {
...
+            for (hypothesis, assigned_percept) in self
+                .ball_filter
+                .hypotheses
+                .iter_mut()
+                .zip_eq(assignment.iter())
+            {
+                if let Some(assigned_percept) = assigned_percept {
+                    let mahalanobis_distance = -assigned_percept.cost;
+                    if mahalanobis_distance > filter_parameters.maximum_matching_cost {
+                        hypothesis.validity *= 1.0 / 7.0;
```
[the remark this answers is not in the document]
oleflb: is a parameter now

crates/control/src/ball_filter.rs
```diff
@@ -186,7 +219,7 @@ impl BallFilter {
         );
 
         let filter_parameters = context.ball_filter_configuration;
-        let removed_hypotheses = self.advance_all_hypotheses(
+        let _removed_hypotheses = self.advance_all_hypotheses(
```
ThagonDuarte, line 222: Should we perhaps just remove this returned `Vec<BallHypothesis>` from all functions returning the removed hypothsis?

ThagonDuarte: What do you think @oleflb?

crates/control/src/ball_filter.rs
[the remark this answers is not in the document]
ThagonDuarte: I was actually thinking of just changing them to an `AdditionalOutput`. Or do you think they don't provide useful information for debugging? @oleflb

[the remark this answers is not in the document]
oleflb: I don't think the output is required for now (at least I don't know what to debug with it from the top of my head)

crates/control/src/ball_filter.rs
oleflb: removed it from all functions in the node. I don't want to remove it from `remove_hypotheses` since I think there it makes sense to have that return value

ThagonDuarte:
Another remark, how exactly would a `BallHypothesis` with `BallMode` `Moving` go back into the mode `Resting`. I can't find that in the code. Additionally, when testing, the current implementation on main produces `Resting` `BallHypothesis`, while this never produces any `Resting` `BallHypothesis`. Am I missing something here?

ThagonDuarte (approved):
LGTM!

Cargo.toml
rmburg, line 1: was ist denn mit Cargo.toml los?
das sieht gar nicht gut aus :grimacing:

PR #1264, opened 2024-07-17
Obstructed kick decisions

### LINKED ISSUE #1050: No kick targets when obstacle is near goal
In this scenario there are no kick targets even though there is plenty of space to kick into the goal.
@schmidma suggested this may be because one of the kick targets is blocked and the other one is split into two because of obstacle intersections.
But the positions the ball would end up at with the split kick targets aren't inside the field anymore, leading to them being dropped too.
We should always have at least one kick target if there is a ball.
The dribble action executes `Stand` here.

Tthis may be related to #1012, @Vivituhh

image

### LINKED ISSUE #1184: Obstacles in Kick Decisions
Kick decisions use obstacles in their planning and deciding. The way obstacles are used is currently very conservative and also leads to instability as obstacles and the robots position are depending on each other and jockel around.

This issue represents the task of rethinking the way obstacles are used in general as well as just tuning some little parameters to improve the issue of obstacles leading to jockeling.

Some ideas might be:
- disabling obstacles all-together
- putting some hysteresis on the decision
- changing the way obstacles alter the kick targets
- reduce the obstacle size significantly, combined with the radius of the ball the obstructed area is gigantic.

schmidma:
## Why? What?

Removes the old obstacle avoidance, only scores kick decisions less when their target is obstructed by an obstacle.

Should also fix #1050

Closes #1184

## ToDo / Known Issues

none

none

## How to Test

Testgame + Replay, investigate targets


crates/control/src/kick_selector.rs
```diff
@@ -132,53 +146,209 @@ impl KickSelector {
...
+            game_phase: GamePhase::PenaltyShootout { .. },
+            kicking_team: Team::Hulks,
+            ..
+        })
+        | Some(FilteredGameControllerState {
+            sub_state: Some(SubState::PenaltyKick),
+            kicking_team: Team::Hulks,
+            ..
+        }) => PlayingSituation::PenaltyShot,
+        _ if is_ball_in_opponent_corner => PlayingSituation::CornerKick,
```
schluis, line 194: I would limit this to `FilteredGameState::Playing` to reduce the probability of errors

crates/control/src/kick_selector.rs
```diff
@@ -132,53 +146,209 @@ impl KickSelector {
...
+    vec![target]
+}
+
+fn generate_goal_line_kick_targets(context: &CycleContext) -> Vec<Point2<Ground>> {
+    let field_to_ground = context.ground_to_field.inverse();
+    let field_dimensions = &context.field_dimensions;
+
+    let left_goal_half = field_to_ground
+        * point![
+            field_dimensions.length / 2.0 + 0.1,
```
schluis, line 228: Can we make the additional offset dependable on distance to the goal? 
- Distance should be clamped to zero if we're inside the goal
- This could also allow us to still score goals if we're really delocalized in x

crates/control/src/kick_selector.rs
schmidma: There is no way these decisions can be used outside `Playing`. There is no ball and there is no striker...

crates/control/src/kick_selector.rs
schmidma: Maybe next iteration. This is what @PatrickGoettsch decided to do for now. I did not change this in comparison to `main`

knoellle (changes requested):

crates/control/src/kick_selector.rs
```diff
@@ -132,53 +146,209 @@ impl KickSelector {
...
+    target: Point2<Ground>,
+    parameters: &DecisionParameters,
+) -> bool {
+    let ball_to_target = LineSegment::new(ball_position, target);
+    let circles = obstacles
+        .iter()
+        .map(|obstacle| Circle::new(obstacle.position, obstacle.radius_at_foot_height))
+        .filter(|circle| circle.intersects_line_segment(&ball_to_target))
+        .collect::<Vec<_>>();
+    let closest_obstructing = circles.into_iter().max_by(|left, right| {
```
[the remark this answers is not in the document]
schmidma: Done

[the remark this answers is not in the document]
schmidma: Done

knoellle (approved):
LGTM

crates/control/src/kick_selector.rs
schluis: Ah sorry, I think I meant `GamePhase`

PR #1269, opened 2024-07-17
no more Pinguin

### LINKED ISSUE #1211: 4 hat Pinguin
4 klemmt Arme an die Seite, URG fragen ob man das kalibrieren kann.

JackyBloxx:
## Why? What?
no more pinguin for 32 and fixed addition of the offset parameters

Fixes #1211 

## How to Test

change offset parameters and look if they function right and look if 32 has no more pinguin


crates/control/src/motion/command_sender.rs
```diff
@@ -17,10 +17,10 @@ pub struct CreationContext {}
 pub struct CycleContext {
     optimized_motor_commands: Input<MotorCommands<Joints<f32>>, "optimized_motor_commands">,
     leds: Input<Leds, "leds">,
-
+    joint_calibration_offsets: Parameter<Joints<f32>, "joint_calibration_offsets">,
     motion_safe_exits: CyclerState<MotionSafeExits, "motion_safe_exits">,
-    last_actuated_motor_commands:
-        CyclerState<MotorCommands<Joints<f32>>, "last_actuated_motor_commands">,
+    last_actuated_motor_commands_before_offset:
```
schmidma, line 22: As you are now definitely the last who adds the offsets, there is no way for the user to access any joint angles after the offsets, so the renaming should stay the same as in main

```suggestion
    last_actuated_motor_commands:
```

schmidma (changes requested):

crates/control/src/motion/command_sender.rs
JackyBloxx: true
done

etc/parameters/body.P0000073A10S98U00008.json
```diff
@@ -1 +1,40 @@
...
+      "hip_roll": 0.0,
+      "hip_yaw_pitch": 0.0,
+      "knee_pitch": 0.0
+    },
+    "right_arm": {
+      "elbow_roll": 0.0,
+      "elbow_yaw": 0.0,
+      "hand": 0.0,
+      "shoulder_pitch": 0.0,
+      "shoulder_roll": -0.18,
```
schmidma, line 28: You can just put this very line instead of the entire struct

schmidma (approved):

PR #1283, opened 2024-07-17
More Score per Good Match

schmidma:
## Why? What?

Increases the score for good matches in localization. That leads to faster decision in walking in from penalized.

## ToDo / Known Issues

none

none

## How to Test

Let it walk in from various good or bad walk-in poses and check that it decides for the correct pose (tuned for Field E)


knoellle (approved):
Sah gut aus

PR #1295, opened 2024-07-17
Introduce circle radius offset and increase it.

PatrickGoettsch:
## Why? What?

Makes an offset scaling factor to an offset parameter. Parameter is increased from 14.5cm to 27.5cm

Fixes: Postgame issue #1285

### DIFF
--- a/crates/control/src/rule_obstacle_composer.rs
+++ b/crates/control/src/rule_obstacle_composer.rs
@@ -23,8 +23,7 @@ pub struct CycleContext {
         RequiredInput<Option<FilteredGameControllerState>, "filtered_game_controller_state?">,
     ball_state: Input<Option<BallState>, "ball_state?">,
 
-    center_circle_obstacle_increase:
-        Parameter<f32, "rule_obstacles.center_circle_obstacle_increase">,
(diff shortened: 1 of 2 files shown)


crates/control/src/rule_obstacle_composer.rs
```diff
@@ -23,8 +23,7 @@ pub struct CycleContext {
         RequiredInput<Option<FilteredGameControllerState>, "filtered_game_controller_state?">,
     ball_state: Input<Option<BallState>, "ball_state?">,
 
-    center_circle_obstacle_increase:
-        Parameter<f32, "rule_obstacles.center_circle_obstacle_increase">,
+    center_circle_obstacle_offset: Parameter<f32, "rule_obstacles.center_circle_obstacle_offset">,
```
schmidma, line 26: `offset` sounds like moving the obstacle. You want to add some additional delta to the radius. The `..._increase` was kinda fine to me. Maybe keep it? Or find an even better name

schmidma (changes requested):
one small naming remark

crates/control/src/rule_obstacle_composer.rs
PatrickGoettsch: Renamed

schmidma (approved):

PR #1300, opened 2024-07-17
Add `defender_distance_to_be_aligned`

### LINKED ISSUE #1277: Defender not facing ball when walking back
When Defender Naos walk back from an offensive position of a previous role, they always stop facing the ball. For short distances they could just walk backwards and thereby be able to see the ball and do intercepting steps. 

This could be implemented using a separate larger hybrid alignment distance only for the defenders.

ThagonDuarte:
## Why? What?

Adds a new parameter to differentiate the distance to be aligned by hybrid alignment between defender actions and all other actions. 
The motivation behind this is to make defenders turn around earlier when walking back to their position to face the ball and perhaps do intercepting steps.   

Fixes #1277 

## How to Test

Use behavior simulator to check the orientation of defenders when walking back.

### DIFF
--- a/crates/control/src/behavior/defend.rs
+++ b/crates/control/src/behavior/defend.rs
@@ -46,12 +46,14 @@ impl<'cycle> Defend<'cycle> {
         pose: Pose2<Ground>,
         path_obstacles_output: &mut AdditionalOutput<Vec<PathObstacle>>,
         walk_speed: WalkSpeed,
+        distance_to_be_aligned: f32,
     ) -> Option<MotionCommand> {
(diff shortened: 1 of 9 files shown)


schmidma:
Does this lead to the defenders walking backwards earlier?

ThagonDuarte:
This now causes the defenders to be aligned earlier.

ThagonDuarte:
However, I am not entirely sure, if all the Defend... Actions should receive the new distance to be aligned. I suggest brainstorming that in person.

schmidma (approved):

PR #1304, opened 2024-07-17
Actually score goals

schmidma:
## Why? What?

This change allows instant kick decisions if they are scoring goals. Previously they had to also be "strategic" that is "closer to the opponent goal"

Based on #1264 

## ToDo / Known Issues

none

none

## How to Test

when close to the goal, there should be red (instant) kick decisions in twix.

use replay of first half devils game on 34.


schluis (dismissed):
nice

schluis (approved):
I think it's good 👍🏻

PR #1308, opened 2024-07-17
Stop horizontal segments at uppermost limb segment

Narcha:
## Why? What?

what the title says (almost). This is needed for horizontal segments to be used in line detection, as otherwise we would detect our own feet/knees as lines.

## ToDo / Known Issues

This implementation is not entirely correct. Ideally, we should clip the line segments to the image rectangle, and then get the minimum over the clipped endpoints. This PR achieves 80% of that, using 20% of the effort. I'm working on a follow-up with a better implementation.

see above

## How to Test

twix

### DIFF
--- a/crates/vision/src/image_segmenter.rs
+++ b/crates/vision/src/image_segmenter.rs
@@ -5,6 +5,7 @@ use std::{
 
 use color_eyre::Result;
 use itertools::iproduct;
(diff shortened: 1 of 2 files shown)


schmidma (changes requested):
one small remark

PR #1310, opened 2024-07-17
Detect edges on saturation

h3ndrk:
## Why? What?

We wanted to try more color channels like HSV for image segmenter edge detection.

TL;DR: It is not promising, see images below. But let's add it anyway, maybe we will use it at some event.

# Y (as before)

# Saturation (new)

# Saturation YHS2 (new)

## ToDo / Known Issues

None

None

## How to Test

Deploy, look through Twix (select different image sources in image segmenter panel and set parameter `edge_detection_source` and tune the `edge_detection_threshold`)

### DIFF
--- a/crates/types/src/color.rs
+++ b/crates/types/src/color.rs
@@ -252,6 +252,74 @@ impl From<Rgb> for Hsv {
     }
 }
 
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Deserialize,
+    Eq,
+    PartialEq,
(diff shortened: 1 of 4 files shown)


schmidma:
To use this in line detection we might have to check that line detection is currently searching for rising and then falling edges, this may be inverted in the alternative segmentation approaches

oleflb:
Someone should fix the tests

oleflb:
Decided to not need this since results dont look promising

PR #1317, opened 2024-07-18
Empty Field Decimator

knoellle:
## Why? What?

Avoid invalid goals by preventing kicking at the goal if only one robot touched the ball since the last "set play".

Fixes #

## ToDo / Known Issues

## How to Test


oleflb:
We probably don't want this anymore, right? Instead we want to have a proper implementation in the future I guess. If you feel the same, please close

PR #1309, opened 2024-07-17
add whistle fake in simulator for opponent kickoff

MaikRe:
## Why? What?

We have no current method of simulating whistles in the behavior simulator.
The only transition into playing is through the `game_controller_state` `Playing`, which immediately sets `ball_is_free` to `true`. Testing midfielder or other positioning in the `ball_is_free = false` phase was impossible until this PR.

Fixes #

## How to Test

Run  behavior simulator with `golden_goal_opponent_kickoff` and observe that between frames 1700 and 1713 the ball is not free and thus the correct behavior from the midfielders can be observed.
If you then revert #1307 you can see the midfielders plan paths into the opponent half.

Issue #1205, opened 2024-07-15
Role assignment in behaviour simulator broken for penalty kicks


ThagonDuarte:
In the behaviour simulator when walking to their positions for a penalty kick, the Striker, MidfielderRight, MidfielderLeft and StrikerSupport get assigned the Searcher role. This results in three of them walking to the exact same position. 
This may be due to the `spl_striker_message` not being sent correctly.

PR #1321, opened 2024-07-18
Penalty kick behavioral fixes

MaikRe:
## Why? What?

- Ignore striker message timeout to prevent roles from becoming looser and searcher in penalty kick as these seem to violate obstacles
- Keep striker in specific position for penalty kick because he would ignore the penalty box obstacle due to #1051 
- Clamp midfielders to opponent half to prevent them from walking back to own half during penalty kick

Fixes #

## How to Test

Try it in game or merge #1309 to test in behavior simulator

### DIFF
--- a/crates/control/src/behavior/support.rs
+++ b/crates/control/src/behavior/support.rs
@@ -86,13 +86,22 @@ fn support_pose(
         .filtered_game_controller_state
         .as_ref()
         .map(|filtered_game_controller_state| filtered_game_controller_state.game_state);
-    let mut clamped_x = match filtered_game_state {
-        Some(FilteredGameState::Ready { .. })
-        | Some(FilteredGameState::Playing {
-            ball_is_free: false,
-            kick_off: true,
-            ..
-        }) => supporting_position
+    let sub_state = world_state
+        .filtered_game_controller_state
+        .map(|filtered_game_controller_state| filtered_game_controller_state.sub_state);
+    let mut clamped_x = match (filtered_game_state, sub_state) {
+        (Some(FilteredGameState::Ready { .. }), Some(Some(SubState::PenaltyKick))) => {
+            supporting_position.x().max(field_dimensions.length / 4.0)
+        }
+        (Some(FilteredGameState::Ready { .. }), _)
+        | (
+            Some(FilteredGameState::Playing {
+                ball_is_free: false,
+                kick_off: true,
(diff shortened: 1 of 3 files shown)


crates/control/src/behavior/support.rs
```diff
@@ -83,13 +83,22 @@ fn support_pose(
...
-        | Some(FilteredGameState::Playing {
-            ball_is_free: false,
-            kick_off: true,
-            ..
-        }) => supporting_position
+    let sub_state = world_state
+        .filtered_game_controller_state
+        .map(|filtered_game_controller_state| filtered_game_controller_state.sub_state);
+    let mut clamped_x = match (filtered_game_state, sub_state) {
+        (Some(FilteredGameState::Ready { .. }), Some(Some(SubState::PenaltyKick))) => {
```
oleflb, line 90: What is the difference between filtered_game_state and world_state.filtered_game_controller_state?

oleflb (changes requested):

crates/control/src/role_assignment.rs
```diff
@@ -211,7 +211,14 @@ impl RoleAssignment {
...
 
-        if spl_striker_message_timeout {
+        let is_in_penalty_kick = matches!(
+            context.filtered_game_controller_state,
+            Some(FilteredGameControllerState {
+                sub_state: Some(SubState::PenaltyKick),
+                ..
+            })
+        );
+        if spl_striker_message_timeout && !is_in_penalty_kick {
```
oleflb, line 221: isnt this a band-aid fix and we want proper path-planning instead?

MaikRe: A little bit? As far as I can tell the problem is that searchers ignore obstacles in path planning.
On the other hand I have never seen a robot become searcher in a penalty kick, which means that striker message timeout could be a behavior simulator specific bug. See #1205

crates/control/src/behavior/support.rs
MaikRe: `filtered_game_state` is the `world_state.filtered_game_controller_state.game_state` which is first properly handled in line 83

oleflb (changes requested):
It seems like striker panics when executing a penalty kick, is this related? Otherwise please rebase

oleflb (approved):

PR #1333, opened 2024-07-18
Reset detection time in referee_pose_detection_filter

### LINKED ISSUE #1278: Sacrificial lamb issue in second half

JackyBloxx:
## Why? What?

fixes that we skip standbyphase in 2. half  if we didnt redeploy in halftime so we dont get penaltys for motion in standby

Fixes #1278

## How to Test

see if it happens

### DIFF
--- a/crates/control/src/referee_pose_detection_filter.rs
+++ b/crates/control/src/referee_pose_detection_filter.rs
@@ -10,9 +10,10 @@ use serde::{Deserialize, Serialize};
 use context_attribute::context;
 use framework::{AdditionalOutput, MainOutput, PerceptionInput};
 use hardware::NetworkInterface;
-use spl_network_messages::{HulkMessage, PlayerNumber, VisualRefereeMessage};
+use spl_network_messages::{GameState, HulkMessage, PlayerNumber, VisualRefereeMessage};
 use types::{
     cycle_time::CycleTime,
+    game_controller_state::GameControllerState,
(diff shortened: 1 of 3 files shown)


crates/control/src/referee_pose_detection_filter.rs
```diff
@@ -85,6 +86,14 @@ impl RefereePoseDetectionFilter {
         let (is_referee_ready_pose_detected, did_detect_any_referee_this_cycle) =
             self.update(&context)?;
 
+        let is_standby = matches!(
+            context.game_controller_state.unwrap().game_state,
+            GameState::Standby
+        );
```
julianschuler, line 92: This will panic when we do not have `game_controller_state`. Use this instead:
```suggestion
        let is_standby = matches!(
            context.game_controller_state,
            Some(GameControllerState {
                game_state: GameState::Standby,
                ..
            })
        );
```

julianschuler (changes requested):

julianschuler (approved):

MaikRe:
Whats the state of this? Is a 'merge when ready' all that is missing? @JackyBloxx

JackyBloxx:
> Whats the state of this? Is a 'merge when ready' all that is missing? @JackyBloxx

the problem is it isnt tested so i doesnt realy now if it realy fixed the probelm.

pejotejo:
@oleflb du willst dinge testen

paudar:
Outdated

PR #1334, opened 2024-07-18
Line semantics

julianschuler:
## Why? What?

This redefines a line as a point and a direction vector to further emphasize the difference to a line segment.

It additionally adds test cases for all functions defined for it.

## ToDo / Known Issues

None.

None.

## How to Test

Build the code, run the tests.

### DIFF
--- a/crates/geometry/src/line.rs
+++ b/crates/geometry/src/line.rs
@@ -3,83 +3,70 @@ use std::ops::Mul;
 use approx::{AbsDiffEq, RelativeEq};
 use serde::{Deserialize, Serialize};
 
-use linear_algebra::{distance_squared, point, Point, Point2, Transform};
+use linear_algebra::{distance_squared, Point, Point2, Transform, Vector};
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
 
 use crate::{direction::Direction, line_segment::LineSegment, Distance};
 
 #[derive(
-    Copy,
-    Clone,
-    Debug,
-    Default,
-    Deserialize,
-    Serialize,
-    PathSerialize,
-    PathIntrospect,
-    PathDeserialize,
+    Copy, Clone, Debug, Deserialize, Serialize, PathSerialize, PathIntrospect, PathDeserialize,
 )]
-pub struct Line<Frame, const DIMENSION: usize>(
-    pub Point<Frame, DIMENSION>,
-    pub Point<Frame, DIMENSION>,
-);
+pub struct Line<Frame, const DIMENSION: usize> {
+    pub point: Point<Frame, DIMENSION>,
+    pub direction: Vector<Frame, DIMENSION>,
+}
 
 pub type Line2<Frame> = Line<Frame, 2>;
 pub type Line3<Frame> = Line<Frame, 3>;
 
 impl<Frame> Line2<Frame> {
     pub fn slope(&self) -> f32 {
-        let difference = self.0 - self.1;
-        difference.y() / difference.x()
+        self.direction.y() / self.direction.x()
     }
 
     pub fn y_axis_intercept(&self) -> f32 {
-        self.0.y() - (self.0.x() * self.slope())
-    }
-
-    pub fn project_onto_segment_in_x_axis(&self, point: Point2<Frame>) -> f32 {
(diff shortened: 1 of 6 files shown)


crates/geometry/src/line.rs
```diff
@@ -3,83 +3,70 @@ use std::ops::Mul;
...
-            ((((x1 * y2) - (y1 * x2)) * (y3 - y4)) - ((y1 - y2) * ((x3 * y4) - (y3 * x4))))
-                / (((x1 - x2) * (y3 - y4)) - ((y1 - y2) * (x3 - x4)))
-        )
+        let point_difference = self.point - other.point;
+
+        let direction_factor = (point_difference.x() * self.direction.y()
+            - point_difference.y() * self.direction.x())
+            / (self.direction.y() * other.direction.x() - self.direction.x() * other.direction.y());
+
+        other.point + other.direction * direction_factor
```
schmidma, line 49: How can I check that this implementation is correct?

crates/geometry/src/line.rs
```diff
@@ -90,21 +77,35 @@ impl<Frame, const DIMENSION: usize> Distance<Point<Frame, DIMENSION>> for Line<F
...
 }
 
+impl<Frame, const DIMENSION: usize> Default for Line<Frame, DIMENSION> {
+    fn default() -> Self {
+        Self {
+            point: Default::default(),
+            direction: Vector::zeros(),
+        }
+    }
+}
```
schmidma, line 87: You should be able to derive that?!

crates/geometry/src/line.rs
```diff
@@ -90,21 +77,35 @@ impl<Frame, const DIMENSION: usize> Distance<Point<Frame, DIMENSION>> for Line<F
...
+            point: self * line.point,
+            direction: self * line.direction,
+        }
     }
 }
 
 impl<Frame, const DIMENSION: usize> PartialEq for Line<Frame, DIMENSION> {
     fn eq(&self, other: &Self) -> bool {
-        (self.0 == other.0 && self.1 == other.1) || (self.0 == other.1 && self.1 == other.0)
+        self.point == other.point && self.direction == other.direction
```
schmidma, line 108: The previous implementation considered a line with negative `direction` to be the same. Is it correct to drop this assumption? Also, it is possible to define the same line with multiple `point`s and `direction`s like scaling them, or moving the point along the direction.

schmidma (changes requested):

crates/geometry/src/line.rs
julianschuler: IMO two lines are the same if their normalized direction is the same (also with the same sign, since the behavior of `signed_distance_to` is inconsistent otherwise) and if the point of one line lies also on the other one. I am not sure whether this may result in problems with floating point inaccuracies, i.e. whether `PartialEq` should be implemented at all for `Line`. What is your opinion on this?

crates/geometry/src/line.rs
julianschuler: I derived it yesterday (on paper), I have added test cases covering this. Is this sufficient for you?

crates/geometry/src/line.rs
julianschuler: I don't know why, but Vector<Framed, DIMENSION> doesn't seem to implement Default.

crates/geometry/src/line.rs
schmidma: Is it used somewhere? And when yes, for what?

crates/geometry/src/line.rs
julianschuler: I now added a comment explaining why default can not be derived for `[f32; DIMENSION]`.

crates/geometry/src/line.rs
julianschuler: No, it is not used anywhere. Though `RelativeEq` is used in some test cases, so maybe we should drop the `PartialEq` implementation and do `RelativeEq` according to the prior definition.

julianschuler: I now updated the implementations to consider lines as (relative) equal if their normalized directions are the same and the point of one line also lies on the other one.

schmidma (approved):

PR #1337, opened 2024-07-18
Remote control fixes

Narcha:
## Why? What?

- fix crash that occurred when receiving a shared autonomy challenge game controller message
- add head joint control to remote panel

## ToDo / Known Issues

it's perfect

## How to Test

twix, remote control panel

### DIFF
--- a/crates/control/src/motion/head_motion.rs
+++ b/crates/control/src/motion/head_motion.rs
@@ -2,6 +2,7 @@ use std::f32::consts::PI;
 
 use color_eyre::Result;
 use context_attribute::context;
+use filtering::low_pass_filter::LowPassFilter;
 use framework::MainOutput;
 use serde::{Deserialize, Serialize};
 use types::{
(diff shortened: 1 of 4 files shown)


crates/control/src/motion/head_motion.rs
```diff
@@ -48,10 +51,22 @@ impl HeadMotion {
     pub fn new(_context: CreationContext) -> Result<Self> {
         Ok(Self {
             last_positions: Default::default(),
+            lowpass_filter: LowPassFilter::with_smoothing_factor(Default::default(), 0.075),
```
schluis, line 54: Please convert this to a parameter 🙃

schluis (changes requested):
I like

crates/control/src/motion/head_motion.rs
schluis: Ah, nevermind

schluis (approved):

PR #1339, opened 2024-07-19
Re(altime)-Play

knoellle:
## Why? What?

- Press `Space` in the replayer to toggle real time playback.
- Press `<`, `>` to decrease/increase playback speed in 0.25x increments

Fixes #scrubbing pain

- "Dense replay", replaying intermediate frames and preventing duplicate replay if it would just replay the same frame again for that cycler.
- Playback rate clamping
- Reverse playback

## How to Test

Replayer should continue to work as before but with the aforementioned features.-

### DIFF
--- a/crates/hulk_replayer/src/window.rs
+++ b/crates/hulk_replayer/src/window.rs
@@ -73,7 +73,10 @@ impl App for Window {
             }
         });
         CentralPanel::default().show(context, |ui| {
-            ui.label(format!("{}", self.time_sender.borrow().playback_rate));
+            ui.label(format!(
+                "Speed: {}",
+                self.time_sender.borrow().playback_rate
+            ));
             ui.horizontal_top(|ui| {
                 ui.add(Labels::new(&self.indices));
                 let absolute_position = AbsoluteTime::new(self.time_sender.borrow().time);
--- a/crates/hulk_replayer/src/window.rs
+++ b/crates/hulk_replayer/src/window.rs
@@ -1,7 +1,7 @@
 use std::{collections::BTreeMap, time::SystemTime};
 
 use eframe::{
-    egui::{CentralPanel, Context},
+    egui::{CentralPanel, Context, Event, Key, Modifiers},
     App, Frame,
 };
 use tokio::sync::watch;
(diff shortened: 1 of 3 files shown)


crates/hulk_replayer/src/worker_thread.rs
```diff
@@ -1,36 +1,73 @@
...
         runtime.block_on(async move {
+            let mut receiver = sender.subscribe();
             let mut parameters_receiver = replayer.get_parameters_receiver();
+            let mut last_autoplay_time = Instant::now();
             loop {
                 select! {
                     _ = parameters_receiver.wait_for_change() => {}
                     _ = sleep(Duration::from_secs(1)) => {}
-                    result = time.changed() => {
+                    _ = sleep(Duration::from_millis(12)), if receiver.borrow().playing => {
```
h3ndrk, line 43: This ticks every 12 milliseconds even if you don't play which may happen most of the time. This should probably be also event-based. It smells as you can see with your comment and `continue` :wink: Maybe use a proper timer object instead of managing time manually? Your current solution also violates single responsibility principle and single layer of abstraction. You could/should hide the time handling better to keep this function easier to read.

crates/hulk_replayer/src/worker_thread.rs
```diff
@@ -1,36 +1,73 @@
...
+                        let elapsed = last_autoplay_time.elapsed();
+                        last_autoplay_time = Instant::now();
+                        sender.send_modify(|state| {
+                            state.time += elapsed.mul_f32(state.playback_rate);
+                        });
+                        // receiver is updated anyway, prevent duplicate replaying
+                        continue;
```
h3ndrk, line 50: Why do you need this match arm here? Isn't this duplicating the code outside of the `select!`? :thinking:

crates/hulk_replayer/src/window.rs
```diff
@@ -12,56 +12,83 @@ use crate::{
...
+                .iter()
+                // egui is scheiße
+                .any(|e| *e == Event::Text(">".to_string()))
+            {
+                self.time_sender
+                    .send_modify(|state| state.playback_rate += 0.25);
+            }
+        });
         CentralPanel::default().show(context, |ui| {
+            ui.label(format!("{}", self.time_sender.borrow().playback_rate));
```
h3ndrk, line 76: This label will claim a lot of horizontal space for a information that is probably not so important to show it all the time. Could you integrate it more intelligently in the UI to save space or maybe show it only shortly? Or in the time? Or with some other visual elements graphically?

crates/hulk_replayer/src/worker_thread.rs
```diff
@@ -1,36 +1,73 @@
...
+    pub time: SystemTime,
+    pub playing: bool,
+    pub playback_rate: f32,
+}
+
+impl Default for PlayerState {
+    fn default() -> Self {
+        Self {
+            time: SystemTime::UNIX_EPOCH,
+            playing: Default::default(),
```
h3ndrk, line 21: Write explicit `false` to make more visible what the default is?

crates/hulk_replayer/src/worker_thread.rs
knoellle: This only ticks if playing is enabled.
The if on the right hand side is a precondition of the select macro. The future is evaluated but never polled if the precondition is false.

crates/hulk_replayer/src/worker_thread.rs
knoellle: No, this match arm updates the time and the part after the select evaluates the replay.

crates/hulk_replayer/src/window.rs
knoellle: - I don't think it's a lot of vertical space
- In this UI, horizontal space is a lot more valuable than vertical
- I'd like to add more elements to that row like a button to toggle playing (as an alternative to hotkey), mode selection, etc, which won't make it any higher either.

crates/hulk_replayer/src/worker_thread.rs
knoellle: Done

h3ndrk (approved):

PR #1349, opened 2024-07-19
Less activations to activate foot bumper obstacles

PatrickGoettsch:
## Why? What?

Have less activation / touches per Foot bumper needed to spawn an obstacle.


schluis (approved):
Let's try it

PR #1353, opened 2024-07-19
Haaaaaaaalt stop

rmburg:
## Why? What?

gif

Based on #1337 

Injects `MotionCommand::Stand` when not moving with the remote control panel.

## ToDo / Known Issues

It's perfect


Narcha:
Note to self: allow turning in place

oleflb:
No longer needed on K1.

PR #1357, opened 2024-07-19
Wide stance improvements

julianschuler:
## Why? What?

This PR improves the wide stance by adding an additional distance check, tuning parameters and adjusting active vision to let the keeper look more at the ball.

## ToDo / Known Issues

None.

- Make active vision dependent on position uncertainty (i.e. if the keeper knows exactly where it is, look at the corner less often).

## How to Test

Deploy a robot as keeper (player number one and two other robots in the game controller) and see how the keeper behaves.

### DIFF
--- a/crates/control/src/behavior/head.rs
+++ b/crates/control/src/behavior/head.rs
@@ -1,5 +1,6 @@
 use types::{
     motion_command::{HeadMotion, ImageRegion},
+    roles::Role,
     world_state::WorldState,
 };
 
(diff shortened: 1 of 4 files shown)


crates/control/src/behavior/head.rs
```diff
@@ -14,10 +15,26 @@ impl<'cycle> LookAction<'cycle> {
...
-            target: self.world_state.position_of_interest,
-            image_region_target: ImageRegion::Center,
-            camera: None,
+        if self.world_state.robot.role == Role::Keeper {
+            if let Some(ball_position) = self.world_state.ball {
+                HeadMotion::LookAt {
+                    target: ball_position.ball_in_ground,
+                    image_region_target: ImageRegion::Center,
+                    camera: None,
+                }
```
schmidma, line 24: We have to check whether this has negative consequences for the keeper localization... but okay for now

(resolve when read)

schmidma (approved):

PR #1367, opened 2024-07-20
Use hybrid AlignWithPath in Dribble

schmidma:
## Why? What?

Instead of facing the ball, the striker uses proper hybrid alignment in dribble. This somewhat mitigates the problem of drifting towards the ball... but this also changes behavior on how to approach the ball, we should test whether this looks good.

## ToDo / Known Issues

none

none

## How to Test

Let the robot walk around the ball... Look at its alignment and path/step planning

### DIFF
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1139,8 +1139,8 @@
       }
     },
     "dribbling": {
-      "hybrid_align_distance": 0.5,
-      "distance_to_be_aligned": 0.1,
+      "hybrid_align_distance": 1.0,
+      "distance_to_be_aligned": 0.15,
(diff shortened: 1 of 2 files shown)


etc/parameters/default.json
```diff
@@ -1139,8 +1139,8 @@
       }
     },
     "dribbling": {
-      "hybrid_align_distance": 0.5,
-      "distance_to_be_aligned": 0.1,
+      "hybrid_align_distance": 1.0,
+      "distance_to_be_aligned": 0.15,
```
schluis, line 1143:
```suggestion
      "hybrid_align_distance": 1.0,
      "distance_to_be_aligned": 0.1,
```

schmidma:
I don't think this is a good idea

Issue #1325, opened 2024-07-18
Time to reach always has a value


rmburg:
It seems like fallen robots often calculate a short TTRKP and thus claim striker. Investigate.

Examples:
- 18.07. game vs. NomadZ, Robot 36:5, during walk-in at the start of the first half
- 17.07. game vs. Nao Devils, Robot 34:4, at 9:25 remaining in the first half

MaikRe:
Because of the inputs needed for `time_to_reach_kick_position` it runs after all of behavior and hands it's values to the cyclerstate in order for them to be used in the next cycle. As tracked in #402, this cycler state does not support options.
Therefor the time to reach is always Duration::MAX rather than None, and because Lua in the behavior simulator cannot handle duration max it is reduced to 30 minutes.
TLDR We cant have a time to reach of None meaning if everyone is looser everyone claims striker for at least one cycle

knoellle:
`CyclerState` *can* contain an option, but it's impossible to set it to `Some(..)`, this is a framework/code-gen bug

PR #1371, opened 2024-07-20
quick and dirty check for time to reach

MaikRe:
## Why? What?

See #1325 for description of issue.
This PR ensures we don't claim striker when our time to reach is too high

Fixes #

## ToDo / Known Issues

## How to Test

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -625,7 +625,9 @@ fn decide_if_claiming_striker_or_other_role(
     filtered_game_controller_state: Option<&FilteredGameControllerState>,
     optional_roles: &[Role],
 ) -> (Role, bool, Option<BallPosition<Field>>) {
-    if time_to_reach_kick_position < spl_message.time_to_reach_kick_position {
+    if time_to_reach_kick_position < spl_message.time_to_reach_kick_position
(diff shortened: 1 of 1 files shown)


crates/control/src/role_assignment.rs
```diff
@@ -625,7 +625,9 @@ fn decide_if_claiming_striker_or_other_role(
     filtered_game_controller_state: Option<&FilteredGameControllerState>,
     optional_roles: &[Role],
 ) -> (Role, bool, Option<BallPosition<Field>>) {
-    if time_to_reach_kick_position < spl_message.time_to_reach_kick_position {
+    if time_to_reach_kick_position < spl_message.time_to_reach_kick_position
+        && time_to_reach_kick_position.unwrap_or(Duration::MAX) < Duration::from_secs(1200)
```
schmidma, line 629: I think this is not correct. You are comparing options in 682. This is really weird to reason about. This should probably be some `.is_some() && `

schmidma (changes requested):

crates/control/src/role_assignment.rs
MaikRe: Again what learned. Fixed

schmidma (approved):

PR #1385, opened 2024-07-20
Need always only two contacts for Footbumper

PatrickGoettsch:
## Why? What?

What title says

### DIFF
--- a/crates/control/src/foot_bumper_filter.rs
+++ b/crates/control/src/foot_bumper_filter.rs
@@ -145,18 +145,17 @@ impl FootBumperFilter {
             (self.right_count >= *context.activations_needed) && self.right_in_use;
         let obstacle_detected_on_middle = ((self.right_count + self.left_count)
             >= *context.activations_needed)
-            && (self.left_in_use || self.right_in_use);
+            && (self.left_in_use && self.right_in_use);
 
-        let mut obstacle_angle = 0.0;
-
-        if obstacle_detected_on_left {
-            obstacle_angle = *context.sensor_angle;
+        let obstacle_angle = if obstacle_detected_on_left {
+            *context.sensor_angle;
         } else if obstacle_detected_on_right {
-            obstacle_angle = -*context.sensor_angle;
+            -*context.sensor_angle;
         } else if obstacle_detected_on_middle {
+            0.0
         } else {
             return Ok(Default::default());
-        }
+        };
 
         let obstacle_position = Rotation2::<Ground, Ground>::new(obstacle_angle)
             * point![*context.obstacle_distance, 0.0];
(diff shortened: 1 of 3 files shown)


crates/control/src/foot_bumper_filter.rs
```diff
@@ -133,24 +133,31 @@ impl FootBumperFilter {
...
-            (false, true) => -context.sensor_angle,
-            _ => return Ok(Default::default()),
-        };
+        let obstacle_detected_on_left =
+            (self.left_count >= *context.activations_needed) && self.left_in_use;
+        let obstacle_detected_on_right =
+            (self.right_count >= *context.activations_needed) && self.right_in_use;
+        let obstacle_detected_on_middle = ((self.right_count + self.left_count)
+            >= *context.activations_needed)
+            && (self.left_in_use || self.right_in_use);
```
schmidma, line 148: Shouldn't this be
```suggestion
            && (self.left_in_use && self.right_in_use);
```

???

crates/control/src/foot_bumper_filter.rs
```diff
@@ -133,24 +133,31 @@ impl FootBumperFilter {
...
+        let mut obstacle_angle = 0.0;
+
+        if obstacle_detected_on_left {
+            obstacle_angle = *context.sensor_angle;
+        } else if obstacle_detected_on_right {
+            obstacle_angle = -*context.sensor_angle;
+        } else if obstacle_detected_on_middle {
+        } else {
+            return Ok(Default::default());
+        }
```
schmidma, line 159: use something like to avoid `mut`:

```suggestion
        let obstacle_angle = if obstacle_detected_on_left {
            *context.sensor_angle;
        } else if obstacle_detected_on_right {
            -*context.sensor_angle;
        } else if obstacle_detected_on_middle {
            0.0
        } else {
            return Ok(Default::default());
        };
```

schmidma (changes requested):

crates/control/src/foot_bumper_filter.rs
PatrickGoettsch: Thanks

crates/control/src/foot_bumper_filter.rs
PatrickGoettsch: Done.

crates/control/src/foot_bumper_filter.rs
```diff
@@ -133,24 +133,30 @@ impl FootBumperFilter {
...
+            >= *context.activations_needed)
+            && (self.left_in_use && self.right_in_use);
+
+        let obstacle_angle = if obstacle_detected_on_left {
+            *context.sensor_angle;
+        } else if obstacle_detected_on_right {
+            -*context.sensor_angle;
+        } else if obstacle_detected_on_middle {
+            0.0
+        } else {
```
schluis, line 156: Please check `obstacle_detected_on_middle` first to prioritize it.

schluis (changes requested):

schluis (approved):

PR #1396, opened 2024-07-20
draw team ball in map panel

schmidma:
## Why? What?

What it says, makes replay debugging a lot easier.
Team Ball is read, own ball white

image

### DIFF
--- a/tools/twix/src/panels/map/layers/ball_position.rs
+++ b/tools/twix/src/panels/map/layers/ball_position.rs
@@ -5,7 +5,10 @@ use eframe::epaint::Color32;
 
 use coordinate_systems::{Field, Ground};
 use linear_algebra::Isometry2;
-use types::field_dimensions::FieldDimensions;
+use types::{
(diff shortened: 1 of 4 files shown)


tools/twix/src/panels/map/layers/ball_position.rs
```diff
@@ -26,9 +30,11 @@ impl Layer<Field> for BallPosition {
         );
         let ball_position = nao
             .subscribe_buffered_value("Control.main_outputs.ball_position", Duration::from_secs(2));
+        let world_state = nao.subscribe_value("Control.main_outputs.world_state");
```
knoellle, line 33: Why do you subscribe the whole worldstate here instead of using the teamball provided by role assigment?
The `ball_in_field` inside the world state is equal to the own ball if there is one and only if there isn't, it uses the actual team ball.

schmidma: Done

knoellle (approved):
LGTM

PR #1404, opened 2024-07-21
Vision tuner panel: add save button

Narcha:
## Why? What?

Does just what you think it would do.

## ToDo / Known Issues

It's perfect

More save buttons, location selector

## How to Test

Use the panel, watch git diff


julianschuler (approved):
LGTM!

PR #1423, opened 2024-07-31
Revert WiFi passwords for RC24

schmidma:
## Why? What?

What it says. We should revert back to the rule default, what is also locally deployed in our lab.


crates/constants/src/lib.rs
```diff
@@ -14,7 +14,7 @@ pub const HULA_DBUS_SERVICE: &str = "org.hulks.hula";
 pub const HULA_SOCKET_PATH: &str = "/tmp/hula";
 pub const OS_IS_NOT_LINUX: bool = !cfg!(target_os = "linux");
 pub const OS_RELEASE_PATH: &str = "/etc/os-release";
-pub const OS_VERSION: &str = "7.5.5";
+pub const OS_VERSION: &str = "7.5.6";
```
julianschuler, line 17: Please bump the pepsi version too.

julianschuler (changes requested):

crates/constants/src/lib.rs
schmidma: Done

julianschuler (approved):

PR #1426, opened 2024-08-07
Add missing dependencies for fresh install

tuxbotix:
## Why? What?

For some reason dependencies that were already listed in the docs (or I thought so) were removed at one point. But these are **required** for a fresh install (esp if you reinstall Fedora or you are a newbie).

~If someone else can test for Arch & Ubuntu, I can add cmake, etc to these two as well.~ - Referred `tools/ci/github-runners/Dockerfile`.

P.S. `systemd-devel` provides `udev`

Fixes #

FIxes rust-analyser and pepsi build errors caused by missing dependencies for Fedora.

## How to Test

1. All commands on each distro should run without an error
2. On a fresh install (i.e. a freshly pulled docker container), these instructions should get `./pepsi run` complete as the baseline.
3. The LSP (via VSCode, etc) - connected to the above mentioned docker container should index without any errors.
  - When something is missing, this will fail even if `pepsi run` works.

### DIFF
--- a/docs/setup/development_environment.md
+++ b/docs/setup/development_environment.md
@@ -32,7 +32,7 @@ Use your distribution's package manager to install the following dependencies:
 === "Fedora"
 
     ```sh
-    sudo dnf install git git-lfs clang python3 which zstd xz file rsync alsa-lib-devel
+    sudo dnf install git git-lfs clang cmake python3 which zstd xz file rsync alsa-lib-devel opusfile-devel hdf5-devel luajit-devel systemd-devel
     ```
 
 === "Ubuntu"
--- a/docs/setup/development_environment.md
+++ b/docs/setup/development_environment.md
@@ -16,29 +16,34 @@ Use your distribution's package manager to install the following dependencies:
 
 -   [Git](https://git-scm.com/), [Git LFS](https://git-lfs.com/)
 -   [clang](https://clang.llvm.org/)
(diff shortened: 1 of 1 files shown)


tuxbotix:
Aha that explains what happened 😅.

I wholeheartedly agree with you that these dependencies should lesser or at least should be explained why and what.

That said, I found out that these are the bare minimum to get the lsp running and building with pepsi. It is pointless if we can upload to the Nao but if the user cannot use the LSP in a meaningful way.

Things to do IMO:

- decide and clear out edge cases that require these (i.e. HDF5 are no longer actively used AFAIK)
- how to make rust analyser work with the minimum set?
- Or could we rethink of splitting as you said, but maybe for a different baseline?

schluis:
Okay, how do we continue here?
I suggest that we start by splitting the requirements as @tuxbotix suggested into required for standard development and required for other tools

tuxbotix:
I would also prefer to continue with this PR after testing the contents. IMO the baseline should be uploading and also running the LSP in a reasonable manner.

For simple deployment only, we could write a subset in a different PR.

schluis:
Okay, sounds good. Will you test it? In that case I would propose to move it back to "In Progress"

schluis (approved):
After discussion with @tuxbotix fine for me 👍🏻

PR #1427, opened 2024-08-07
Remove `Uncertain` kind from Team type

knoellle:
## Why? What?

~~`Team::Uncertain` is removed and wrapped with in `Option` where applicable (only in `FilteredGameState::Ready`)~~
**Update**: there is now only one kicking team in the `GameControllerState` and the `Playing` state contains a flag whether the kicking_team sent by the game controller is trustworthy.

This resolves the necessity to match the uncertain Team kind in places where it is always either Hulks or Opponent anyways.
This was especially annoying with the new behavior simulator but I wanted to make this a separate PR.

### Background

The FilteredGameControllerState contains a `kicking_team` field as well as two fields of type `FilteredGameState`, one for each team.
The `FilteredGameState::Ready` kind is the only place in our code where the `Team` could be `Uncertain` and if it wasn't uncertain, then the kicking team inside this kind was equal to the kicking team in the `FilteredGameState`.

Fixes #

## How to Test

Everything should work as before, there are no functional changes.

### DIFF
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -167,8 +167,7 @@ fn get_rule_hypotheses(
     filtered_game_controller_state: &FilteredGameControllerState,
     field_dimensions: FieldDimensions,
 ) -> Vec<Point2<Field>> {
-    let kicking_team_half =
-        conservative_kicking_team_half(filtered_game_controller_state.kicking_team);
+    let kicking_team_half = kicking_team_half(filtered_game_controller_state.kicking_team);
 
     match (primary_state, filtered_game_controller_state.sub_state) {
         (PrimaryState::Ready, Some(SubState::PenaltyKick)) => {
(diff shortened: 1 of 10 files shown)


crates/control/src/search_suggestor.rs
```diff
@@ -195,11 +194,9 @@ fn get_rule_hypotheses(
     }
 }
 
-fn conservative_kicking_team_half(kicking_team: Team) -> Half {
+fn kicking_team_half(kicking_team: Team) -> Half {
```
oleflb, line 197: What do you think of adding an into trait impl for this?

oleflb (changes requested):

crates/control/src/search_suggestor.rs
knoellle: I'm not so sure about this. The mapping between team hand half may be flipped depending on the logic one is trying to implement. I'd rather have well named functions for this than implement the into trait.

knoellle:
I basically changed everything again, see comment in PR description.
Please have another look.

oleflb (commented):
Looks good, but have not tested yet

crates/types/src/filtered_game_state.rs
```diff
@@ -19,7 +18,7 @@ pub enum FilteredGameState {
     #[default]
     Initial,
     Ready {
-        kicking_team: Team,
+        kicking_team_known: bool,
```
schluis, line 21: Why have this here at all? It's never read.

knoellle:
Rebased due to conflict with #1429.
@oleflb what is the state here?

oleflb (approved):
Abfahrt :steam_locomotive:

PR #1429, opened 2024-08-14
Clean up construct_*_filtered_game_state

knoellle:
## Why? What?

While working on the bevyhavior simulator, I noticed that this part of the game controller state filter can be simplified significantly.

Fixes #

## How to Test

Nothing should change functionally.
There is one minor difference, the opponent game state now considers kicking team `Uncertain` as the HULKs having kick off whereas previously it was only only kicking team `Team::Hulks`. However, this should be solved by #1427.


schluis (approved):
👍🏻

PR #1432, opened 2024-09-03
Twix Lüfter abschalten (mal wieder)

knoellle:
## Why? What?

There was an unnecessary change notification leading to twix repainting at vsync frequency, even when incoming data is much slower or even when disconnected.

Fixes #

## How to Test

Testing changes: compare to main by looking at CPU usage.
Testing possible regressions: check if there are any cases where the UI isn't repainted but should.

### DIFF
--- a/crates/communication/src/client.rs
+++ b/crates/communication/src/client.rs
@@ -328,7 +328,6 @@ impl Client {
                     }
                 }
             }
-            let _ = self.change_watch.send(());
         }
         // TODO: properly shut down open tasks, like the protocol
     }


schmidma:
I am confused that this should lead to redraw in vsync frequency?! The client loop should only spin if there is a change in any of the incoming event channels. Like connecting, disconnecting, etc. When idle, this loop should be idle as well. What do I miss?

knoellle:
One of the "Events" is `GetState`, which is called during every render cycle, meaning also the repaint request was done every frame.

schmidma:
Ah, true... missed that. Thanks.

But should we then remove this completely? In the proposed version, we loose updates when disconnected or connected, etc... do we?

knoellle:
The `Protocol` sends its own change events but the `GetStatus` messages never reach the protocol.
Thus, this change prevents overly ambitious redraws without lost wakeups from connection events.

schmidma (approved):

PR #1451, opened 2024-10-02
prefix with docker.io

schmidma:
## Why? What?
```
Error: creating build container: short-name "rust:1.78.0" did not resolve to an alias and no unqualified-search registries are defined in "/etc/containers/registries.conf"
```


knoellle (approved):
LGTM

PR #1452, opened 2024-10-02
Run CI workflow on github runners

schmidma:
## Why? What?

I did not get buildah running on our runners as they are inside an LXC container... This (as well as `meta-nao`) now moves the container building and releasing stuff to the github hosted runners


knoellle (approved):
LGTM

PR #1461, opened 2024-10-18
remove impl for default when #[default] is enough

MaikRe:
## Why? What?

We have a few spots where we custom implement default() even though it is unnecessary.
This cleans that up.

Fixes #

### DIFF
--- a/crates/types/src/initial_look_around.rs
+++ b/crates/types/src/initial_look_around.rs
@@ -17,7 +17,7 @@ pub enum BallSearchLookAround {
 impl Default for BallSearchLookAround {
     fn default() -> Self {
         Self::Center {
-            moving_towards: Side::Left,
+            moving_towards: Side::default(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


schmidma:
I have some concerns regarding the introduction of `default` constructors for these types.

Firstly, the idea of a "default" value implies a universally applicable and meaningful starting point for a type. However, for many types, like `Side` in this instance, there is no inherently obvious or universally correct default. Whether `Left` or `Right` should be set as the default depends heavily on the context in which the type is used. By assigning a default value that lacks contextual relevance, we risk introducing ambiguity and reducing the explicitness that is one of Rust's strengths.

Moreover, relying on `::default()` can obscure the developer's intent. When reviewing code, it becomes less clear why a particular default value was chosen, making reasoning about the code's behavior more challenging. This could potentially lead to subtle bugs or misunderstandings, especially if the code is revisited or maintained by someone who wasn't part of the original design decision.

MaikRe:
Those are really good points that I did not consider!

PR #1480, opened 2024-10-30
fix center circle violation

oleflb:
## Why? What?

Fixes the robot walking through the center circle obstacle if the kicking team is unknown. Was a bug of #1427.

Fixes #

## ToDo / Known Issues

## How to Test

- Let the robot score a goal and it doesnt walk through the center circle back to its pose


knoellle (approved):
LGTM

PR #1496, opened 2024-11-09
Update pose detection to yolo v11

ThagonDuarte:
## Why? What?

We experienced multiple issues with the old yolo pose detection results during RoboCup 2024. Some poses were not detected at all - arms straight up and either completely parallel or even pointing towards each other - or low background contrast to the person of interest resulted in bad confidence values and poses thus being rejected. 
This is an update from yolo v8 n pose to yolo v11 n pose. Initial tests resulted in substantially better detection of the challenging arms straight up pose as compared to the v8 version.   

||yolov8n-pose|yolov11n-pose|
|---|---|---|
|`ObjectDetectionTop.additional_outputs.inference_duration.millis`| ~270ms mean (with range 260 - 290ms) |~250ms mean (with range 220 - 270ms) |

Fixes #

## ToDo / Known Issues

## How to Test

- Setup a visual referee scenario
- Check all necessary poses for visual referee are detected and correctly filtered


crates/object_detection/src/pose_detection.rs
```diff
@@ -32,6 +29,8 @@ const DETECTION_IMAGE_HEIGHT: usize = 480;
 const DETECTION_IMAGE_WIDTH: usize = 192;
 const DETECTION_IMAGE_START_X: usize = (640 - DETECTION_IMAGE_WIDTH) / 2;
 
+const EXPECTED_OUTPUT_NAME: &str = "1525";
+
```
oleflb, line 33: Just a thought, maybe you can edit the xml by hand (technical term: model surgery :wink:) to remove the other outputs and also name this one properly

oleflb (changes requested):

crates/object_detection/src/pose_detection.rs
ThagonDuarte: Done

oleflb (approved):
Nice

PR #1497, opened 2024-11-09
Add parameter to force enable pose_detection

ThagonDuarte:
## Why? What?

It's annoying to always setup a visual referee scenario, when simply wanting to test the pose detection.
This adds a simple parameter called `override_pose_detection`, that overrides the enable condition for the `pose_detection.rs` module.   

Fixes #

## ToDo / Known Issues

## How to Test

- Upload to NAO with `override_pose_detection = true` in the `default.json`
- In twix, enable the `Pose Detection` overlay in an image tab and see if pose detection is running


oleflb (approved):
LGTM

PR #1507, opened 2024-11-19
Add `secsf32` and `millis` path_serde path to `Duration`

ThagonDuarte:
## Why? What?

It was annoying to read the `Duration` in twix. 265314121 ns is just not as legible as 265 ms. 
Adds a `.millis` and `.secs_f32` path implementation to `Duration` path_serde. 
In twix one can now for example subscribe the text `ObjectDetectionTop.additional_outputs.inference_duration.millis` to see the `Duration` milliseconds. 

Fixes #

## How to Test

- Subscribe to any Duration in twix with `.millis` and `.secs_f32` appended


oleflb (approved):
Finally! Nice

Issue #1185, opened 2024-07-15
Consider Side-Kicks in the Ball Model


schmidma:
When we do side-kicks, the robot is unable to see the ball. This leads to several side kicks, as the ball model does not register to be moved.

Idea to fix this:
Use the event of an actually executed side-kick to displace the ball model, or even loose it entirely. Both is better than the current state.

oleflb:
closed in #1523

PR #1523, opened 2024-11-27
Sidekicks in ball model

pejotejo:
## Why? What?

Now the Ball filter forgets the kickdecisions after the robot has kicket the ball. Therefore when the robot does side kicks, it goes back to ball search. See #1185 

Fixes #

## ToDo / Known Issues

## How to Test

disable forward and turn kicks and let the robot kick the ball


schluis:
Did you verify that the ball was really always kicked when the walking engine selected `Mode:Kicking`?

oleflb:
> Did you verify that the ball was really always kicked when the walking engine selected `Mode:Kicking`? 

That's pretty much impossible, especially for side kicks. The idea is for the robot to go looking for the ball after a side kick, instead of assuming the ball is in the same position

larskna (approved):

Issue #1524, opened 2024-11-28
Refinement of Turnkick Parameters and Behavior


schluis:
The Turnkick implementation has several parameter and behavior inconsistencies that need to be addressed for improved performance and stability. Below is a detailed analysis and proposed solutions:

---

#### 1. Turnkick Rotation Behavior
- **Prestep Rotation**: The robot rotates by `"turn": -0.8`.
- **KickStep Rotation**: The robot rotates by `"turn": -0.2`.

This creates the following scenarios depending on whether the sign changes with the foot switch:
- If the sign **changes**, the total rotation sums to ~1 radian (approx. 54 degrees).
- If the sign **does not change**, the total rotation is -0.6 radians (approx. 34 degrees).

---

#### 2. Configured Kick Direction
The configured kick direction is set to **0.9 radians**. This seems inconsistent with the actual behavior, suggesting calibration or parameter alignment issues.

- **Observed behavior during RoboCup (RC) and RoHOW testing**:
  - The robot appeared to expect a larger shot angle, but the actual kick had a smaller angle. This aligns with the cumulative rotation (-0.6 radians) versus the configured direction (0.9 radians).

---

##### Forward Speed Adjustment
- **Current** `"forward": 0.06` corresponds to a movement speed of approximately **240 mm/s**, which is very fast for a single walking step.
- **Suggestion**: Reduce `"forward"` to **0.03**. This would likely:
  - Improve stability and reduce fall risks.
  - Maintain reasonable shot range.

##### Timing Adjustment for Shot Step
- Consider extending the walking height interpolation during the kick step. This allows:
  - The kicking foot to stay in the air longer.
  - Joint offsets (e.g., HipPitch) to recover closer to zero before foot contact, mitigating lagging issues.

##### Directional Calibration
- Verify if the Turnkick aligns with the intended shot direction:
  - Determine if parameter changes affect accuracy positively or negatively.
  - Ensure the behavior matches expectations for shot angles.

---

#### Summary
- Evaluate and align Turnkick rotation (Prestep and KickStep) with the intended kick direction.
- Adjust forward movement speed and timing to improve stability and performance.
- Verify calibration and behavior consistency during testing.

Best regards,  
Philip

PR #1525, opened 2024-11-28
Tune kick and walk parameters

philipniklas-r:
## Why? What?

Handles Issue #1524 
Also tune walk parameters while I am at it.

Reasoning for the kicks:
- Forward kick: do less large forward steps, which caused falls
- Turn kick: prestep shall move feet slightly together, to make kick step more stable. Also less forward step size, which caused falls
- Turn kick orientation: judging from videos, the kick is often kick with less turn in reality, therefore reduce planned direction.

Reasoning for the walking:
- foot_lift_apex => how high the swing foot is lift up from the ground.
  - I see no advantage from changing the height. This only causes problems (e.g. doing a large side step followed by executing a zero step. Therefore while the feet are moving together again the swing height is only 8 mm instead of an actually needed higher value. This makes no sense -> changed to a constant value)
  - foot_lift_apex_increase set to 0, as this height should be constant. The only expected side effect is when stopping and standing still and starting to walk again before the robot settled from left over swinging. If this is actually the case, this special problem should be handled separate.
- foot_offset_left -> This kind of defines the origins of the legs/soles. Shifting both 50 mm to the sides makes sense (because of the kinematic of the robot). But then actually use the 50 mm from the specification. If the robots like to fall sideways (which a slight increase could handle) then there is another problem, which should not be handled with those parameters.
- step_duration -> For forward and turn steps I see no advantage from increasing the step duration. Only side steps need a correction.
  - Therefore: Max side speed of 0.1 and increase factor of 0.6 => 0.1 * 0.6 = 0.06. Step duration for max side step => 0.25 ms + 0.06 ms = 0.31 ms.
  - I recommend to implement an own variable for the step height duration. The robot only needs a slower interpolation for the swing height, but not for the side translation interpolation. As long as those use the same duration variable, the full step duration needs to be increased.
- torso_offset -> With the code rewrite in April it seems the torso_offset got removed and replaced with the torso tilt. From my testing a translation shift works better, therefore use the "typical" 12 mm shift, to shift the feet backwards and therefore the torso forward. In case the sign is wrong, change the -0.012 to 0.012.
  - torso_tilt -> The tilt should no longer be necessary.
- starting_step -> The first step should use the same duration, but half of the normal swing height. NOTE: I assume that the step_duration increase is applied for the starting step too. Therefore starting with a large side step should work in theory
- step_midpoint -> what the fuck? why were the mid points shifted ??? Judging from the commit history, those seems to be adjusted very often when calibrating the walk parameters. I am open to feedback and other point of views, but I would recommend: don't touch them. If there are stability problems -> add a separate step height duration variable instead.

## ToDo / Known Issues

- I can't code in rust and doing the below ideas by myself :(
- Also I don't know if the sign for `torso_offset` is correct.
- PR #1514 needs to be merged first. Afterwards `step_duration_increase.left` might need to be adjusted.

Add a step height duration variable, similar to the normal step duration one. step height duration should also be at least as big as the normal step duration value, but it is allowed to be bigger.

With that the side walking could interpolate the step height slower (with the mid point being still fixed at 125 ms, or 50% of the normal step duration). Also the kicks could use it too.

In both cases a too early support foot switch could be prevented. The increase could be from the (now configured by be) normal 250 ms step duration to something like 280-300 ms. Such "small" increase is sufficient enough.

For the kicks, a parameter for the turn values representing to which % time of the step the turn value should be finished could improve the accuracy.

## How to Test

First merge PR #1514.

For the walk parameter changes:
- Let the robot walk around and play with the ball a bit. In case it is tilting too far back/falling a lot backwards, change `torso_offset` from -0.012 to 0.012, because this could be a case of a false sign for this parameter.
- Also look if the side walking and walking around the ball looks better (e.g. being faster and also more stable).
- Check if the first walking step looks good in different scenarios

For the kicks:
- Disable all kicks except the forward kick. Check if the kick is stable and also kicks the ball far
- Disable all kicks except the turn kick. Check if the kick is stable and also kicks the ball far

In both cases I recommend to also test with the main branch, to get a good comparision of before and after. Also record the testing with a smartphone/camera. This is helpful to better compare the before and after state.

### DIFF
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -535,7 +535,7 @@
   "kick_steps": {
     "forward": [
       {
-        "base_step": { "forward": 0.04, "left": 0.0, "turn": 0.0 },
+        "base_step": { "forward": 0.02, "left": 0.0, "turn": 0.0 },
         "step_duration": { "nanos": 260000000, "secs": 0 },
         "foot_lift_apex": 0.015,
         "midpoint": 0.5,
@@ -543,7 +543,7 @@
         "ankle_pitch_overrides": null
       },
       {
-        "base_step": { "forward": 0.08, "left": 0.0, "turn": 0.0 },
+        "base_step": { "forward": 0.05, "left": 0.0, "turn": 0.0 },
         "step_duration": { "nanos": 280000000, "secs": 0 },
         "foot_lift_apex": 0.02,
         "midpoint": 0.5,
@@ -579,7 +579,7 @@
     ],
     "turn": [
       {
-        "base_step": { "forward": 0.0, "left": 0.0, "turn": -0.8 },
+        "base_step": { "forward": 0.0, "left": 0.005, "turn": -0.8 },
         "step_duration": { "nanos": 260000000, "secs": 0 },
         "foot_lift_apex": 0.01,
         "midpoint": 0.5,
@@ -587,7 +587,7 @@
         "ankle_pitch_overrides": null
       },
       {
-        "base_step": { "forward": 0.06, "left": 0.0, "turn": -0.2 },
+        "base_step": { "forward": 0.03, "left": 0.0, "turn": -0.2 },
         "step_duration": { "nanos": 260000000, "secs": 0 },
         "foot_lift_apex": 0.02,
         "midpoint": 0.5,
@@ -1078,7 +1078,7 @@
     "turn": {
       "position": [-0.15, -0.11],
       "position_offset": [0.0, 0.0],
-      "orientation": 0.9,
+      "orientation": 0.8,
       "reached_thresholds": [0.04, 0.04, 0.1],
       "shot_distance": 3.5,
       "enabled": true


etc/parameters/default.json
```diff
@@ -477,12 +477,12 @@
...
-      "foot_lift_apex": 0.009,
-      "step_duration": { "nanos": 200000000, "secs": 0 }
+      "foot_lift_apex": 0.0065,
+      "step_duration": { "nanos": 250000000, "secs": 0 }
     },
     "step_midpoint": {
-      "forward": 0.45,
-      "left": 0.35,
+      "forward": 0.5,
+      "left": 0.5,
```
schluis, line 485: The idea behind this is to do the interpolation similar as you for side-steps. You always reach the mid-point after a fixed time and increase the other half of the interpolation (as far as I understood you)

schluis (changes requested):
I tested the parameters.

I like the kick parameters. This looks equal / better.
The walking parameters unfortunately don't seem to improve the walking quality. Have you considered testing on a robot in your lab? The deploying should be fairly easy, just update the `team.toml`, set your team number and add a nao, run `pepsi gammaray <ip>` and run `pepsi upload <ip>`

philipniklas-r:
> The walking parameters unfortunately don't seem to improve the walking quality. Have you considered testing on a robot in your lab? The deploying should be fairly easy, just update the `team.toml`, set your team number and add a nao, run `pepsi gammaray <ip>` and run `pepsi upload <ip>`

Did you remember to combine it with the PR of #1514 ?

etc/parameters/default.json
philipniklas-r: Yes, you understood correctly.
If I see it correctly, side walks are (with my suggested parameters) increased in the duration by 16.5 ms. Less than I wanted. I planned to increase it to from 250 ms to about 285 ms => `step_duration_increase` needs to be 1.4.

(35 ms increase -> 35 / 250 = 0.14
Duration increase is based on side step size times increase factor -> 0.1 * 1.4 = 0.14)

With that a midpoint a 125 ms would mean a value of 125 / 285 = 0.43.
(*Keeping in mind, that a separate interpolation time for the translation/rotation and the height would be better)

philipniklas-r:
> > The walking parameters unfortunately don't seem to improve the walking quality. Have you considered testing on a robot in your lab? The deploying should be fairly easy, just update the `team.toml`, set your team number and add a nao, run `pepsi gammaray <ip>` and run `pepsi upload <ip>`
> 
> Did you remember to combine it with the PR of #1514 ?

In case you did, on one hand, I would need to see it on a nao. For that I would actually need to set everything up on my computer 😅 It could be that the walk might only improve in minor details, but overall does not decrease the fall rate further (which #1514 already should do mostly).

On the other hand, judging from the fall numbers of the RoboCups in the past years, it could be that the walk is (again with #1514) already as stable as it can get, without adding further features. 
Therefore if you agree, I can change the PR to only include the kick parameters. I would only add the walk parameters, as those are more cleaned up and simplify the walk.

schluis:
> > The walking parameters unfortunately don't seem to improve the walking quality. Have you considered testing on a robot in your lab? The deploying should be fairly easy, just update the `team.toml`, set your team number and add a nao, run `pepsi gammaray <ip>` and run `pepsi upload <ip>`
> 
> Did you remember to combine it with the PR of #1514 ?

No, but I don't think this would make a large difference. I tested mainly walking forwards and sidesteps.

schluis:
> > > The walking parameters unfortunately don't seem to improve the walking quality. Have you considered testing on a robot in your lab? The deploying should be fairly easy, just update the `team.toml`, set your team number and add a nao, run `pepsi gammaray <ip>` and run `pepsi upload <ip>`
> > 
> > 
> > Did you remember to combine it with the PR of #1514 ?
> 
> In case you did, on one hand, I would need to see it on a nao. For that I would actually need to set everything up on my computer 😅 It could be that the walk might only improve in minor details, but overall does not decrease the fall rate further (which #1514 already should do mostly).
> 
> On the other hand, judging from the fall numbers of the RoboCups in the past years, it could be that the walk is (again with #1514) already as stable as it can get, without adding further features. Therefore if you agree, I can change the PR to only include the kick parameters. I would only add the walk parameters, as those are more cleaned up and simplify the walk.

Yes, that sounds good! Maybe you're right 👍🏻

schluis (approved):
Thanks for you contribution! 🚀

PR #1072, opened 2024-06-23
Switch to vendored Lua

h3ndrk:
## Why? What?

When compiling `mlua` unvendored, you need to install Lua on your system. Especially under Windoof, this can be painful, therefore we want to switch to a vendored version.

Does not heilmachen #1071 completely because we still need further fixes because we are still using symlinks which are not supported under Windoof.

## ToDo / Known Issues

It's perfect

## Ideas for Next Iterations (Not This PR)

It's perfect

## How to Test

Check that Twix compiles under Windoof

Issue #1071, opened 2024-06-23
Can't build tools/twix on Windows


brookman:
Steps to reproduce (on Windows 11):

```
cd tools/twix
cargo build
```

Error:
```
...
error: failed to run custom build command for `mlua v0.8.10`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_DEV_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `C:\develop\rust\hulk\target\debug\build\mlua-38e686f85069d76a\build-script-main` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=LUA_INC
  cargo:rerun-if-env-changed=LUA_LIB
  cargo:rerun-if-env-changed=LUA_LIB_NAME
  cargo:rerun-if-env-changed=LUA_LINK
  cargo:rerun-if-env-changed=LUAJIT_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=LUAJIT_STATIC
  cargo:rerun-if-env-changed=LUAJIT_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
[...]

PR #1530, opened 2024-12-04
vendored lua

oleflb:
## Why? What?

This is a main compatible version of #1072 by @h3ndrk 

When compiling mlua unvendored, you need to install Lua on your system. Especially under Windoof, this can be painful, therefore we want to switch to a vendored version.

Does not heilmachen https://github.com/HULKs/hulk/issues/1071 completely because we still need further fixes because we are still using symlinks which are not supported under Windoof.

Fixes #

## ToDo / Known Issues

## How to Test

* Nothing should change but lua shouldnt be required as a dependency anymore for twix


schmidma (dismissed):

schmidma (dismissed):

schmidma:
@oleflb ping, the CI is unhappy with your `Cargo.lock`

oleflb:
@schmidma Fixed now. Newer versions of cargo are setting the lockfile version to 4 by default and it seems like the cargo version in the sdk cannot handle that

PR #1540, opened 2024-12-18
Fix CI (hopefully)

rmburg:
## Why? What?

CI was broken because the `cargo install` build step created a cargo cache owned by root.

## How to Test

Pray :pray:


schmidma (approved):
Might be fine for now. As discussed, our CI setup is a bit brittle anyway, so lets go with this

PR #1554, opened 2025-01-09
Feature/twix port input

### LINKED ISSUE #1541: Twix address port appending
Append the default port `1337` only when no port is specified in the address bar. If a port is specified, only use that port.

paudar:
## Why? What?

To be able to connect to different Simulations running on the same host it is necessary to enter a port to connect with. Therefore a parsing of the entered ip-address in the top-panel of the TwixApp is added. 

Fixes #1541

## ToDo / Known Issues

Discuss whether an entered ip-address with tailing ':' ist suppose to connect to the standard port 1337.

Check for the validity of entered ip-addresses. (e.g. that the ip address consists of four octets). 

## How to Test

- start Twix
- enter an ip-address without a port
    - In the terminal you can observe: `[INFO] connecting to ws://127.0.01:1337 ... `
- enter an ip-address with a port
    - In the terminal you can observe the use of the given port: `[INFO] connecting to ws://127.0.01:1234 ... `


schluis:
Found a bug: When using twix with e.g. address `10.0.24.33:1337` twix tries to reconnect at the next start to `10.0.24.33:1337:1337`. Currently, only upon `address_input.changed() || address_input.lost_focus()` this string is split correctly

schluis (approved):
I like it! Thanks for you contribution! 🚀

PR #1562, opened 2025-01-09
Vista

knoellle:
## Why? What?

I became the crisis when trying to figure out how all of the nodes relate in the hunt for the "time to reach kick position"-bug.
This tool should help me and others to understand the implicit relationships between nodes.

Fixes #

## ToDo / Known Issues

- Render dependency lines better
- Highlight relevant nodes, fade out nodes not directly related to the currently selected one
- Also show cross-inputs
- Documentation

- Allow dragging nodes around to explore the realm of possible node orders given the current dependency relations.
  This could for example be used to check if a given node can possibly run before another before the user introduces a new dependency.

## How to Test

`cargo run --manifest-path tools/dependency-inspector/Cargo.toml`


schmidma (dismissed):
lgtm

PR #1563, opened 2025-01-10
Fix Enum plot tooltip

oleflb:
## Why? What?

Fixes the enum plot tooltip becoming small after hovering over an enum variant which does not have a lot of text.

Fixes #

## ToDo / Known Issues

It's perfect

## How to Test

* subscribe fall_state and fall_direction for example and do some hovering. should look normal


rmburg (approved):
Sieht gut zu mir :+1:

PR #1581, opened 2025-01-12
Fix clippy lints

schmidma:
What it says, just applies suggested code actions


oleflb:
@schmidma do you know what rust version we need in order for all features you used to be stable?

schmidma:
No, I did not ... :disappointed:

PR #1586, opened 2025-01-15
Upload source distribution

julianschuler:
## Why? What?

This PR adds also uploading the source code when uploading to a NAO to allow for reproducible build of e.g. the replayer at a later stage.

Blocked on #1486.

## ToDo / Known Issues

None.

Currently the whole repository except the `webots` folder is uploaded. We should consider whether we want to exclude other subfolders like e.g. `tools`.

## How to Test

Upload to a NAO, connect to it and observe the source code in the `hulk/logs/source` subfolder.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -311,9 +311,14 @@ impl Nao {
         let mut command = self.rsync_with_nao()?;
         command
             .arg("--mkpath")
-            .arg("--keep-dirlinks")
+            .arg("--copy-dirlinks")
             .arg("--copy-links")
             .arg("--info=progress2")
+            .arg("--exclude=.git")
+            .arg("--exclude=webots")
+            .arg("--exclude=tools/machine-learning")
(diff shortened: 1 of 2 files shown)


oleflb:
Is this finished?

paudar:
Es kann sein, dass mein Rechner noch nicht fertig eingerichtet ist, aber mir fehlen für einen Upload Rechte (siehe angehangenen Terminal Output), die ich auf dem Main nicht benötige:

[fehlermeldung.txt](https://github.com/user-attachments/files/18851671/fehlermeldung.txt)

knoellle:
I think you still have a `naosdk` symlink in your repo which is obsolete since the advent of pepsi max.
Removing that should fix the errors you are seeing.

paudar:
You`re right. That was the Problem.

Is the location `hulk/logs/source` set intentionally (since it is not the same path as in the PR-description)?

paudar (approved):
Looks good for me.

PR #1588, opened 2025-01-18
Add OS version to gammaray status message

knoellle:
## Why? What?

Having juggled a bunch of branches with different OS versions recently, I got tired of guessing which OS version gammaray was actually uploading, so I added it to the status messages.

Should™️  be conflict-free with #1486 

Fixes #

Also print the currently installed version on each robot, i.e. `v10.5.42 -> v10.6.1`

## How to Test

Gammaray a NAO and watch the status messages.


oleflb (approved):
Noice

PR #1589, opened 2025-01-20
Cleanup

rmburg:
Very minor cleanup, to make the diff of upcoming PRs more legible.

The direction was computed three times in a row, and once in a very convoluted way.


rmburg:
Edit: optimized the math a bit, by using the normalized direction we precomputed anyway

schmidma (approved):

PR #1609, opened 2025-01-27
`geometry::Arc`: use angles for `start` and `end`

rmburg:
## Why? What?

What the title says. Previously, we used `Point2`s for the start and end points of arcs.
This made arcs over-defined. A neat side effect of this change is reduced complexity of most of the affected calculations.

## ToDo / Known Issues

It's perfect

It's perfect

## How to Test

Stare at the code.
I can add more unit tests if there are any areas where you are unsure if the changes are correct.


crates/control/src/path_planner.rs
```diff
@@ -333,10 +334,14 @@ impl PathPlanner {
                             .shape
                             .as_circle()
                             .ok_or_else(|| eyre!("obstacle from path node was not a circle"))?;
+
+                        let start_direction = current_node.position - circle.center;
+                        let end_direction = next_node.position - circle.center;
```
schmidma, line 339: I do not like the names, as they overlap with the `Direction` name for naming clock-wise etc.

crates/geometry/src/angle.rs
```diff
@@ -0,0 +1,217 @@
...
+        Self(value)
+    }
+
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
+impl<T: Copy + RealField> Angle<T> {
+    pub fn from_direction<Frame>(value: Vector2<Frame, T>) -> Self {
```
schmidma, line 29: What do you think of `look_at` instead? to reduce the overlapping nomenclature with `Direction`?

schmidma (commented):
only a few comments, leaving the approval to @knoellle as he is assigned to this PR

crates/control/src/motion/step_planner.rs
```diff
@@ -105,11 +105,12 @@ impl StepPlanner {
                 Pose2::from_parts(line_segment.1, rotation)
             }
             PathSegment::Arc(arc) => {
+                let start_point = arc.start_point();
```
knoellle, line 108: I think this variable can be inlined

crates/control/src/path_planner.rs
knoellle: To resolve this, maybe do the `Angle::from_direction` here instead.

crates/geometry/src/arc.rs
```diff
@@ -83,32 +81,27 @@ impl<Frame> Arc<Frame> {
...
-        if (self.direction == Direction::Clockwise) && (angle > 0.0) {
-            angle -= TAU;
-            angle *= -1.0;
-        }
-        if (self.direction == Direction::Counterclockwise) && (angle < 0.0) {
-            angle += TAU;
-        }
-        (angle * self.circle.radius).abs()
+    pub fn start_point(&self) -> Point2<Frame> {
+        self.circle.point_at_angle(self.start)
```
knoellle, line 90: Even if currently not used, I think the existence of this function should imply the existence of an `end_point` function.

crates/geometry/src/line_segment.rs
```diff
@@ -78,14 +78,14 @@ impl<Frame> LineSegment<Frame> {
         self.signed_acute_angle_to_orthogonal(other).abs() < epsilon
     }
 
-    pub fn projection_factor(&self, point: Point2<Frame>) -> f32 {
+    pub fn projection_factor(&self, point: &Point2<Frame>) -> f32 {
```
knoellle, line 81: Why this change?

crates/linear_algebra/src/rotation.rs
```diff
@@ -7,7 +7,7 @@ pub type Rotation3<Frame, To, T = f32> = Transform<Frame, To, nalgebra::UnitQuat
 
 impl<From, To, T> Rotation2<From, To, T>
 where
-    T: SimdRealField + Copy,
+    T: SimdRealField,
```
knoellle, line 10: Why this change?

crates/control/src/motion/step_planner.rs
rmburg: The variable is used twice below, and I don't want to recompute as it's relatively expensive

crates/geometry/src/line_segment.rs
rmburg: To save a copy here. `filter` only gives me a reference.
https://github.com/HULKs/hulk/blob/f595b94255d44e4abe7c26aa80b5ff2c9737e56d/crates/geometry/src/line_segment.rs#L164-L166

crates/linear_algebra/src/rotation.rs
rmburg: Without this, I need to put the `T: Copy` bound on every place where I use a rotation generic over T. `SimdRealField` implies `Clone`, so I replaced the only implicit copy by an explicit `.clone()` in the impl block

crates/geometry/src/angle.rs
rmburg: I'm open to rename this, but I'm not sure if `look_at` is much better. First, I think keeping the `from_` prefix is more idiomatic rust. Second, the argument to the function is a relative vector (a Vector2), and look_at sounds like we're looking at an absolute point in space (a `Point2`).

crates/control/src/path_planner.rs
rmburg: done. I inlined the two variables

crates/geometry/src/angle.rs
```diff
@@ -0,0 +1,217 @@
...
+
+use linear_algebra::{vector, Rotation2, Vector2};
+use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
+
+use crate::direction::Direction;
+
+#[derive(
+    Clone, Copy, Debug, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub struct Angle<T>(pub T);
```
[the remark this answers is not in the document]
rmburg: Ideally one would have only one of the two. Angles were much easier to work with in my project thesis work, and this is why i made this newtype.

Example: I have structs (think `Pose` or `Step`, for example) in the step planner which contain rotations/orientations. These are generic over `T`, because they are also used to store derivatives. For example, the derivative of some cost function by a step somewhere in the step plan.
When computing the loss value for some step plan, we use `Step<f32>` or `Pose<f32>`, but for the gradient we use `Step<Derivative<f32>>` etc.

The issue is that the `Derivative` type, which stores the derivative of some scalar by arbitrarily many input values, does not implement the necessary traits to be used in a `UnitComplex`. The issue is the `Unit` wrapper: it is not possible to e.g. calculate the "magnitude" of a `Derivative` type, which is needed for the normalization of the `Unit` wrapper.

We can talk about this in more detail in person if needed.

crates/geometry/src/line_segment.rs
knoellle: Those two floats are probably in a register already or the function may be optimized away completely.
I think we copy our linear algebra types almost everywhere.

"Optimizations" at this level are irrelevent I think

rmburg: I completely agree, this was not meant as an optimization. I could just as well create an explicit copy in `filter`.

knoellle: Please do

knoellle:
This branch has a lot of conflicts, please resolve.

rmburg:
> This branch has a lot of conflicts, please resolve.

done

crates/geometry/src/line_segment.rs
rmburg: 2c1cd536a027dce5479a68965cfb4b48f4f987af

crates/geometry/src/arc.rs
```diff
@@ -21,8 +19,8 @@ use crate::{circle::Circle, direction::Direction};
 )]
 pub struct Arc<Frame> {
     pub circle: Circle<Frame>,
-    pub start: Point2<Frame>,
-    pub end: Point2<Frame>,
+    pub start: Angle<f32>,
+    pub end: Angle<f32>,
```
knoellle, line 23: Replace angle with `Orientation2`

oleflb (approved):

PR #1632, opened 2025-02-15
Add `pepsi gamebranch` subcommand to create a game branch from a `deploy.toml` file

julianschuler:
## Why? What?

This PR implements adds a `pepsi gamebranch` subcommand for creating a game branch from a `deploy.toml` file. The game branch is intended to be used with `pepsi pregame`, replacing the use of positional arguments (i.e. no more `pepsi pregame smd SPL_A 33:1 [...]` to set the location, network, etc.).

A corresponding example file is provided as `deploy.toml.example`.

## ToDo / Known Issues

None.

- [x] Add a script for outputting the remote/branch for deployment.
- [ ] Add an additional argument for performing the pregame to a subset of configured NAOs.

## How to Test

Copy the deploy.toml.example file, adjust it accordingly and execute a pregrame:

```sh
cp deploy.toml.example deploy.toml
helix deploy.toml
./pepsi gamebranch
./pepsi pregame
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,5 @@ rust-toolchain.toml
 
 site/
 
+deploy.toml
 logs
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+print_help() {
+    cat <<-__helpText__
+Usage: $0 <remote> <branch>
+
+Deploys a branch from a given remote by merge-squashing it to the current one.
+__helpText__
+}
+
+if [ $# -lt 2 ]; then
+  print_help
+  exit
+fi
+
+remote="$1"
+branch="$2"
+
+# ignore errors from the deploy remote not being present
+git remote remove deploy 2> /dev/null
+
+# exit on error for the following commands
+set -e
+
+git remote add deploy "https://github.com/$remote/hulk.git"
+git fetch deploy $branch
+git merge --squash "$remote/$branch"
+git commit -m "$remote/$branch"
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -24,4 +24,4 @@ set -e
 
 git remote add deploy "https://github.com/$remote/hulk.git"
 git fetch deploy $branch
-git merge --squash "$remote/$branch"
+git merge --squash "deploy/$branch"
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -25,4 +25,3 @@ set -e
 git remote add deploy "https://github.com/$remote/hulk.git"
 git fetch deploy $branch
 git merge --squash "$remote/$branch"
-git commit -m "$remote/$branch"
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -32,4 +32,5 @@ rust-toolchain.toml
 
 site/
 
+deploy.toml
```
oleflb, line 35: we don't want to commit this? This is also not included in the source distribution if listed in the .gitignore, right?

scripts/deploy
```diff
@@ -0,0 +1,27 @@
...
+remote="$1"
+branch="$2"
+
+# ignore errors from the deploy remote not being present
+git remote remove deploy 2> /dev/null
+
+# exit on error for the following commands
+set -e
+
+git remote add deploy "https://github.com/$remote/hulk.git"
```
oleflb, line 25: haha, lets hope nobody renames his/her remote :D Feel free to resolve if you don't think this is a concern

tools/pepsi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pepsi"
-version = "4.0.0"
+version = "4.1.0"
```
oleflb, line 3: technically this is a breaking change, since you removed many arguments from `pregame`

tools/pepsi/src/pre_game.rs
```diff
@@ -58,69 +53,30 @@ pub struct PreGameArguments {
         long,
         value_delimiter=',',
         value_parser = parse_key_value::<String, usize>,
-        default_value = "Control=1,VisionTop=30,VisionBottom=30,SplNetwork=1",
     )]
-    pub recording_intervals: Vec<(String, usize)>,
+    pub recording_intervals: Option<Vec<(String, usize)>>,
```
oleflb, line 57: Oh wait this is the reason pregame messed with the recording intervals?

oleflb (changes requested):

.gitignore
julianschuler: Yes, since if the `deploy.toml` were to be tracked, switching to and creating the a branch using `pepsi gamebranch` could result in conflicts. This unfortunately also excludes it from the source distribution, though we might be able to include in manually (next iteration).

scripts/deploy
julianschuler: You mean like someone changing their name? Not sure, whether that is actually a problem, since the `show_deploy` script will resolve the correct remote information from the URL, independent of any local naming (i.e. origin can still point e.g. julianschuler and will be resolved to julianschuler).

tools/pepsi/Cargo.toml
julianschuler: Fair point, since #1659 will also remove some arguments (this time for `postgame`), do we want to postpone the major version bump to this PR instead? :D

tools/pepsi/src/pre_game.rs
julianschuler: Do you mean the previous version of `pregame` ignoring any recording intervals set using e.g. `pepsi recording`?. Yes, this was indeed the case due to the default implementation, which are now only overwritten if any recording intervals are actually specified.

oleflb (approved):

PR #1634, opened 2025-02-16
Time to reach rework

knoellle:
## Why? What?

This PR refactors
- Role assignment
- Time to reach kick position calculation
- Team ball handling

I still haven't been able to reproduce the exact bug related to time to reach that we found at RoboCup last year but this PR should hopefully fix it.

Fixes #

## ToDo / Known Issues

There are still some todos left in the code

- Use network balls as percepts in the ball filter

## How to Test

Best approach is probably to play around with robots, both in simulation and on real hardware.


crates/spl_network_messages/src/lib.rs
```diff
@@ -34,10 +35,17 @@ impl Default for HulkMessage {
 pub struct StrikerMessage {
     pub player_number: PlayerNumber,
     pub pose: Pose2<Field>,
-    pub ball_position: Option<BallPosition<Field>>,
+    pub ball_position: BallPosition<Field>,
+    // TODO: make non-optional
```
schmidma, line 39: What is blocking you from that? Is this something for a next iteration?

crates/control/src/motion/stand_up_back.rs
```diff
@@ -62,19 +62,15 @@ impl StandUpBack {
...
+            self.interpolator
+                .advance_by(last_cycle_duration, condition_input);
 
-                Some(self.interpolator.estimated_remaining_duration())
-            } else {
-                self.interpolator.reset();
-                None
-            };
+            self.interpolator.estimated_remaining_duration()
+        };
```
schmidma, line 73: This does not look right?! You somehow miss the reset now?

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -90,27 +79,17 @@ impl TimeToReachKickPosition {
...
-            .fill_if_subscribed(|| time_to_reach_kick_position);
-        *context.time_to_reach_kick_position = time_to_reach_kick_position.unwrap_or(Duration::MAX);
+        let time_to_reach_kick_position = [
+            walk_time,
+            *context.stand_up_back_estimated_remaining_duration,
+            *context.stand_up_front_estimated_remaining_duration,
+            time_to_turn,
+        ]
+        .into_iter()
+        .fold(Duration::ZERO, Duration::saturating_add);
```
schmidma, line 89: Why is this a fold and not a sum? Is the saturation property actually so important? Do we run into overflows?

crates/control/src/time_to_reach_kick_position.rs
```diff
@@ -16,67 +14,58 @@ pub struct TimeToReachKickPosition {}
...
-                            Some(line_segment.1.coords().angle(&Vector2::x_axis()).abs())
+                        PathSegment::Arc(_) => {
+                            length / context.configuration.path_planning.arc_walking_speed
                         }
-                        _ => None,
-                    },
+                    }
+                })
+                .sum(),
+        );
```
schmidma, line 61: Might be a little picky, but try to reduce some levels of indentation ;)

e.g.
```rust
        let secs = dribble_path
            .iter()
            .map(|segment: &PathSegment| {
                let length = segment.length();
                match segment {
                    PathSegment::LineSegment(_) => {
                        length / context.configuration.path_planning.line_walking_speed
                    }
                    PathSegment::Arc(_) => {
                        length / context.configuration.path_planning.arc_walking_speed
                    }
                }
            })
            .sum();
        let walk_time = Duration::from_secs_f32(secs);
```

crates/control/src/team_ball_receiver.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        // === Obstacles ===
+        // let sender_position = ground_to_field.inverse() * spl_message.pose.position();
+        // if spl_message.player_number != *context.player_number {
+        //     network_robot_obstacles.push(sender_position);
+        // }
+
+        let team_ball = self.get_best_received_ball(
+            context.cycle_time.start_time,
+            context.striker_trusts_team_ball.mul_f32(4.5),
```
schmidma, line 105: Should this be dependent on this other parameter?

schmidma (changes requested):
Did not have a look at `crates/control/src/role_assignment.rs` yet. Nevertheless, here are already some questions and comments ;)

crates/control/src/motion/stand_up_back.rs
```diff
@@ -62,19 +62,15 @@ impl StandUpBack {
     }
 
     pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
-        let stand_up_back_estimated_remaining_duration =
-            if let MotionType::StandUpBack = context.motion_selection.current_motion {
-                let last_cycle_duration = context.cycle_time.last_cycle_duration;
-                let condition_input = context.condition_input;
+        let stand_up_back_estimated_remaining_duration = {
```
MaikRe, line 65: Reset is definitely missing and breaks standup, use something like:

```suggestion
        if let MotionType::StandUpBack = context.motion_selection.current_motion {
            let last_cycle_duration = context.cycle_time.last_cycle_duration;
            let condition_input = context.condition_input;

            self.interpolator
                .advance_by(last_cycle_duration, condition_input);
        } else {
            self.interpolator.reset();
        };
        let stand_up_back_estimated_remaining_duration =
            self.interpolator.estimated_remaining_duration();
```

crates/control/src/motion/stand_up_back.rs
```diff
@@ -34,18 +34,18 @@ pub struct CycleContext {
 
     condition_input: Input<ConditionInput, "condition_input">,
     cycle_time: Input<CycleTime, "cycle_time">,
-    motion_selection: Input<MotionSelection, "motion_selection">,
     angular_velocity:
         Input<Vector3<Robot>, "sensor_data.inertial_measurement_unit.angular_velocity">,
 
     motion_safe_exits: CyclerState<MotionSafeExits, "motion_safe_exits">,
+    stand_up_back_estimated_remaining_duration:
+        CyclerState<Duration, "stand_up_back_estimated_remaining_duration">,
```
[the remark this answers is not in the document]
knoellle: The incorrectness should be fixed™️ now.

The cycler state is now always set to 0 unless the robot is currently performing this motion, in which case it uses the remaining duration from the iterator. The default case of the duration being 0 is perfectly fine since it is only used in an addition.

I need to use a cycler state here because this replaces the time to reach cycler state. We have a loop for certain and I think it makes sense to make the cut for unrolling it in this place.

crates/spl_network_messages/src/lib.rs
knoellle: Is done

crates/control/src/motion/stand_up_back.rs
knoellle: Reset has returned

crates/control/src/time_to_reach_kick_position.rs
knoellle: The remaining duration of the standup motions may be `Duration::MAX` if the motion was aborted.
This happens for example when the robot falls down again during standup.

Adding anything to this value results in an overflow panic.

crates/control/src/time_to_reach_kick_position.rs
knoellle: Done, also cleaned up the turn duration code below.

crates/control/src/team_ball_receiver.rs
```diff
@@ -0,0 +1,150 @@
...
+            .network_message
+            .persistent
+            .iter()
+            .flat_map(|(time, messages)| {
+                messages
+                    .iter()
+                    .filter_map(|message| Some((*time, (*message)?)))
+            })
+            .filter_map(|(time, message)| match message {
+                IncomingMessage::Spl(HulkMessage::Striker(message)) => Some((time, *message)),
```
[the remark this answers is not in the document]
knoellle: Done

crates/control/src/team_ball_receiver.rs
```diff
@@ -0,0 +1,150 @@
...
+
+#[context]
+pub struct CreationContext {}
+
+#[context]
+pub struct CycleContext {
+    cycle_time: Input<CycleTime, "cycle_time">,
+    network_message: PerceptionInput<Option<IncomingMessage>, "SplNetwork", "filtered_message?">,
+    network_message_debug:
+        AdditionalOutput<Vec<(SystemTime, StrikerMessage)>, "network_message_debug">,
```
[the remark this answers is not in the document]
knoellle: Removed

crates/control/src/motion/stand_up_back.rs
knoellle: Reset is back now

crates/control/src/motion/stand_up_back.rs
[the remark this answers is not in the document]
schmidma: Okay, I agree in parts. Maybe we should discuss this in person to reduce the chance for miscommunication.

I currently struggle a little to have the remaining duration set to 0 in the default case. This somehow mixes information of the actual time the interpolation will take to finish with the motion selection of actually standing up... and both do the cycle-loop which you cut here... Maybe that's correct... :thinking:

crates/control/src/team_ball_receiver.rs
```diff
@@ -0,0 +1,155 @@
...
+    striker_trusts_team_ball: Parameter<Duration, "spl_network.striker_trusts_team_ball">,
+
+    team_balls: AdditionalOutput<Players<Option<BallPosition<Field>>>, "team_balls">,
+}
+
+#[context]
+#[derive(Default)]
+pub struct MainOutputs {
+    pub team_ball: MainOutput<Option<BallPosition<Field>>>,
+    pub network_robot_obstacles: MainOutput<Vec<Point2<Ground>>>,
```
schmidma, line 42: I'm fine with the current solution, but this may also be moved to a completely independent separate Node...
What do you think?

crates/control/src/team_ball_receiver.rs
```diff
@@ -0,0 +1,155 @@
...
+            );
+            let in_penalty_kick = game_controller_state.sub_state == Some(SubState::PenaltyKick);
+
+            if in_penalty_shootout || in_penalty_kick {
+                return Ok(MainOutputs {
+                    team_ball: None.into(),
+                    network_robot_obstacles: Vec::new().into(),
+                });
+            }
+        }
```
schmidma, line 96: This can be way earlier at the beginning of the function, can it?

knoellle: Yes, but I'd like to keep updating the internal state during these times. Not important for penalty shootout, but maybe remembering that someone saw a ball a few seconds ago when a penalty kick ends helps.

crates/control/src/team_ball_receiver.rs
knoellle: Yes, was planning on doing that; will do.

crates/control/src/role_assignment.rs
```diff
@@ -740,85 +640,119 @@ fn pick_role_with_penalties(
...
-        } else if needs_assignment(PlayerNumber::Six, penalties, &role_assignment) {
-            role_assignment[PlayerNumber::Six] = Some(optional_role);
-        } else if needs_assignment(PlayerNumber::Seven, penalties, &role_assignment) {
-            role_assignment[PlayerNumber::Seven] = Some(optional_role);
-        }
+    for (player_number, &optional_role) in unassigned_players.into_iter().zip(optional_roles) {
+        role_assignment[player_number] = Some(optional_role)
     }
 
     role_assignment[own_player_number].unwrap_or_default()
```
schmidma, line 663: Can we remove `Default` from `Role` entirely? I do not like to have `Default` for a role, as this depends on the context. There is no general, natural default value to a role. Being a striker is a implementation in some fallback cases of the role assignment.

The consequence of that is to explicitly state that we want to assign striker(s) in respective cases

crates/control/src/team_ball_receiver.rs
schmidma: maybe this gets clearer, if we chop the cycle method into even smaller functions, just to improve readability, to easier realize, that you intentionally first update internal state and then return for penalty shootout.
This need is also visible as you wrote comments for structuring the function. These should probably better be functions with descriptive names. 

something like
```
fn cycle(...) {
  self.do_foo();
  let stuff = self.collect_stuff();
  self.update_state(stuff);

  if penalty_shootout {
    return None;
  }

  self.do_bar();
  return Some;
}
```

crates/control/src/role_assignment.rs
```diff
@@ -155,198 +148,119 @@ impl RoleAssignment {
...
-                self.last_system_time_transmitted_game_controller_return_message
-                    .unwrap(),
-            )? > context.spl_network.game_controller_return_message_interval;
+        self.try_sending_game_controller_return_message(&context, ground_to_field)?;
 
-        let mut send_spl_striker_message = self.last_transmitted_spl_striker_message.is_none()
-            || cycle_start_time
-                .duration_since(self.last_transmitted_spl_striker_message.unwrap())?
-                > context.spl_network.spl_striker_message_send_interval;
+        // TODO: reimplement whatever this did
```
schmidma, line 155: ?? now i'm curious :eyes:

crates/control/src/role_assignment.rs
```diff
@@ -35,10 +39,9 @@ use crate::localization::generate_initial_pose;
 pub struct RoleAssignment {
     last_received_spl_striker_message: Option<SystemTime>,
```
schmidma, line 40: This can never be `None`, can it? This is only `None` in construction and the first cycle immediately sets a `Some` to it because `self.role_initialized == false`

crates/control/src/role_assignment.rs
```diff
@@ -155,198 +148,119 @@ impl RoleAssignment {
...
+                ..
+            })
+        );
 
         let spl_striker_message_timeout = match self.last_received_spl_striker_message {
-            None => false,
+            None => true,
             Some(last_received_spl_striker_message) => {
                 cycle_start_time.duration_since(last_received_spl_striker_message)?
                     > context.spl_network.spl_striker_message_receive_timeout
```
schmidma, line 168: This is a horrible name... I'm trying to wrap my head around what it actually is for minutes now. If I'd only stumble upon that in the parameters, I'd have even less of an idea what this controls

schmidma: This is something like the duration we believe in the last information a striker sent us before we assume the robot which sent us the information is dead and we have to fallback and continue somehow, is it?

schmidma: so it is a `duration_this_robot_believes_in_the_last_information_of_another_striker_before_assuming_it_fell_dead_and_we_have_to_continue_doing_something_else`
@h3ndrk would be happy :laughing:

crates/control/src/role_assignment.rs
```diff
@@ -155,198 +148,119 @@ impl RoleAssignment {
...
-        let mut team_ball = self.team_ball;
-
-        let is_in_penalty_kick = matches!(
-            context.filtered_game_controller_state,
-            Some(FilteredGameControllerState {
-                sub_state: Some(SubState::PenaltyKick),
-                ..
-            })
-        );
         if spl_striker_message_timeout && !is_in_penalty_kick {
```
schmidma, line 172: Don't we have to first handle incoming messages before we decide to distrust a striker because of `spl_striker_message_timeout`?
I'm just thinking of the order in time of things. Incoming messages may be in the past, and the fact that the striker timeouted is happening *now*, so definitely after all possible incoming messages.

In my head, first for example becoming a searcher in the `new_role` changes things for the incoming messages afterwards...

crates/control/src/role_assignment.rs
```diff
@@ -155,198 +148,119 @@ impl RoleAssignment {
...
             match new_role {
-                Role::Keeper => {
-                    team_ball = None;
-                }
-                Role::ReplacementKeeper => {
-                    team_ball = None;
-                }
+                Role::Keeper | Role::ReplacementKeeper => {}
```
schmidma, line 174: We should discuss the meaning/reason of this `match` statement and its implication. How can a striker message timeout and this very own robot be striker?!?! ... to then become loser... I'm confused

crates/control/src/role_assignment.rs
```diff
@@ -355,7 +269,6 @@ impl RoleAssignment {
         } else {
             self.role = new_role;
         }
-        self.team_ball = team_ball;
 
         if let Some(game_controller_state) = context.filtered_game_controller_state {
```
schmidma, line 273: Is it correct that `last_time_player_was_penalized` is updated **after** it is used in this cycle?

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
+                > context
+                    .spl_network
+                    .remaining_amount_of_messages_to_stop_sending
+        })
+}
+
+#[derive(Clone, Copy, Debug)]
+enum Event {
+    None,
+    Striker(StrikerEvent),
```
schmidma, line 453: what about making this a structures enum variant?

knoellle: The inner type is also used as a function argument.

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
-            Some(..) => {
-                return (
-                    current_role,
-                    false,
-                    team_ball_from_seen_ball(detected_own_ball, current_pose, cycle_start_time),
-                )
-            }
-        }
+        return current_role;
     }
```
schmidma, line 491: why do we trigger the state machine during all the other states? i'd expected role assignment to return early in the current way we handle ready and set...

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
-            None => (Role::Loser, false, None), //edge-case, a striker (which should not exist) lost the ball
-            _ => decide_if_claiming_striker_or_other_role(
-                spl_message,
-                time_to_reach_kick_position,
-                player_number,
-                cycle_start_time,
-                filtered_game_controller_state,
-                optional_roles,
-            ),
+    match (current_role, detected_own_ball, event) {
```
schmidma, line 500: can we make the `detected_own_ball` an enum? this would make reading the match arms much easier

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
+        // Striker lost Ball
+        (Role::Striker, false, Event::None | Event::Loser) => match team_ball {
```
schmidma, line 502:
```suggestion
        (Role::Striker, false, Event::Tick | Event::Loser) => match team_ball {
```

just naming that came to mind. what do you think?

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
-            None => (Role::Searcher, false, team_ball), //edge-case, a striker (which should not exist) lost the ball
-            _ => decide_if_claiming_striker_or_other_role(
-                spl_message,
-                time_to_reach_kick_position,
-                player_number,
-                cycle_start_time,
-                filtered_game_controller_state,
-                optional_roles,
-            ),
+        // Striker remains Striker, sends message after timeout
```
schmidma, line 515: why this comment part about "sends message after timeout"?

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
+
+        // Edge-case, another Striker became Loser, so we claim striker since we see a ball
+        // TODO: On main, this sends a striker message immediately, ignoring the spl_striker_message_send_interval
+        //       but not the silence_interval_between_messages.
+        //       With the new implementation this is no longer possible since the message sending
+        //       is only based on the previous and new roles and does not consider the cause of the
+        //       transition.
+        //       This was already broken on main since the message would silently be dropped if the
+        //       edge case occured within the silence interval but might lead to a faster
+        //       striker convergence in this rare circumstance.
```
schmidma, line 526: marking this with a comment for discussion...

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
+        (Role::Striker, true, Event::Loser) => Role::Striker,
+
+        // Loser remains Loser
+        (Role::Loser, false, Event::None) => Role::Loser,
```
schmidma, line 530: We could `use` the nested values for this local scope. allowing to write

```suggestion
        (Loser, false, None) => Role::Loser,
```

the current naming would produce a weird conflict between Event and Role (Striker and Striker), where i'd suggest renaming the events. maybe `StrikerMessage`

what do you think? the current formulation is also not annoyingly long. so i'd also be fine keeping it as is

crates/control/src/role_assignment.rs
```diff
@@ -372,289 +285,320 @@ impl RoleAssignment {
...
-                optional_roles,
-            ),
-            false,
-            team_ball_from_spl_message(cycle_start_time, spl_message),
-        )
+) -> Role {
+    let shorter_time_to_reach =
+        time_to_reach_kick_position.unwrap() < striker_event.time_to_reach_kick_position;
+    let time_to_reach_viable =
+        time_to_reach_kick_position.is_some_and(|duration| duration < Duration::from_secs(1200));
```
schmidma, line 585: do we need this?

crates/control/src/role_assignment.rs
```diff
@@ -667,70 +611,26 @@ fn seen_ball_to_game_controller_ball_position(
     })
 }
 
-fn seen_ball_to_hulks_network_ball_position(
-    ball: Option<&BallPosition<Ground>>,
+fn own_ball_to_hulks_network_ball_position(
```
schmidma, line 614: what about making this and the other similar conversion functions methods of one of the types?

crates/control/src/role_assignment.rs
```diff
@@ -86,19 +88,10 @@ impl RoleAssignment {
         Ok(Self {
             last_received_spl_striker_message: None,
             last_system_time_transmitted_game_controller_return_message: None,
-            last_transmitted_spl_striker_message: None,
+            last_transmitted_striker_message: None,
             role: Role::Striker,
             role_initialized: false,
```
schmidma, line 93: try to remove this variable. this is only used in the very first cycle to initialize `self.role`

crates/control/src/role_assignment.rs
knoellle: Done

crates/control/src/role_assignment.rs
knoellle: Removed

crates/control/src/role_assignment.rs
knoellle: Removed

crates/control/src/role_assignment.rs
knoellle: Removed

crates/control/src/team_ball_receiver.rs
knoellle: cycle function is much shorter now

crates/control/src/role_assignment.rs
```diff
@@ -155,198 +148,119 @@ impl RoleAssignment {
...
         }
 
-        let send_game_controller_return_message = self
-            .last_system_time_transmitted_game_controller_return_message
-            .is_none()
-            || cycle_start_time.duration_since(
-                self.last_system_time_transmitted_game_controller_return_message
-                    .unwrap(),
-            )? > context.spl_network.game_controller_return_message_interval;
+        self.try_sending_game_controller_return_message(&context, ground_to_field)?;
```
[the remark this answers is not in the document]
knoellle: I moved this glorified unwrap_or_default to the individual functions that were using it.
This should avoid confusion if someone wants to use ground to field in the future.

crates/control/src/role_assignment.rs
knoellle: Moved out of the state machine

schmidma (approved):

schluis:
Does this fix https://github.com/hulks/hulk/issues/1325? @knoellle

knoellle:
If the bug I fixed was the cause of the issue, then yes, but I'm not sure.
I haven't been near a nao since this PR reached a testable state. 😭

PR #1636, opened 2025-02-17
upgrade CI image to 1.84.1

schmidma:
## Why? What?

get the newest stuff

Pipeline already ran upstream, and CI image is already published.


knoellle (approved):
LGTM

PR #1646, opened 2025-02-20
Visual referee update: Free kick signal detection

### LINKED ISSUE #1476: Rule Changes: Visual Referee
There are multiple changes in the rule book regarding Visual referee:

1. Game Controller does not send the kicking team anymore. It still sends the SetPlay, but the indication of the kicking team is done visually. 
In most of the situation we can guess the team (Goal Kick, Kickoff, Freekick, Penaltykick, Cornerkick), but in others it is more difficult (Kickin). 

2. Initial Ready signal is not send at all. Motion ins standby will be updated

Also on could think about switching from YOLO to MoveNet

ThagonDuarte:
## Why? What?

The [rule update for the 2025 season](https://spl.robocup.org/wp-content/uploads/SPL-Rules-2025.pdf) introduces a change to the set play of free kicks. Instead of sending the kicking team in the game controller message during all free kicks, now, during `CornerKick`, `GoalKick`, `KickIn`, and `PushingFreeKick` no kicking team will be sent in the game controller message. Instead, the head referee will give a visual signal to indicate the direction of play for the kicking team. This means a signal in the direction of the goal of the HULKs indicates that the Opponent is the kicking team. This signal can be seen below:

Set play visual referee signal

This expands our current visual referee implementation to now also detect this free kick visual referee signal. 
We only try to detect the visual referee signal in `KickIn` and `PushingFreeKick`, since the `kicking_team` during `CornerKick` and `GoalKick` can be deduced from the current ball position and the teams `GlobalFieldSide`, i.e. whether their are home or away. During `KickIn` and `PushingFreeKick`, when `Home` the `DefenderRight` and `MidfielderRight` should look at the referee and do the detection. When `Away`, this is done by `DefenderLeft` and `MidfielderLeft` and the first two non-penalized Searchers that are not the keeper.

This PR only implements the detection of the signal. It does not yet apply the detection to the `kicking_team` in the `FilteredGameControllerState`, since the game controller currently still always sends the kicking team, as it has not been updated yet (see [RoboCup-SPL/GameController3](https://github.com/RoboCup-SPL/GameController3)). 

Fixes #1476 

## ToDo / Known Issues

- [ ] Make NAOs look further up to better see referee during detection

- Determine `kicking_team` directly by looking at the `Penalty` set in the same `GameControllerMessage` as the transition to the `SubState` `PushingFreeKick`.

## How to Test

- In play, trigger the four free kicks  `CornerKick`, `GoalKick`, `KickIn`, and `PushingFreeKick`.
  - Check that during `KickIn` and `PushingFreeKick` only the correct defender and midfielder or searcher look at the referee and do the visual signal detection. 
  - Check that the visual referee free kick pose signal is detected, filtered, and interpreted correctly.


etc/parameters/default.json
```diff
@@ -7,6 +7,7 @@
     "minimum_overall_keypoint_confidence": 0.5,
     "minimum_visual_referee_keypoint_confidence": 0.8,
     "minimum_shoulder_angle": 0.2,
+    "free_kick_signal_angle_range": [1.3,1.8],
```
oleflb, line 10: add space after comma

crates/types/src/pose_kinds.rs
```diff
@@ -16,6 +18,9 @@ use serde::{Deserialize, Serialize};
 )]
 pub enum PoseKind {
     AboveHeadArms,
+    FreeKickPose {
```
oleflb, line 21: Is `FreeKickPose` the correct name? This also happens for corner kicks and kick ins as well

crates/types/src/pose_detection.rs
```diff
@@ -6,9 +6,18 @@ use serde::{Deserialize, Serialize};
...
 use coordinate_systems::Pixel;
 use linear_algebra::{point, Point2};
+use spl_network_messages::Team;
 
 use crate::bounding_box::BoundingBox;
 
+#[derive(
+    Debug, Clone, Copy, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub enum ImageSize {
```
oleflb, line 16: maybe prefer `DetectionRange` or something similar?

crates/types/src/motion_command.rs
```diff
@@ -59,6 +59,7 @@ pub enum MotionCommand {
     },
     Stand {
         head: HeadMotion,
+        should_look_for_referee: bool,
```
oleflb, line 62: I think it would be better to have a separate `HeadMotion::LookForReferee`

crates/spl_network_messages/src/game_controller_state_message.rs
```diff
@@ -325,6 +326,7 @@ impl GameState {
 )]
 pub enum Team {
     Hulks,
+    #[default]
```
oleflb, line 329: Why should `Team::Opponent` be default? I don't think it is a good idea to add this default impl

crates/control/src/role_assignment.rs
```diff
@@ -407,6 +407,16 @@ fn process_role_state_machine(
         if let Some(SubState::PenaltyKick) = game_controller_state.sub_state {
             return (current_role, false, None);
         }
+        if let FilteredGameControllerState {
+            sub_state: Some(SubState::KickIn | SubState::PushingFreeKick),
+            // kicking_team: None, TODO: Change this to None when the game controller is fixed
```
oleflb, line 412: isn't it fixed now?

oleflb (changes requested):
review not yet complete

crates/types/src/pose_kinds.rs
ThagonDuarte: Yes, a free kick is defined as the encompassing term for penalty kick, kick off, kick in, corner kick and pushing free kick, as I understand it.
I did however remove the redundant `Pose` in the kind and renamed the other kind to be more consistent.

etc/parameters/default.json
ThagonDuarte: Done

crates/bevyhavior_simulator/src/bin/visual_referee_free_kick_behavior.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Is used to generate the test functions for cargo test
+#[scenario]
+fn visual_referee_free_kick_behavior(app: &mut App) {
+    app.add_systems(Startup, startup);
+    app.add_systems(Update, update);
+}
+
+#[derive(SystemParam)]
+struct State<'s> {
+    detecting_robots_when_home: Local<'s, usize>,
```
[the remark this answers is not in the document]
ThagonDuarte: Yes, they are a number. Renamed and updated the scenario to new behavior, where Searcher can also look at the referee.

crates/types/src/pose_detection.rs
ThagonDuarte: Renamed to `DetectionRegion`

crates/control/src/role_assignment.rs
ThagonDuarte: Yes, that is however implemented in #1647. Removed the comment here

crates/spl_network_messages/src/game_controller_state_message.rs
ThagonDuarte: True. I only did this, because our implementation of `PathDeserialize` for `Option<T>` requires `Default` and I changed the `kicking_team` to a `Option<Team>`. I don't think there is another way to specialize the implementation of `PathDeserialize` specifically for `Option<Team>`... Dunno what else to do :/

etc/parameters/default.json
oleflb: No you didn't :wink: currently it's `[1.2,2.3]`

crates/types/src/pose_detection.rs
oleflb: :+1:

crates/types/src/motion_command.rs
oleflb: hmm, right now the `LookAtReferee` is the same as `LookAt`. I think it would be better to remove the `target` from the enum and use the referee position directly in the look at node. This way we know that the `LookAtReferee` motion can indeed only look at referees and is not just a renamed LookAt

crates/bevyhavior_simulator/src/bin/visual_referee_free_kick_behavior.rs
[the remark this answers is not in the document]
oleflb: nice :+1:

crates/spl_network_messages/src/game_controller_state_message.rs
oleflb: hmm yes you are right about that. You can still do a manual implementation of `PathDeserialize` when `Team` does not implement `Default`. @schmidma I am a bit confused that `PathDeserialize` for `Option<T>` requires `T: Default`. Looking at the code It looks like a totally different value is deserialized in the `None` case

crates/control/src/behavior/node.rs
```diff
@@ -193,10 +220,64 @@ impl Behavior {
...
+                    actions.push(Action::SupportRight);
+                }
+                _ => actions.push(Action::SupportRight),
+            },
             Role::ReplacementKeeper => actions.push(Action::DefendGoal),
-            Role::Searcher => actions.push(Action::Search),
+            Role::Searcher => match world_state.filtered_game_controller_state {
+                Some(FilteredGameControllerState {
+                    sub_state: Some(SubState::KickIn) | Some(SubState::PushingFreeKick),
+                    // kicking_team: None, TODO: Change this to None when the game controller is fixed
```
oleflb, line 259: TODO?

crates/control/src/behavior/node.rs
```diff
@@ -193,10 +220,64 @@ impl Behavior {
...
+                    penalties,
+                    ..
+                }) => {
+                    let mut first_two_nonpenalized_nonkeeper_playernumbers = penalties
```
oleflb, line 263: `playernumber` -> `player_number` for consistency

crates/control/src/behavior/node.rs
```diff
@@ -193,10 +220,64 @@ impl Behavior {
...
+                        .iter()
+                        .filter_map(|(player_number, penalty)| {
+                            penalty.is_none().then_some(player_number)
+                        })
+                        .skip(1)
```
oleflb, line 268: what if the keeper is penalized? If 1 is penalized and rest of the team is not, 3 & 4 would be selected instead of 2 & 3

crates/control/src/behavior/node.rs
```diff
@@ -193,10 +220,64 @@ impl Behavior {
...
+                        .take(2);
+                    if first_two_nonpenalized_nonkeeper_playernumbers
```
oleflb, line 270: same here

oleflb (changes requested):

crates/control/src/behavior/node.rs
ThagonDuarte: Renamed

crates/control/src/behavior/node.rs
ThagonDuarte: Renamed

crates/control/src/behavior/node.rs
ThagonDuarte: Removed comment

crates/control/src/free_kick_signal_filter.rs
```diff
@@ -0,0 +1,260 @@
...
+    earlier_time: SystemTime,
+    grace_period: Duration,
+) -> bool {
+    cycle_start_time
+        .duration_since(earlier_time)
+        .expect("Time ran backwards")
+        < grace_period
+}
+
+fn unpack_own_detection_tree(
```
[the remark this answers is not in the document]
ThagonDuarte: Renamed

crates/types/src/motion_command.rs
ThagonDuarte: I have removed `target` from the enum. This did however create some ugly defaulting and made the `visual_referee_free_kick_behavior.rs` scenario a bit less thorough in its testing of the behavior. 
But is think your point is definitely valid. Do you think these caveats are acceptable? I'm currently not sure how to handle them differently.

etc/parameters/default.json
ThagonDuarte: Now it really is :wink:

crates/spl_network_messages/src/game_controller_state_message.rs
ThagonDuarte: I did not manage to manually implement `PathDeserialize` for `Option<Team>`, since adding `spl_network_messages`, which defines `Team`, as a dependency to the `path_serde` crate, creates a cyclic dependency :melting_face:. 
I now simply removed the requirement for the `Default` derive for `Option<T>` by just not deserializing in the `None` case. This should be fine, since the value is a `None` anyway, right?

knoellle: Depends on what you mean by "fine". It changes the behavior.
If the previous value was none, we'd construct a default one and recurse with the deserialization.
That was a design decision we made at some point, though I'm not sure we depend on this behavior anywhere currently.

crates/control/src/free_kick_signal_filter.rs
```diff
@@ -0,0 +1,258 @@
...
+            } else {
+                continue;
+            }
+        }
+
+        self.detected_free_kick_detections_queue
+            .truncate(*context.referee_pose_queue_length);
+
+        let (own_detected_kicking_team, number_of_detections) =
+            most_detections(self.detected_free_kick_detections_queue.clone().into());
```
oleflb, line 132: use references here

crates/control/src/free_kick_signal_filter.rs
```diff
@@ -0,0 +1,258 @@
...
+        })
+    }
+}
+
+fn most_detections(detections: Vec<Team>) -> (Team, usize) {
+    let own_detections_hulks: Vec<Team> = detections
+        .iter()
+        .cloned()
+        .filter(|kicking_team| *kicking_team == Team::Hulks)
+        .collect();
```
oleflb, line 161: use `.count()` instead of `.collect().len()`

crates/control/src/free_kick_signal_filter.rs
```diff
@@ -0,0 +1,258 @@
...
+    let own_detections_opponent: Vec<Team> = detections
+        .iter()
+        .cloned()
```
oleflb, line 164: prefer `copied` here

crates/control/src/free_kick_signal_filter.rs
```diff
@@ -0,0 +1,258 @@
...
+            }),
+        ) => Some(Team::Opponent),
+        (
+            GlobalFieldSide::Away,
+            Some(PoseKind::FreeKick {
+                global_field_side: GlobalFieldSide::Home,
+            }),
+        ) => Some(Team::Hulks),
+        _ => None,
+    }
```
oleflb, line 205:
```suggestion
free_kick_signal_pose.and_then(|a| if a.global_field_side == global_field_side { Team::Opponent } else { Team::Hulks })
```

oleflb (changes requested):

crates/spl_network_messages/src/game_controller_state_message.rs
ThagonDuarte: We now throw an error when path deserializing a `None` of an `Option<T>`

crates/control/src/behavior/node.rs
```diff
@@ -156,6 +155,17 @@ impl Behavior {
         actions.push(Action::InterceptBall);
 
         match world_state.robot.role {
+            Role::DefenderLeft
+                if world_state
+                    .filtered_game_controller_state
+                    .clone()
+                    .is_some_and(|filtered_game_controller_state| {
+                        is_free_kick(filtered_game_controller_state, world_state.clone())
```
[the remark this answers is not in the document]
ThagonDuarte: Done

ThagonDuarte:
To be tested during a full test game

oleflb (changes requested):

crates/control/src/ready_signal_detection_filter.rs
```diff
@@ -42,108 +43,121 @@ pub struct CycleContext {
...
     ) -> Result<MainOutputs> {
-        let cycle_start_time = context.cycle_time.start_time;
+        if !matches!(
+            context.game_controller_state.game_state,
+            GameState::Standby { .. }
+        ) {
+            self.detected_ready_signal_queue = Default::default();
+            self.motion_in_standby_count = Default::default();
+            self.detection_times = Default::default();
+            self.ready_signal_state = Default::default();
```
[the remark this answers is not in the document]
ThagonDuarte: Done

crates/control/src/motion/head_motion.rs
```diff
@@ -163,3 +153,65 @@ impl HeadMotion {
...
+            let head_motion_parameters = HeadMotionParameters {
+                inner_maximum_pitch: 0.61,
+                inner_minimum_pitch: -0.61,
+                outer_maximum_pitch: 0.0,
+                outer_minimum_pitch: 0.0,
+                outer_yaw: 1.3,
+                ..Default::default()
+            };
+
+            let controlled_positions = HeadJoints { yaw, ..Default::default() };
```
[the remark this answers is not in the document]
ThagonDuarte: This test in particular was just designed to test for problem I had with having a larger min than max value. I could also add a test, that checks the computed `clamped_pitch` for certain pitches, but that would then be a separate test.

crates/control/src/led_status.rs
```diff
@@ -314,54 +325,60 @@ impl LedStatus {
                     Role::Striker => Rgb::RED,
                     Role::StrikerSupporter => Rgb::TURQUOISE,
                 };
-                let filtered_referee_ready_color = if is_own_referee_ready_pose_detected {
-                    Some(Rgb::YELLOW)
-                } else {
-                    None
-                };
-                let referee_ready_percept_color = if did_detect_any_referee_this_cycle {
+                let majority_voted_viusal_referee_color =
```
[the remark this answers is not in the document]
ThagonDuarte: Done

crates/control/src/free_kick_signal_filter.rs
ThagonDuarte: Done a long time ago and is way outdated

crates/control/src/free_kick_signal_filter.rs
ThagonDuarte: Done a long time ago and is way outdated

crates/control/src/free_kick_signal_filter.rs
ThagonDuarte: Done a long time ago and is way outdated

crates/control/src/game_controller_filter.rs
```diff
@@ -118,7 +119,11 @@ impl GameControllerFilter {
             penalties: message.hulks_team.clone().into(),
             opponent_penalties: message.opponent_team.clone().into(),
             sub_state: message.sub_state,
-            hulks_team_is_home_after_coin_toss: message.hulks_team_is_home_after_coin_toss,
+            global_field_side: if message.hulks_team_is_home_after_coin_toss {
+                GlobalFieldSide::Home
+            } else {
+                GlobalFieldSide::Away
+            },
```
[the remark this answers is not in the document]
ThagonDuarte: Done a long time ago and is way outdated

PR #1647, opened 2025-02-21
Visual referee update: Kicking team filtering

### LINKED ISSUE #1476: Rule Changes: Visual Referee

ThagonDuarte:
## Why? What?

This PR expands and depends on #1646. It uses the detected free kick signals and uses them to filter the kicking team. In the case, we are not sent a kicking team, the filter either determines it by use of the last ball position, or uses the detected free kick signal. This is then propagated to the `FilteredGameControllerState`. 

Fixes #1476 

## ToDo / Known Issues

- [x] Handle not knowing the `kicking_team`
- [x] Use detection to set `kicking_team` and propagate to other Naos.
- [x] Filter `kicking_team` to directly compute it in the `SubState`s `CornerKick` and `GoalKick`.
- Implement bevyhavior simulator scenarios for
  - [x] Check that all roles only perform passive `Action`s while `kicking_team` is not known yet (is `None`), i.e. that no NAO directly walks to the ball and tries to kick it. 
  - [x] Check that `None` as `kicking_team` is handled correctly
  - [x] Check `KickingTeamFilter` works as intended

- Determine `kicking_team` directly by looking at the `Penalty` set in the same `GameControllerMessage` as the transition to the `SubState` `PushingFreeKick`.
- Also use `team_ball` in `KickingTeamFilter`. Depends on  #1634 
- Use `WalkInStand` `MotionCommand` in action `LookAtReferee` to rotate torso in direction of referee

## How to Test

- In play, trigger the four free kicks  `CornerKick`, `GoalKick`, `KickIn`, and `PushingFreeKick`.
  - Check that during `GoalKick` and `CornerKick` the correct `kicking_team` is produced by `KickingTeamFilter` and is correctly propagated to all nodes ready `kicking_team`. 
 - Check that the detected `kicking_team` is propagated correctly to all nodes importing `kicking_team`.

Please also test  where behavior had to be adapted to not knowing the `kicking_team` some of the time:

- [ ] In ball_state_composer.rs a side_factor uses the kicking_team. What is its value for kicking_team == None`?
- [ ] In stand.rs the kicking_team is read multiple times to determine the Half.
- [ ] In localization.rs the kicking_team is used to modify_state().
- [ ] In rule_obstacle_composer.rs the create_penalty_box() uses the kicking_team.
- [ ] In search_suggestor.rs uses the kicking_team to determine the Half.
- [ ] The PathDeserialize implementation of Option<T> requires a default to be defined for the underlying type T. Here this means Team has to receive a default value. Is Opponent a reasonable choice for this?


ThagonDuarte:
Closing as a simplified version as already been merged #1766

PR #1650, opened 2025-02-25
Fix camera matrix correction method

tuxbotix:
## Why? What?

The current implementation of `to_corrected()` applies the new correction on top of existing corrected values. Since this method is used in camera calibration, the parameters obtained by the solver also reflects this issue.

Current state:
* `T = C_o * T_m` at construction (C_o from parameter file)
* `T_n = C_n * T` when using  `to_corrected()`
* So the total correction `C_t` will be `C_o + C_n`
* Neither `C_o`, nor `T_m` are stored in `CameraMatrix`, so I'll have to read from parameter files and re-calculate this which is wonky IMO.

What we need:
* `T = C_o * T_m` at construction (C_o from parameter file)
* `T_n = C_n * T_m` when using  `to_corrected()`

Changes:
* This PR fixes it by storing the transformation values as per the theoritical model `T_m` so that corrections can always be done from that state. It is also verified by the added unit test.
* Also introduces robot rotation values as extra corrections.

Fixes #

## ToDo / Known Issues

- [ ] Discuss and evaluate if we can do this differently

## How to Test


oleflb (changes requested):
I did not fully get your explanation. Why do you have to apply the correction differently now? Reading the test case it seems like you want the correction to be idempotent, is this what you are after?

tuxbotix:
> I did not fully get your explanation. Why do you have to apply the correction differently now? Reading the test case it seems like you want the correction to be idempotent, is this what you are after?

I'm sorry for my bad explanation, the reasoning starts from how it is constructed by camera matrix calculator Vs how get_corrected is used.

The parameters give the full amount of correction needed and camera matrix calculator applies it to the bead to camera matrix before building CameraMatrix.

In the current state, get_corrected means whatever additional offsets needed, not the absolute correction. So the calibration solver using this function also returns relative values.
That means I have to re read the existing parameters and calculate the full amount of correction before writing to the parameters file.

What I want instead is to store the head to camera matrix before correction or less preferably the original correction value we added.
It is simply encapsulation of all the information instead of sourcing it from different places and build back.

oleflb:
I now also ran into this issue and now fully understand your pain :D I still think having an uncalibrated and a calibrated variant is the best solution to this problem. Did you implement something like this already?

tuxbotix:
I partly did it but didn't finish it, but should be quite easy to do, so if
you or someone else have time please go ahead 😅.

oleflb:
This is implemented now in #1690. Please close if that PR suits your use case as well

tuxbotix:
Oh neat, that's the alternative I wanted to implement. So all good 😋

PR #1657, opened 2025-03-02
Power-Panic Service

paudar:
## Why? What?

Implementation of an service, that provides auditive Feedback when the battery of the NAO runs out.

## ToDo / Known Issues

The service still needs to be added to the meta-nao repository so that it becomes part of the image.
You may can argue about the time-interval of the sound. Right now it is played every charge in % in seconds.

In the next iteration you think of a possibility to mute the alarm. 

## How to Test

After creating the binary of the service with `./pepsi build power-panic`
you will find the binary in `~/worktree/hulk/services/power-panic/target/debug`.
Copy the binary to a NAO e. g. using `rsync power-panic nao@10.1.24.x:~/hulk/bin/ `
and execute it by connecting to the robot e. g. with `./pepsi shell x`


services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
+use color_eyre::eyre::{eyre, Result, WrapErr};
+use zbus::{proxy, zvariant::Optional};
+
+use hula_types::Battery;
+
+use rodio::{source::Buffered, Decoder, OutputStream, Sink, Source};
+use std::io::Cursor;
```
oleflb, line 7: please sort the import blocks.
First block are std imports, next block are external crates (color_eyre, zbus, ...) and last block are hulk dependencies (hula_types)

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+
+const AUDIO_FILE: &[u8] = include_bytes!("../sound/water-drop.mp3");
+
+#[proxy(
+    default_service = "org.hulks.hula",
+    interface = "org.hulks.hula",
+    default_path = "/org/hulks/HuLA"
```
oleflb, line 14: Is it correct that it says "hula" here?

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn battery(&self) -> zbus::Result<Optional<Battery>>;
+}
+
+struct BatteryInfo {
+    proxy: BatteryInfoProxy<'static>,
+}
+
+impl BatteryInfo {
+    pub async fn initialize() -> Result<Self> {
+        let connection = zbus::Connection::system().await?;
```
oleflb, line 26: whenever the name is unique, try to to import instead of qualifying.

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+async fn main() -> Result<()> {
+    let audio_player = AudioPlayer::new(AUDIO_FILE);
+    loop {
+        let battery = get_battery_info().await?;
+        println!("Battery: {:?}", battery);
+
+        let mut time_to_sleep = 60;
+
+        // ceck if battery is low and is not charging
+        if battery.charge < 0.20 && battery.current < 0.0 {
```
oleflb, line 82: instead of using a comment here, you could extract the checks into two boolean variables. So you'd have `let battery_is_low = ..` and `let is_charging = ..`.
This way the `if` condition becomes nicer to read and you don't need a comment

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+            println!("Battery low, playing sound");
+            audio_player.play();
+            time_to_sleep = (battery.charge * 100.0) as u64;
+        }
+        std::thread::sleep(std::time::Duration::from_secs(time_to_sleep));
```
oleflb, line 87: use imports instead of qualifying paths

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    let audio_player = AudioPlayer::new(AUDIO_FILE);
+    loop {
+        let battery = get_battery_info().await?;
+        println!("Battery: {:?}", battery);
+
+        let mut time_to_sleep = 60;
+
+        // ceck if battery is low and is not charging
+        if battery.charge < 0.20 && battery.current < 0.0 {
+            println!("Battery low, playing sound");
```
oleflb, line 83: is this println! necessary?

oleflb (changes requested):
Just some minor remarks, nice work :+1:

services/power-panic/src/main.rs
paudar: No, it's not.

services/power-panic/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let audio_player = AudioPlayer::new(AUDIO_FILE);
+    loop {
+        let battery = get_battery_info().await?;
+        println!("Battery: {:?}", battery);
+
+        let mut time_to_sleep = 60;
```
oleflb, line 81: Please try to use sophisticated types, in this case `Duration` for representing time. When coding your time as an integer, it is not clear whether you mean seconds, minutes, etc. . You can read more about it here: https://refactoring.guru/smells/primitive-obsession

oleflb (changes requested):
Looks really good, just one more request :)

services/power-panic/src/main.rs
paudar: Good point. I've changed the type of the variable `time_to_sleep`.

oleflb (approved):

oleflb:
@paudar please enable merge when ready

PR #1655, opened 2025-02-26
I am speed

oleflb:
## Why? What?

Disables all of the unneeded features to **drastically** /s speed up compilation (and linking) of `mio`.

Benchmark on Compileknecht:
Before:
real	1m12.175s
user	25m57.557s
sys	1m46.786s

After:
real	1m6.043s
user	18m50.101s
sys	1m35.691s

Fixes #

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

* Let it compile

PR #1658, opened 2025-03-02
Optional Shadows for mio

oleflb:
## Why? What?

Based on #1655 

Because my laptop is a :carrot:, rendering with shadows enabled results in around 20fps. 
Therefore I added a toggle to disable shadows which increases fps to 60.
Also adds a fps indicator in the top right corner.
This is my first experience using bevy so I'd love some feedback whether this implementation is idiomatic.

Fixes #

## ToDo / Known Issues

## How to Test

* Run `mio` and check if toggling shadows works and if it increases fps/responsiveness


tools/mio/src/fps.rs
```diff
@@ -0,0 +1,35 @@
...
+            .add_systems(Startup, setup_fps_display)
+            .add_systems(Update, update_fps);
+    }
+}
+#[derive(Component)]
+struct FpsText;
+
+fn setup_fps_display(mut commands: Commands) {
+    commands.spawn((
+        bevy::prelude::Text::new("FPS:"),
```
schmidma, line 18:
```suggestion
        Text::new("FPS:"),
```

schmidma (changes requested):
one small question

Cargo.toml
```diff
@@ -233,6 +233,9 @@ serde = { git = "https://github.com/HULKs/serde.git", rev = "2e5e545dc9958ab39d7
 # Pinned to forked serde version since https://github.com/serde-rs/serde/pull/2513 is not merged
 serde_derive = { git = "https://github.com/HULKs/serde.git", rev = "2e5e545dc9958ab39d7b83120c16f15d864ea8a6" }
 
+[profile.dev.package.bevy]
+opt-level = 3
```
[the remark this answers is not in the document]
oleflb: yes, removed

tools/mio/src/fps.rs
oleflb: done

tools/mio/Cargo.toml
```diff
@@ -6,13 +6,14 @@ license.workspace = true
 homepage.workspace = true
 
 [dependencies]
-bevy = { workspace = true, features = [
+bevy = { version = "0.15.2", features = [
```
schmidma, line 9: why do you manually specify the version here and not use the one specified by the workspace?

tools/mio/src/shadow_toggle.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[derive(Resource)]
+pub struct ShadowState {
+    pub enabled: bool,
+}
+
+impl Plugin for ShadowTogglePlugin {
+    fn build(&self, app: &mut App) {
+        app.add_systems(Update, toggle_shadows)
+            .insert_resource(ShadowState { enabled: true });
```
schmidma, line 13: Is there a conflict of specifying the default to `true` and in the `main.rs` specify the shadows of the directional light to `false` by default?

tools/mio/Cargo.toml
oleflb: now back to workspace

tools/mio/src/shadow_toggle.rs
oleflb: Now, since I check [`is_changed`](https://docs.rs/bevy/latest/bevy/prelude/struct.Res.html#method.is_changed). TLDR: resources which are added also count as being changed

schmidma (approved):

PR #1672, opened 2025-03-07
Wait for NAOs during pre- and postgame

julianschuler:
## Why? What?

This PR implements waiting for NAOs to become available via the network.

## ToDo / Known Issues

None.

None.

## How to Test

```sh
pepsi pregame
pepsi postgame first-half
```

Pre- and postgame will now wait until the NAOs specified in the `deploy.toml` become available instead of throwing an error.

### DIFF
--- a/crates/nao/src/lib.rs
+++ b/crates/nao/src/lib.rs
@@ -55,6 +55,14 @@ impl Nao {
         Self { address }
     }
 
+    pub async fn ping_until_available(host: Ipv4Addr) -> Self {
+        loop {
+            if let Ok(nao) = Self::try_new_with_ping(host).await {
+                return nao;
+            }
+        }
+    }
+
     pub async fn try_new_with_ping(host: Ipv4Addr) -> Result<Self> {
         Self::try_new_with_ping_and_arguments(host, PING_TIMEOUT).await
     }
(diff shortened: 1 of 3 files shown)


crates/nao/src/lib.rs
```diff
@@ -55,6 +55,14 @@ impl Nao {
         Self { address }
     }
 
+    pub async fn ping_until_available(host: Ipv4Addr) -> Self {
+        loop {
+            if let Ok(nao) = Self::try_new_with_ping(host).await {
+                return nao;
+            }
```
oleflb, line 62: maybe add a small sleep (1 second) here? :D

oleflb: Another idea, just use `Self::try_new_with_ping_and_arguments` with `Duration::MAX` (or 20 minutes or something)

oleflb (changes requested):
Nice idea!

oleflb:
tested, works great

crates/nao/src/lib.rs
julianschuler: Why would you want to add a delay? The timeout itself is already rate limiting the number of ping packets sent (currently one per iteration due to the `-c 1` flag). Similarly, I don`t think the second idea works, since only one ping packet is send, which a later connected NAO won't receive.

oleflb (approved):

PR #1704, opened 2025-03-13
Field color labeling and decision tree

Franziska-Sophie:
## Why? What?

New pipeline for automatic field color calibration:
- Start the python script
- Select the images you want to label
- Label one image after another with the build-in labeling tool (the label masks will be cached in case you want to use them again)
- After the last image was labeled, the decision trees (one for top and bottom camera) is trained
- The tree is converted to if-else statements in rust where the code is printed to the console and stored in a .rs file 

## How to Test
- Run the script to get the decision tree
- Use twix to check whether the image segmentation looks good


tools/field_color_detection/.gitignore
```diff
@@ -0,0 +1 @@
+*.dot
```
oleflb, line 1: you are missing an empty new line

tools/field_color_detection/src/decision_tree/__init__.py
```diff
@@ -0,0 +1 @@
+from .generate_tree import optimize_thresholds, predict_image, save_tree_as_dot_file
```
oleflb, line 1: same here, missing empty new line. Just run a `uv run ruff format` in the python workspace

tools/field_color_detection/src/decision_tree/generate_tree.py
```diff
@@ -0,0 +1,129 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class Feature:
+    identifier: str
+    data_type: str
+
+
+def convert_pixels_BGR2bgrI(pixel_BGR):
```
oleflb, line 13: please add type info to parameters and return type

tools/field_color_detection/src/decision_tree/generate_tree.py
```diff
@@ -0,0 +1,129 @@
...
+    pixel_sums = np.sum(pixel_BGR, axis=-1, keepdims=True)
+    pixel_sums[pixel_sums == 0] = 1
+    chromaticity = pixel_BGR / pixel_sums
+    intensity = np.array(pixel_sums / 3).astype(np.uint8)
+    image_bgI = np.concatenate((chromaticity, intensity), axis=-1)
+    return image_bgI
+
+
+def optimize_thresholds(pixels_BGR, pixels_YCrCb, y, camera):
```
oleflb, line 22: same here. Tip: use `Literal["top", "bottom"]` for `camera` (or define an enum)

oleflb (changes requested):
Would be nice to add types to all of your functions :)

oleflb (dismissed):

PR #1721, opened 2025-03-13
Reduces Step planner size

MeiiMoon:
## Why? What?

Reduces the size of the step planner. So that it walks with more stability.

## How to Test

Compare with the bigger step size on twix.


schluis (dismissed):
Perfect

oleflb:
@MeiiMoon @schluis the CI isn't happy

schluis:
Not needed anymore, thank's a lot nonetheless

PR #1725, opened 2025-03-13
Weighted residuals for calibration

tuxbotix:
## Why? What?

The residual calculation is done in Ground plane and affected by the pixel-noise (+-0.5px position). While this variance is constant in Pixel plane, it varies on Ground plane once projected depending on Pixel position.

This PR introduces weights based on the variance and successfully reduced/ eliminated bias caused on least square optimisation

Fixes #

## ToDo / Known Issues

## How to Test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2169,8 +2169,10 @@ checksum = "2d2c12f985c78475a6b8d629afd0c360260ef34cfef52efccdcfd31972f81c2e"
 name = "calibration"
 version = "0.1.0"
 dependencies = [
+ "approx",
  "coordinate_systems",
  "geometry",
+ "itertools 0.14.0",
  "levenberg-marquardt",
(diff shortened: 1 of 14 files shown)


oleflb:
why don't you simply do the optimization in pixel space?

pejotejo:
@tuxbotix what is the state of this pr

tuxbotix:
Closing this PR [maybe for now] as weighting didn't help as much as I've expected. Instead letting the radius to vary (compensate) helped - maybe the proper fix is to let intrinsics also to be tuned at the same time.

@oleflb optimising in pixel space would be nice and avoid all these headaches, but I didn't find a good enough way to fit detected points to a projected circle (ellipse like thingy). Geometric methods looked rather complicated to try. If someone can propose a good way or help me with this, I can integrate it and test

PR #1731, opened 2025-03-13
absolute hypothesis validity reset

oleflb:
## Why? What?

Decrease number of cycles where there is no ball hypothesis.

Fixes #

## ToDo / Known Issues

- Ball Filter → 🗑️

## How to Test

- check single unexpected percepts don't throw of the hypothesis


oleflb:
not so nice. for reference see first half against SPQR 2025-03-14

PR #1739, opened 2025-03-14
CloseAll panels in Twix

schmidma:
Add CloseAll binding to close all tabs
Focus initial leaf when starting twix


knoellle (approved):
Wörks 👍🏻

Issue #1676, opened 2025-03-09
Decay ball search heatmap based on viewing direction


oleflb:
Currently, the ball search heatmap is very basic, since it decays all cells independent of each other and the state of the Nao. An improvement would be to cooldown cells in the viewing cone of the Nao up to a certain distance faster than others. This would also fix our Naos from running into one corner during corner kick and not checking the other one (related to #1288)

knoellle:
I'm afraid just decaying the values where we don't see a ball won't help in the corner kick case.
The current search suggestor simply overwrites the heatmap values in both corners with a static 1.0 every cycle during the corner kick.
The correct way would be to only set these values when entering the corner kick state.

PR #1741, opened 2025-03-14
heatmap diffusion convolution confusion

pejotejo:
## Why? What?

changes heatmap decay in single cells to diffusion.  

decay based on viewing direction (#1676)

## How to Test
Behavior Sim
or let robot play and look at heatmap

### DIFF
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -128,7 +128,7 @@ impl SearchSuggestor {
             self.heatmap
                 .map
                 .conv(&kernel, ConvMode::Same, PaddingMode::Replicate)?;
-        self.heatmap.map /= self.heatmap.map.sum();
+        self.heatmap.map /= self.heatmap.map.sum() + f32::EPSILON;
         Ok(())
     }
 }
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -7,7 +7,8 @@ use framework::{AdditionalOutput, MainOutput};
 use itertools::Itertools;
 use linear_algebra::{point, Isometry2, Point2};
 use nalgebra::clamp;
-use ndarray::Array2;
+use ndarray::{array, Array2};
+use ndarray_conv::{ConvExt, ConvMode, PaddingMode};
 use serde::{Deserialize, Serialize};
 use spl_network_messages::{SubState, Team};
 use types::{
@@ -70,7 +71,7 @@ impl SearchSuggestor {
     }
 
     pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
-        self.update_heatmap(&context);
+        let _ = self.update_heatmap(&context);
         let suggested_search_position = self
             .heatmap
             .get_maximum_position(context.search_suggestor_configuration.minimum_validity);
@@ -84,7 +85,7 @@ impl SearchSuggestor {
         })
     }
 
-    fn update_heatmap(&mut self, context: &CycleContext) {
+    fn update_heatmap(&mut self, context: &CycleContext) -> Result<()> {
         if let Some(ball_position) = context.ball_position {
             if let Some(ground_to_field) = context.ground_to_field {
                 self.heatmap[ground_to_field * ball_position.position] = 1.0;
@@ -107,7 +108,15 @@ impl SearchSuggestor {
             }
         }
 
-        self.heatmap.map *= 1.0 - context.search_suggestor_configuration.heatmap_decay_factor;
+        let alpha: f32 = 1.0 / 9.0;
+        let kernel: Array2<f32> =
+            1.0 / (8.0 + alpha) * array![[1.0, 1.0, 1.0], [1.0, alpha, 1.0], [1.0, 1.0, 1.0]];
+
+        self.heatmap.map =
+            self.heatmap
+                .map
+                .conv(&kernel, ConvMode::Full, PaddingMode::Replicate)?;
+        Ok(())
     }
 }
 
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -63,7 +63,8 @@ impl SearchSuggestor {
                 .round() as usize,
         );
         let heatmap = Heatmap {
-            map: Array2::zeros((heatmap_length, heatmap_width)),
+            map: Array2::ones((heatmap_length, heatmap_width))
+                / (heatmap_length * heatmap_width) as f32,
             field_dimensions: *context.field_dimensions,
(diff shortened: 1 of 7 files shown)


crates/control/src/search_suggestor.rs
```diff
@@ -70,7 +71,7 @@ impl SearchSuggestor {
     }
 
     pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
-        self.update_heatmap(&context);
+        let _ = self.update_heatmap(&context);
```
oleflb, line 74: why do you choose to silently ignore the error?

crates/control/src/search_suggestor.rs
```diff
@@ -107,7 +108,28 @@ impl SearchSuggestor {
             }
         }
 
-        self.heatmap.map *= 1.0 - context.search_suggestor_configuration.heatmap_decay_factor;
+        let kernel: Array2<f32> = 1.0
```
oleflb, line 111: extract kernel creation to new function

crates/control/src/search_suggestor.rs
```diff
@@ -107,7 +108,28 @@ impl SearchSuggestor {
...
+                        .search_suggestor_configuration
+                        .heatmap_convolution_kernel_weight,
+                    1.0
+                ],
+                [1.0, 1.0, 1.0]
+            ];
+        self.heatmap.map =
+            self.heatmap
+                .map
+                .conv(&kernel, ConvMode::Same, PaddingMode::Replicate)?;
```
oleflb, line 130: are you sure `PaddingMode::Zero` isn't more appropriate here?

crates/control/src/search_suggestor.rs
```diff
@@ -107,7 +108,28 @@ impl SearchSuggestor {
...
+        self.heatmap.map /= self.heatmap.map.sum() + f32::EPSILON;
```
oleflb, line 131: instead of adding `f32::EPSILON`, initialize the heatmap with `1.0 / number_of_grid_cells`. Then the heatmap can never sum up to 0.0

crates/control/src/search_suggestor.rs
```diff
@@ -107,7 +108,28 @@ impl SearchSuggestor {
...
+            * array![
+                [1.0, 1.0, 1.0],
+                [
+                    1.0,
+                    context
+                        .search_suggestor_configuration
+                        .heatmap_convolution_kernel_weight,
+                    1.0
+                ],
+                [1.0, 1.0, 1.0]
```
oleflb, line 125: I find it much more intuitive to have the kernel be `[[a, a, a], [a, 1-a, a], [a, a, a]]` with a in [0, 1]

oleflb (changes requested):

crates/control/src/search_suggestor.rs
pejotejo: done

crates/control/src/search_suggestor.rs
pejotejo: done

crates/control/src/search_suggestor.rs
pejotejo: no, we talked about this on Go

crates/control/src/search_suggestor.rs
pejotejo: done

tools/twix/src/panels/map/layers/ball_search_heatmap.rs
```diff
@@ -50,8 +50,14 @@ impl Layer<Field> for BallSearchHeatmap {
...
                     second_point,
-                    Color32::from_rgba_unmultiplied(0, 0, 255, ((value * 255.0) / 2.0) as u8),
+                    Color32::from_rgba_unmultiplied(
+                        0,
+                        0,
+                        255,
+                        (value.powf(1.2) * 255.0 * 3.0) as u8,
+                    ),
                 );
+                dbg!(value);
```
oleflb, line 60: please remove dbg!

tools/twix/src/panels/map/layers/ball_search_heatmap.rs
```diff
@@ -50,8 +50,14 @@ impl Layer<Field> for BallSearchHeatmap {
                 painter.rect_filled(
                     first_point,
                     second_point,
-                    Color32::from_rgba_unmultiplied(0, 0, 255, ((value * 255.0) / 2.0) as u8),
+                    Color32::from_rgba_unmultiplied(
+                        0,
+                        0,
+                        255,
+                        (value.powf(1.2) * 255.0 * 3.0) as u8,
```
oleflb, line 57: maybe extract into variable so that you can assign a name to this "magic" construct

etc/parameters/default.json
```diff
@@ -1454,8 +1454,8 @@
   },
   "search_suggestor": {
     "cells_per_meter": 2.0,
-    "heatmap_decay_factor": 0.002,
-    "minimum_validity": 0.01
+    "heatmap_convolution_kernel_weight": 0.001007049,
```
oleflb, line 1457: oddly specific. How did you find this number?

oleflb (changes requested):
heatmap implementation looks good, just some remarks on the twix side

tools/twix/src/panels/map/layers/ball_search_heatmap.rs
pejotejo: done

tools/twix/src/panels/map/layers/ball_search_heatmap.rs
pejotejo: done

etc/parameters/default.json
pejotejo: I took the weight of the old kernel from the C++ days and converted it, so that the scale should be the same now

crates/control/src/search_suggestor.rs
```diff
@@ -106,11 +108,30 @@ impl SearchSuggestor {
...
             }
         }
-
-        self.heatmap.map *= 1.0 - context.search_suggestor_configuration.heatmap_decay_factor;
+        let kernel = create_kernel(
+            context
+                .search_suggestor_configuration
+                .heatmap_convolution_kernel_weight,
+        );
+        if let Ok(convolved_heatmap) =
```
oleflb, line 116: When does the convolution return an error? I can imagine this happens only for when the kernel is larger than the heatmap. Right now you are silently ignoring the error. If for some reason this call would only generate Errors, we would not notice it. I though of using `.wrap_err("heatmap convolution failed")?`

pejotejo: this should be better now

oleflb (approved):

PR #1759, opened 2025-03-14
Referee silence interval

schluis:
## Why? What?

- Referee detection caused a lot of sent messages
- Now it's filtered
- Removes the sacrificial lamb 

- Implement a global message filter! 🥳 

## How to Test

- Look at the code

### DIFF
--- a/crates/control/src/role_assignment.rs
+++ b/crates/control/src/role_assignment.rs
@@ -789,6 +789,7 @@ mod test {
             striker_trusts_team_ball_duration in  Just(Duration::from_secs(5)),
             optional_roles in Just(&[Role::DefenderLeft, Role::StrikerSupporter])
         ) {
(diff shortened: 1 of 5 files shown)


crates/control/src/role_assignment.rs
```diff
@@ -472,17 +498,6 @@ fn is_cooldown_elapsed(now: SystemTime, last: Option<SystemTime>, cooldown: Dura
     }
 }
 
-fn is_enough_message_budget_left(context: &CycleContext<impl NetworkInterface>) -> bool {
```
ThagonDuarte, line 475: Why did you remove this function. This now introduces code duplication. I'd suggest simply leaving the function as is.

ThagonDuarte (changes requested):
One comment, otherwise looks good

ThagonDuarte (approved):
LGTM

PR #1763, opened 2025-03-14
Stay Penalized while no ground contact

schmidma:
Often, GCCs send robots back to the game, while they are still in the air. They should unpenalize as soon as they have ground contact, to ensure behavior and localization track the start correctly


oleflb (commented):
Looks good, let's see this in action tomorrow

oleflb (approved):
works like a charm

PR #1764, opened 2025-03-14
Do not use line measurements while getting up

### LINKED ISSUE #1758: Ignore detections while getting up
We should ignore most measurements (ball?, lines?, robots?) while getting up. During this phase, calibration is weirdly off, and introduces false measurements.

schmidma:
Fixes #1758


oleflb:
tested: works good but maybe extend this to falling?

oleflb (approved):

PR #1773, opened 2025-03-15
Disable realtime for hulks_webots explicitly

### LINKED ISSUE #1651: Feature flags are absorbed from other projects when using Pepsi run (since Pepsi Max)
I came across this while trying to run webots -> `pepsi run --bin hulk_webots` and got an error saying realtime scheduler cannot be set. This issue and changed command came after pepsi max.

Looking deeper, the root cause is that the features set for `hulk_nao` leaked into `hulk_webots` when calling pepsi run.

It is also the same behavious by `cargo run --bin hulk_webots`. It seems using cargo run in this way inspects the workspace file and reads feature flags used by everyone and use the combined result (unified?).
So I think the pepsi max PR isn't really at fault, rather how cargo run works?

Old pepsi run uses `cargo run __manifest_path=... toml` (I forgot the exact wording). This seems to read only the needed manifest, so features set for other executables aren't leaked into the build.

tuxbotix:
## Why? What?
This combines two feature flags to actually activate realtime. I don't like it, but this is the easy-way out without digging depe into pepsi or cargo.

Fixes #1651 

## ToDo / Known Issues

There should be a better way than this

## How to Test

The following should run (while webots is active) without the "failed to set scheduler" error in #1651.

```bash
pepsi run --bin hulks_webots
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,7 +55,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f47983a1084940ba9a39c077a8c63e55c619388be5476ac04c804cfbd1e63459"
 dependencies = [
  "accesskit",
- "hashbrown 0.15.2",
+ "hashbrown 0.15.3",
  "immutable-chunkmap",
 ]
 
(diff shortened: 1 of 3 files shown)


paudar:
The error no longer occurs.
Maybe you could give the feature a more self explaining name like `is_webot`

tuxbotix:
I'm a bit confused, did you mean with this PR the error is fixed or did I misunderstood? :D

As for naming, I agree, it should be better.

tuxbotix:
Cargo lock issues fixed :)

PR #1776, opened 2025-03-15
Default Field Dimensions in Mio

schmidma:
Mio has SMD field dimensions. This PR adjusts the constants to default spl field dimensions.

Any actual reading of parameters is for future iterations.


knoellle (approved):
LGTM

PR #1782, opened 2025-03-15
Pushing kicking team filtering

ThagonDuarte:
## Why? What?

Locally infer `kicking_team` during `SubState::PushingFreeKick`. 
Also adds a bevyhaviour simulator scenario `kicking_team_filtering.rs` to test the inference for all sub states where local inference is possible, i.e. `GoalKick`, `CornerKick`, `PenaltyKick` and `PushingFreeKick`.

Depends on #1789 

## ToDo / Known Issues

- One interesting thing to mention, is that during the scenario playback, starting from tick ~6700, player number six stands directly next to the ball during an opponents goal kick. This is an illegal position and should be investigated. Issue tracking this is #1820.

## How to Test

- Trigger a `PushingFreeKick` in game and see if the behavior is correct when both we and the opponents are kicking team.


crates/bevyhavior_simulator/src/autoref.rs
```diff
@@ -226,6 +226,19 @@ pub fn auto_assistant_referee(
                 | Penalty::PlayerStance { .. }
                 | Penalty::Substitute { .. }
                 | Penalty::Manual { .. } => {
+                    if team == Team::Hulks {
+                        if let Some(mut robot) = robots
+                            .iter_mut()
+                            .find(|robot| robot.parameters.player_number == player_number)
+                        {
+                            *robot.ground_to_field_mut() =
+                                Isometry2::from_parts(vector![-3.2, -3.3], FRAC_PI_2);
```
oleflb, line 235: magic values?

crates/bevyhavior_simulator/src/bin/kicking_team_filtering.rs
```diff
@@ -0,0 +1,279 @@
...
+                Some(PlayerNumber::Six)
+            } else {
+                None
+            };
+
+        game_controller_commands.send(GameControllerCommand::SetSubState(
+            Some(checked_sub_state),
+            correct_kicking_team,
+            penalized_player_number,
+        ));
```
oleflb, line 250: I think this is unexpected in a function called sth. with "check". What do you think of extracting the sending part to a different function and then calling that function at the call site of this "check" function?

crates/bevyhavior_simulator/src/game_controller.rs
```diff
@@ -62,19 +62,46 @@ fn game_controller_controller(
...
+                        if sub_state == Some(SubState::PenaltyKick) {
+                            game_controller.state.kicking_team = Some(team);
+                            game_controller.state.game_state = GameState::Ready;
+                        } else {
+                            game_controller.state.kicking_team = None;
+                        }
+                        match team {
+                            Team::Hulks => {
+                                game_controller.state.opponent_penalties[penalized_player_number
+                                    .expect("This sub state requires a penalized player number.")] =
```
oleflb, line 88: please make error messages lower case for consistency

crates/bevyhavior_simulator/src/game_controller.rs
```diff
@@ -62,19 +62,46 @@ fn game_controller_controller(
...
+                                    Some(Penalty::PlayerPushing {
+                                        remaining: Duration::from_secs(45),
+                                    })
+                            }
+                            Team::Opponent => {
+                                game_controller.state.penalties[penalized_player_number
+                                    .expect("This sub state requires a penalized player number.")] =
```
oleflb, line 95: same here

crates/bevyhavior_simulator/src/game_controller.rs
```diff
@@ -62,19 +62,46 @@ fn game_controller_controller(
...
+                            game_controller.state.game_state = GameState::Ready;
+                        } else {
+                            game_controller.state.kicking_team = None;
+                        }
+                        match team {
+                            Team::Hulks => {
+                                game_controller.state.opponent_penalties[penalized_player_number
+                                    .expect("This sub state requires a penalized player number.")] =
+                                    Some(Penalty::PlayerPushing {
+                                        remaining: Duration::from_secs(45),
```
oleflb, line 90: Maybe we should put the specific durations at a separate location (preferably a resource), so the user could change them more easily

etc/parameters/default.json
```diff
@@ -1374,7 +1371,9 @@
       "secs": 4
     },
     "distance_to_consider_ball_moved_in_kick_off": 0.3,
-    "whistle_acceptance_goal_distance": [0.5, 0.5]
+    "whistle_acceptance_goal_distance": [0.5, 0.5],
+    "duration_to_keep_observed_ball": { "nanos": 0, "secs": 14 },
+    "duration_to_keep_new_penalties": { "nanos": 0, "secs": 30 }
```
oleflb, line 1376: isn't that a really long time?

oleflb (changes requested):

crates/bevyhavior_simulator/src/autoref.rs
ThagonDuarte: This is one of the penalized positions at the penalty spot.

etc/parameters/default.json
ThagonDuarte: It is. But in the case that we are kicking team, because an enemy has been penalized, this duration determines how long we will believe we are kicking team. If this is not the entire duration of the sub state, we will stop believing we are kicking team after the duration has run out, if we didnt kick it before then.

crates/bevyhavior_simulator/src/game_controller.rs
ThagonDuarte: I added the duration as a `const` in the file. I'm not sure adding a resource is worth it, since this is the only such rule `Duration` we use atm.

crates/bevyhavior_simulator/src/game_controller.rs
ThagonDuarte: Done

crates/bevyhavior_simulator/src/autoref.rs
ThagonDuarte: Moved to a variable, giving it a name

oleflb (approved):
lgtm

PR #1790, opened 2025-03-15
Always execute kicks to the end

schmidma:
If we requested kicks once, walking is going to execute it to the end. Stepping all steps in the kick, before transitioning to a different mode.

... hopefully fixes penalty shootout


schluis (approved):
Top

PR #1805, opened 2025-03-16
Cd to repo root before paramalarm

knoellle:
## Why? What?

What it says. previously crashed because it was unable to find the json files.

Fixes #

## ToDo / Known Issues

## How to Test

Cd somewhere other than repo root and run parameter tester


tools/parameter_tester/src/main.rs
```diff
@@ -7,16 +11,20 @@ use color_eyre::{
...
 
 fn main() -> Result<()> {
-    let framework_parameters_path = args()
-        .nth(1)
-        .unwrap_or("etc/parameters/framework.json".to_string());
+    let repository_search_path = match args().nth(1) {
+        Some(path) => PathBuf::from(path),
+        None => PathBuf::from("."),
+    };
+    let repository = Repository::find_root(repository_search_path).expect("no repository found");
```
julianschuler, line 23: Why do use `expect` here instead of e.g. `wrap_err` and propagating the error?

tools/parameter_tester/src/main.rs
```diff
@@ -7,16 +11,20 @@ use color_eyre::{
...
+    set_current_dir(repository.root).expect("failed to cd to repo root");
 
-    let file =
-        File::open(framework_parameters_path).wrap_err("failed to open framework parameters")?;
+    let file = File::open("etc/parameters/framework.json")
```
julianschuler, line 26: What do you think about joining this path with the repository root instead of first changing the current directory and then using a relative path?

julianschuler (changes requested):

tools/parameter_tester/src/main.rs
knoellle: Doesn't really matter here since we are in the main function, but I changed it anyways for the sake of consistency.

tools/parameter_tester/src/main.rs
knoellle: The problem is that the framework.json contains the path to the actual configuration files that is relative to the repository root.
We'd have to then also prefix `framework_parameters.parameters_directory`, but only if it is not an absolute path.
I think for this simple program, just changing the working directory is the better choice.

julianschuler (approved):
LGTM

PR #1813, opened 2025-03-24
Foot Leveling restructuring

schmidma:
## Why? What?

only some explanation and restructuring

### DIFF
--- a/crates/walking_engine/src/foot_leveling.rs
+++ b/crates/walking_engine/src/foot_leveling.rs
@@ -27,10 +27,8 @@ impl FootLeveling {
         // The default torso rotation represents the desired, slightly leaned forward/backward configuration
         let default_torso_rotation = context.robot_to_walk.rotation();
 
-        // The current sensed orientation of the robot
         let current_orientation = context.robot_orientation;
 
-        // This difference (leveling_error) represents the misalignment that we need to correct for foot leveling
         let leveling_error = current_orientation.inner * default_torso_rotation.inner.inverse();
         let (roll_angle, pitch_angle, _) = leveling_error.euler_angles();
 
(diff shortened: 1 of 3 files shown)


crates/walking_engine/src/foot_leveling.rs
```diff
@@ -22,36 +22,43 @@ pub struct FootLeveling {
...
-                1.0
-            } else {
-                1.0 - normalized_time_since_start
-            };
+        // The default torso rotation represents the desired, slightly leaned forward/backward configuration
+        let default_torso_rotation = context.robot_to_walk.rotation();
 
-        let pitch_base_factor = if level_pitch > 0.0 {
-            context.parameters.pitch_positive_level_factor
+        // The current sensed orientation of the robot
```
oleflb, line 30: I think this comment does not help understand the code line better because it is very self-explanatory already

crates/walking_engine/src/foot_leveling.rs
```diff
@@ -22,36 +22,43 @@ pub struct FootLeveling {
...
+        let current_orientation = context.robot_orientation;
+
+        // This difference (leveling_error) represents the misalignment that we need to correct for foot leveling
```
oleflb, line 33: Same here: If the variable is called something with error, I already expect it to be the difference of actual and target

oleflb (changes requested):

crates/walking_engine/src/foot_leveling.rs
schmidma: removed

crates/walking_engine/src/foot_leveling.rs
schmidma: removed

oleflb (approved):

PR #1817, opened 2025-03-25
Dont serialize motion interpolators

oleflb:
## Why? What?

We serialized the non-changing interpolators in every control cycle, which is unecessary.
This PR extracts the mutable state from the interpolators and removes the serialization of interpolators in replay. Since we have a source distribution now, no information is lost.

Fixes #

## ToDo / Known Issues

## How to Test

- Check if motions work as before

### DIFF
--- a/crates/control/src/motion/arms_up_squat.rs
+++ b/crates/control/src/motion/arms_up_squat.rs
@@ -1,5 +1,6 @@
 use color_eyre::Result;
 use context_attribute::context;
+use framework::deserialize_not_implemented;
 use framework::MainOutput;
 use hardware::PathsInterface;
 use motionfile::{InterpolatorState, MotionFile, MotionInterpolator};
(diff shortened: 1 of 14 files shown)


knoellle:
When benchmarking this against main, I saw a ~75% reduction in the size of the `Control.bincode` file when accounting for the number of frames. 🎉

|Branch|Frames|File size|
|---------|----------|------------|
|Main|3482|94 MB|
|PR|3132|19 MB|

Considering on main a typical replay size breakdown looks as follows, this is a huge amount of space we're saving!
```
$ ls -lh --sort size
total 1.6G
-rw-r--r-- 1 mystery mystery 1012M Mar 13 16:38 Control.bincode
-rw-r--r-- 1 mystery mystery  268M Mar 13 16:38 VisionTop.bincode
-rw-r--r-- 1 mystery mystery  268M Mar 13 16:38 VisionBottom.bincode
-rw-r--r-- 1 mystery mystery  2.1M Mar 13 16:29 hulk.out
[more smaller files]
```

knoellle (approved):
LGTM

Issue #1822, opened 2025-04-03
Vision and Control cycles regularily take too long


ThagonDuarte:
When uploading current main, both control and vision regularly take too long, i.e. the nao DONKs. This happens both from time to time if no interaction with an Unstiff nao is done and specifically when interacting with the nao. This happened especially often when running ObjectDetectionTop. Here Vision basically always takes between ~35 and ~45 ms.

ThagonDuarte:
![Image](https://github.com/user-attachments/assets/7443231b-8309-4526-aebc-278e32fadaa2)

The left part of the plot are the cycle timings with #1399 reverted. The right part is current main with #1333.

#1333 definitely had some impact on this.

ThagonDuarte:
I investigated a bit further... I checked #1769 out, the PR where the individual cycler DONKing was introduced, and vision still regularly takes too long. This was merged right before the GO, so this state was present during the entire competition. However, I cannot recall, that the NAOs always and continuously DONKed during standby visual referee, as they are doing in my testing now...

PR #1823, opened 2025-04-03
Add individual and total node duration measurements

ThagonDuarte:
## Why? What?

Debugging #1822 was kind of annoying without proper timing measurements for node and cycle durations. 
This PR expands on #710 and adds the `cycle_timings` field to each cycler database. This includes a total time measurement from the start of the first node of a cycler until the last node's cycle function terminates, which is the same as the cycler recording timestamp. This can be used under `$cycler.cycle_timings.total`. 
Same as #710 did, this also adds node cycle durations for each individual node of a cycler. These can be accessed under `$cycler.cycle_timings.$node`

## ToDo / Known Issues

## How to Test

- Subscribe `$cycler.cycle_timings` in a twix text panel.

### DIFF
--- a/crates/code_generation/src/cyclers.rs
+++ b/crates/code_generation/src/cyclers.rs
@@ -916,7 +916,7 @@ fn generate_execute_node_and_write_main_outputs(
         {
             #[allow(clippy::needless_else)]
             if #are_required_inputs_some {
-                let cycle_start = std::time::SystemTime::now();
+                let node_cycle_start = std::time::SystemTime::now();
                 let main_outputs = {
                     let _task = ittapi::Task::begin(&itt_domain, #node_name);
                     self.#node_member.cycle(
@@ -926,8 +926,8 @@ fn generate_execute_node_and_write_main_outputs(
                     )
                     .wrap_err(#cycle_error_message)?
                 };
-                let cycle_duration = cycle_start.elapsed().expect("time ran backwards");
-                own_database.cycle_timings.#node_member = cycle_duration;
+                let node_cycle_duration = node_cycle_start.elapsed().expect("time ran backwards");
+                own_database.cycle_timings.#node_member = node_cycle_duration;
                 #write_main_outputs
             }
             else {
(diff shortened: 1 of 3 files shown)


oleflb:
@knoellle what is your take on this? Should we give this a go now?

knoellle:
Yes!

[7480wz-2569699538.jpg](https://github.com/user-attachments/assets/ad2e2ad8-652f-472f-8d69-5c5c71d95697)

crates/code_generation/src/cyclers.rs
```diff
@@ -910,6 +916,7 @@ fn generate_execute_node_and_write_main_outputs(
         {
             #[allow(clippy::needless_else)]
             if #are_required_inputs_some {
+                let cycle_start = std::time::SystemTime::now();
```
oleflb, line 919: I think you are confusing `cycle_start` with `node_start`, right?

crates/code_generation/src/cyclers.rs
```diff
@@ -919,6 +926,8 @@ fn generate_execute_node_and_write_main_outputs(
                     )
                     .wrap_err(#cycle_error_message)?
                 };
+                let cycle_duration = cycle_start.elapsed().expect("time ran backwards");
```
oleflb, line 929: same here

crates/source_analyzer/src/structs.rs
```diff
@@ -39,7 +40,38 @@ impl Structs {
...
+                        InsertionRule::AppendDataType {
+                            data_type: Type::Verbatim(quote! { std::time::Duration }),
+                        },
+                    ])
+                    .map_err(|source| Error::Hierarchy {
+                        node: node.name.clone(),
+                        cycler: cycler.name.clone(),
+                        source,
+                    })?;
+
```
oleflb, line 74: This code is duplicated (for total and then for all nodes). Can you use something like chain to have only one for loop?

oleflb (changes requested):
I think

crates/code_generation/src/cyclers.rs
ThagonDuarte: I mean, that is the start of the cycle function of that node :wink: But I renamed it to `node_cycle_start` and `node_cycle_duration` for clarity.

crates/code_generation/src/cyclers.rs
ThagonDuarte: See above

crates/source_analyzer/src/structs.rs
ThagonDuarte: I pulled the field insertion out of the for loop and added another with `iter::once().chain()`. Is this what you had in mind?

oleflb (commented):
LGTM, but need to test

oleflb (approved):
wörks

PR #1831, opened 2025-04-09
Teamball in heatmap

pejotejo:
## Why? What?

adds teamball to heatmap

## How to Test

bevyhavior sim

##
Parameters still need to be tuned

### DIFF
--- a/crates/control/src/search_suggestor.rs
+++ b/crates/control/src/search_suggestor.rs
@@ -36,6 +36,7 @@ pub struct CycleContext {
     field_dimensions: Parameter<FieldDimensions, "field_dimensions">,
 
     ball_position: Input<Option<BallPosition<Ground>>, "ball_position?">,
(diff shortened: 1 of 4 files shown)


crates/control/src/search_suggestor.rs
```diff
@@ -106,6 +107,9 @@ impl SearchSuggestor {
                 self.heatmap[rule_ball_hypothesis] = 1.0;
             }
         }
+        if let Some(team_ball) = context.team_ball {
+            self.heatmap[team_ball.position] = 1.0;
+        }
```
oleflb, line 112: What do you think of adding probability only when we get a message?

oleflb (changes requested):

crates/control/src/search_suggestor.rs
MaikRe: Has been changed

MaikRe (approved):
Looks good to me, we'll definitely need to watch out if the parameters need tuning

PR #1834, opened 2025-04-16
here we go again: behavior panel bug

pejotejo:
#1830 again after reset


MaikRe (approved):
Looks good to me now that I am actually allowed to approve it :wink:

PR #1858, opened 2025-05-30
Add negative pitch clamping

alexschmander:
## Why? What?

This is needed for the pose detection during kick in of #1646 . For the pose detection to work, the referee needs to be fully in frame of the camera. When closer to the referee, the pitch clamping causes the image of the top camera to not fully capture the referee.
Allows for head pitch clamping angles smaller than 0. The negative pitch clamp function is similar to the positive pitch clamping function but offset. 

Fixes #

## How to Test

- CI

### DIFF
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -1109,7 +1109,7 @@
     "outer_maximum_pitch": 0.0,
     "inner_maximum_pitch": 0.61,
     "outer_minimum_pitch": 0.0,
-    "inner_minimum_pitch": -0.41,
+    "inner_minimum_pitch": -0.43,
     "outer_yaw": 1.3,
     "injected_head_joints": null
   },


crates/control/src/motion/head_motion.rs
```diff
@@ -163,3 +151,65 @@ impl HeadMotion {
...
+            + interpolation_factor
+                * (head_motion_parameters.inner_maximum_pitch
+                    - head_motion_parameters.outer_maximum_pitch)
+    };
+
+    let minimum_pitch = if controlled_positions.yaw.abs() >= head_motion_parameters.outer_yaw {
+        head_motion_parameters.outer_minimum_pitch
+    } else {
+        let interpolation_factor =
+            0.5 * (PI * controlled_positions.yaw / head_motion_parameters.outer_yaw).cos();
```
rmburg, line 174: Just looking at the code, it looks like there is an `(1.0 + ...)` missing compared to the calculation of `maximum_pitch`. Is this intentional?

rmburg: It's just a bit confusing since in the code above the interpolation factor (as the name implies) is in the range [0..1], and thus `maximum_pitch` lies inside [`inner_maximum_pitch`..`outer_maximum_pitch`]. Without the `(1.0 + ...)`, `minimum_pitch` is not constrained to the range [`inner_minimum_pitch`..`outer_minimum_pitch`]! It feels like what you're trying to achieve is better done by adjusting the parameters (`inner_minimum_pitch`/`outer_minimum_pitch`)

ThagonDuarte: This now needs to be tested again, if the minimum pitch for yaw values where the ear of the Nao is closest to the shoulder is large enough.

ThagonDuarte: If that is not the case, the function could easily be scaled along the yaw-axis (x-axis in this graph).

rmburg: Nice. Let's test this the next time we have a nao at hand :)

ThagonDuarte: We tested with a game pad. Looks beautiful

PR #1862, opened 2025-06-03
stand up counter

vivi-enne:
## Why? What?

To avoid unnecessary stand up attempts when the robot cannot stand up any longer, this is a first step to count the number of stand up attempts.

Fixes #

## ToDo / Known Issues

## How to Test

Please test on a NAO, this is only tested on the replays from GO25 yet. I added an additional output to see in twix, it should increase when the robot makes a new stand up attempt (Dispatching -> Standup) and is reset when this does not hold true and no stand up is in progress.

### DIFF
--- a/crates/control/src/motion/motion_selector.rs
+++ b/crates/control/src/motion/motion_selector.rs
@@ -44,22 +44,24 @@ impl MotionSelector {
         let motion_safe_to_exit = context.motion_safe_exits[self.current_motion];
         let requested_motion = motion_type_from_command(context.motion_command);
 
-        if self.current_motion == MotionType::StandUpBack
+        self.stand_up_count = if self.current_motion == MotionType::StandUpBack
             || self.current_motion == MotionType::StandUpFront
             || self.current_motion == MotionType::StandUpSitting
         {
-            self.stand_up_count = self.stand_up_count;
+            self.stand_up_count
         } else if self.current_motion == MotionType::Dispatching
             && (requested_motion == MotionType::StandUpFront
                 || requested_motion == MotionType::StandUpBack
                 || requested_motion == MotionType::StandUpSitting)
         {
-            self.stand_up_count += 1;
+            self.stand_up_count + 1
         } else {
-            self.stand_up_count = 0;
-        }
+            0
+        };
 
-        context.stand_up_count.fill_if_subscribed(|| self.stand_up_count);
+        context
+            .stand_up_count
+            .fill_if_subscribed(|| self.stand_up_count);
 
         self.current_motion = transition_motion(
             self.current_motion,
--- a/crates/control/src/motion/motion_selector.rs
+++ b/crates/control/src/motion/motion_selector.rs
@@ -1,6 +1,6 @@
 use color_eyre::Result;
 use context_attribute::context;
-use framework::MainOutput;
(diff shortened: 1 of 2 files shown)


crates/control/src/motion/motion_selector.rs
```diff
@@ -34,13 +36,33 @@ impl MotionSelector {
...
+            self.stand_up_count
+        } else if self.current_motion == MotionType::Dispatching
+            && (requested_motion == MotionType::StandUpFront
+                || requested_motion == MotionType::StandUpBack
+                || requested_motion == MotionType::StandUpSitting)
+        {
+            self.stand_up_count + 1
+        } else {
+            0
+        };
```
oleflb, line 60: This code depends on the fact that we use ``MotionType::Dispatching` when we want to stand up. Furthermore, a `requested_motion` does not always become the `current_motion` (safe exits). Therefore I suggest using the last current motion and compare it to the current motion instead of using `requested_motion`.

crates/control/src/motion/motion_selector.rs
```diff
@@ -34,13 +36,33 @@ impl MotionSelector {
...
     }
 
-    pub fn cycle(&mut self, context: CycleContext) -> Result<MainOutputs> {
+    pub fn cycle(&mut self, mut context: CycleContext) -> Result<MainOutputs> {
         let motion_safe_to_exit = context.motion_safe_exits[self.current_motion];
         let requested_motion = motion_type_from_command(context.motion_command);
 
+        self.stand_up_count = if self.current_motion == MotionType::StandUpBack
+            || self.current_motion == MotionType::StandUpFront
+            || self.current_motion == MotionType::StandUpSitting
```
oleflb, line 49: You could add a method `MotionType::is_standup_motion(self) -> bool`. Then you could do something like `!self.current_motion.is_standup_motion() && self.current_motion.is_standup_motion()`. I think this would make the expression even easier to understand

oleflb: Then you could also use a match statement :)

oleflb (changes requested):

crates/control/src/motion/motion_selector.rs
vivi-enne: Changed that, had to make a new member variable though

crates/control/src/motion/motion_selector.rs
```diff
@@ -11,6 +11,8 @@ use types::{
 #[derive(Deserialize, Serialize)]
 pub struct MotionSelector {
     current_motion: MotionType,
+    last_motion: MotionType,
+    stand_up_count: i32,
```
oleflb, line 15: Can this be negative? If not opt for a u32 or similar instead :)

crates/control/src/motion/motion_selector.rs
```diff
@@ -11,6 +11,8 @@ use types::{
 #[derive(Deserialize, Serialize)]
 pub struct MotionSelector {
     current_motion: MotionType,
+    last_motion: MotionType,
```
oleflb, line 14: you don't need a member variable for this. You can just do `let last_motion = self.current_motion` before the motion transition

oleflb (changes requested):
Looks much better already! Only minor remarks now :)

PR #1865, opened 2025-06-04
Fail softly and carry a big error

### LINKED ISSUE #1820: Illegal position during opponent goal kick
During the scenario `kicking_team_filtering.rs` on #1782, starting from tick ~6700, player number six stands directly next to the ball during an opponents goal kick. This is an illegal position and should be investigated. 

Image

knoellle:
## Why? What?

- Introduces soft errors to the bevyhavior simulator. This allows scenarios or other bevy systems to emit errors which do not immediately cause the scenario to abort but still mark it as failed.
- Adds a check to all scenarios emitting such an error if at any point a robot attempts to knowingly walk into a `RuleObstacle`.
- Fixes #1820 

## ToDo / Known Issues

## How to Test

All tests should pass.
Fix for the broken behavior can be observed in `kicking_team_filtering` scenario frame 6700.

### DIFF
--- a/crates/bevyhavior_simulator/src/lib.rs
+++ b/crates/bevyhavior_simulator/src/lib.rs
@@ -14,6 +14,8 @@ pub mod robot;
 pub mod scenario;
 pub mod server;
 pub mod simulator;
+pub mod soft_error;
+pub mod test_rules;
 pub mod time;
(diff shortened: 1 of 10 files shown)


pejotejo (approved):
lgtm

PR #1869, opened 2025-06-05
Fix replayer zoom

oleflb:
## Why? What?

Fixes and refactors the zoom and pan implementation in the replayer.
This effectively fixes the replayer crashing on zoom, since the zoom implementation is now correct.
However it's still possible to zoom very far, which should result in the same issue again

Fixes #

## ToDo / Known Issues

## How to Test

* Zoom and pan in the replayer, it should not crash

### DIFF
--- a/crates/hulk_replayer/src/coordinate_systems.rs
+++ b/crates/hulk_replayer/src/coordinate_systems.rs
@@ -278,6 +278,54 @@ impl ScreenRange {
     }
 }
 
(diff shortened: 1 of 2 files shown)


oleflb:
I think shifting is more intuitive than clamping, since otherwise moving towards the bounds results in a zoom effect

schmidma (approved):

PR #1871, opened 2025-06-07
KeeperMotion in Penalty shootout

pejotejo:
## Why? What?
New keeper motion in penalty shootout

## How to Test

do a penalty shootout


crates/control/src/motion/motion_selector.rs
```diff
@@ -142,7 +142,8 @@ fn transition_motion(
         (_, true, MotionType::KeeperJumpRight, _) => MotionType::KeeperJumpRight,
         (_, true, MotionType::KeeperJumpLeft, _) => MotionType::KeeperJumpLeft,
         (_, _, MotionType::CenterJump, _) => MotionType::CenterJump,
-
```
schluis, line 145: Everything seems good, but why sets `ArmsUpSquat` never motion_safe_exits to true?

schluis (changes requested):

schluis (approved):
@larskna wants to fix the state with motion_safe_exits later. Approving for now

PR #1876, opened 2025-06-12
slower head movement for defender

### LINKED ISSUE #1730: post-game task: make defenders look around slower
the movement of the head could be a bit slower, as the look-around motion caused a fair bit of wobble

JackyBloxx:
## Why? What?
make the defender look around slower too reduce the wobble from the look around motion

Fixes #1730 

## ToDo / Known Issues

## How to Test
look if the defender look around slower than the other rolles

### DIFF
--- a/crates/control/src/motion/head_motion.rs
+++ b/crates/control/src/motion/head_motion.rs
@@ -6,13 +6,7 @@ use filtering::low_pass_filter::LowPassFilter;
 use framework::MainOutput;
 use serde::{Deserialize, Serialize};
 use types::{
-    cycle_time::CycleTime,
-    joints::head::HeadJoints,
-    motion_command::{HeadMotion as HeadMotionCommand, MotionCommand},
-    motion_selection::MotionSelection,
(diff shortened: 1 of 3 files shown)


crates/control/src/motion/head_motion.rs
```diff
@@ -36,6 +37,7 @@ pub struct CycleContext {
     cycle_time: Input<CycleTime, "cycle_time">,
     has_ground_contact: Input<bool, "has_ground_contact">,
     motion_selection: Input<MotionSelection, "motion_selection">,
+    world_state: Input<WorldState, "world_state">,
```
knoellle, line 40: I don't like the practice of using the world state as an input outside of behavior, the interface is way too wide.
Currently only the look_at motion also does this, even though it only needs a ground to field.
You only need the role here, so use that directly.

In general, I think motion information should only be carried inside the MotionCommand, but that needs a larger refactor I think.

schluis (approved):

PR #1884, opened 2025-06-14
Don't crash twix on unknown panel type

knoellle:
## Why? What?

Currently twix panics when loading an app.ron with an unknown panel type.
This can happen if you are on a branch with an experimental new panel and then switch back to main.
Twix will try to restore its old state but doesn't know what to do with the experimental panel.

This PR changes the error handling by instead storing the error as well as the data twix tried to build a panel from.

It is then displayed as such in the UI:
image
In this example I changed one of the `_panel_type` fields in my app.ron from "Text" to "ext".
The JSON body label is collapsed by default.

Fixes #

- Allow editing the json and have a button to try loading it again.

## How to Test

Manually edit `~/.local/share/twix/app.ron`, breaking one of the `_panel_type` fields or change to a branch that introduces a new panel and try to load from a twix without support for this panel type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8794,7 +8794,7 @@ dependencies = [
 
 [[package]]
 name = "twix"
-version = "0.9.6"
+version = "0.10.0"
 dependencies = [
  "aliveness",
  "argument_parsers",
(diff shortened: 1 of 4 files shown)


rmburg (approved):
Abfahrt :steam_locomotive:

PR #1889, opened 2025-06-16
Bump clippy to 1.84.0, fix respective lints

schmidma:
## Why? What?

The SDK now contains 1.84.0, so we can bump clippy a little and fix respective lints by clippy and rustanalyzer.

## How to Test

CI should take care of that...


crates/buffered_watch/src/receiver.rs
```diff
@@ -21,7 +21,7 @@ impl<T> Receiver<T> {
     }
 
     /// Borrows the latest value
-    pub fn borrow(&mut self) -> ReceiverGuard<T> {
+    pub fn borrow(&'_ mut self) -> ReceiverGuard<'_, T> {
```
rmburg, line 24: Why this change? I don't think it improves readability, and I couldn't find a clippy lint that suggests this change.

schmidma: https://github.com/rust-lang/rust/pull/138677

PR #1890, opened 2025-06-18
Cleanup `linear_algebra`

### LINKED ISSUE #1893: Fix Isometry API inconsistency
`Isometry2::from_parts` currently takes an angle, I'd argue it should take a `UnitComplex` instead.
`Isometry2::new` should be added, and take an angle as an argument. `Pose2` already works like this.

schmidma:
## Why? What?

This PR prepares `linear_algebra` to be published, cleaning up structure and some naming, as well as providing feature flags and documentation.

fixes #1893


crates/linear_algebra/src/lib.rs
```diff
@@ -1,3 +1,50 @@
...
+//! ```
+//!
+//! ## Philosophy
+//!
+//! This crate is a thin, zero-cost wrapper around [`nalgebra`], adding type-level tags for
+//! coordinate systems. It does not reimplement linear algebra, but provides a safer API for
+//! geometric programming.
+//!
+//! [`nalgebra`]: https://docs.rs/nalgebra
+
```
oleflb, line 47: I'd like to see links to `Framed` and `Transform` as a form of getting started here

crates/linear_algebra/src/framed.rs
```diff
@@ -1,376 +1,392 @@
...
+    pub const fn wrap(inner: T) -> Self {
+        Framed {
+            frame: PhantomData,
+            inner,
+        }
+    }
+}
+
+// Add
+impl<Frame, T, U> Add<Framed<Frame, U>> for Framed<Frame, T>
```
oleflb, line 78: No strong feelings here, but why did you choose to replace the more expressive generic type parameter names by harder-to-understand ones? What about using `Lhs` and `Rhs` as short but easy to understand alternatives?

crates/linear_algebra/Cargo.toml
oleflb, line 1: Shouldn't we give a separate version to this crate now? Also let's fill the `description`, `categories`, and `keywords`

oleflb: Maybe we should also add a readme, since that is shown on crates.io

oleflb (changes requested):

crates/linear_algebra/Cargo.toml
```diff
@@ -6,9 +6,16 @@ license.workspace = true
...
 num-traits = { workspace = true }
-path_serde = { workspace = true }
-serde = { workspace = true }
+path_serde = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
 simba = { workspace = true }
+
+[features]
+serde = ["dep:serde"]
+path_serde = ["dep:path_serde"]
```
knoellle, line 18: Do we publish this optional dependency as well?
Does not doing so break crates.io?

schmidma: I'm not sure... If that is the case, the only option (besides publishing `path_serde` as well) would be to provide the `path_serde` implementations in the `path_serde` crate...

crates/linear_algebra/Cargo.toml
oleflb: not sure what you mean, don't you want to publish it to crates.io?

crates/linear_algebra/src/orientation.rs
```diff
@@ -86,6 +94,10 @@ where
         Rotation3::wrap(self.inner.rotation_to(&other.inner))
     }
 
+    /// Converts this orientation into a rotation (transform) from a source frame to this frame.
+    ///
+    /// This adds a source frame, yielding a transform from `From` to `Frame`. Use this when you
+    /// want to treat the orientation as a transform from another frame into this frame.
     pub fn rotation<From>(self) -> Rotation3<From, Frame, T> {
```
rmburg, line 101: in the 2D case, this method is called `as_transform`, here it's `rotation`. Is there a reason for this mismatch? Otherwise please rename

schmidma: good catch. done...

schmidma:
I gave this another bump, review, cleanup and rebase...

crates/bevyhavior_simulator/src/bin/intercept_ball.rs
```diff
@@ -32,7 +32,14 @@ fn startup(
...
     let mut robot = Robot::new(PlayerNumber::One);
-    *robot.ground_to_field_mut() = Isometry2::from_parts(vector![-2.0, 0.0], 0.0);
+    *robot.ground_to_field_mut() = Isometry2::new(vector![-2.0, 0.0], 0.0);
+    robot.parameters.step_planner.max_step_size.forward = 1.0;
+    robot.parameters.step_planner.max_step_size.left = 1.0;
+    robot.parameters.step_planner.request_scale = Step {
+        forward: 1.0,
+        left: 1.0,
+        turn: 1.0,
+    };
```
rmburg: Was this change on purpose?
```suggestion
```

rmburg (commented):
only a few nitpicks, otherwise very nice changes! Looking forward to this on crates.io :)

crates/linear_algebra/src/point.rs
```diff
@@ -9,6 +9,17 @@ pub type Point<Frame, const DIMENSION: usize, T = f32> =
...
+///
+/// This macro works like [`nalgebra::point!`], but wraps the result with a frame.
+///
+/// # Example
+/// ```rust
+/// use linear_algebra::{point, Point2};
+///
+/// struct World;
+/// let p: Point2<World> = point![1.0, 2.0];
+/// ```
```
knoellle, line 23: Please add another example that specifies the frame within the macro.
```rs
let p = point![<World>, 1.0, 2.0];
```

crates/linear_algebra/src/vector.rs
```diff
@@ -8,6 +8,17 @@ pub type Vector<Frame, const DIMENSION: usize, Scalar = f32> =
...
+///
+/// This macro works like [`nalgebra::vector!`], but wraps the result with a frame.
+///
+/// # Example
+/// ```rust
+/// use linear_algebra::{vector, Vector2};
+///
+/// struct World;
+/// let v: Vector2<World> = vector![1.0, 2.0];
+/// ```
```
knoellle, line 22: Please add another example that specifies the frame within the macro.
```rs
let p = vector![<World>, 1.0, 2.0];
```

crates/linear_algebra/src/isometry.rs
```diff
@@ -10,46 +10,67 @@ pub type Isometry<From, To, const DIMENSION: usize, T, Rotation> =
...
+    /// Creates an isometry from a translation and orientation in the destination frame.
+    pub fn from_parts(translation: Vector2<To, T>, orientation: Orientation2<To, T>) -> Self {
+        Self::wrap(nalgebra::Isometry2::from_parts(
+            translation.inner.into(),
+            orientation.inner,
+        ))
     }
 
+    /// Creates a pure rotation isometry with zero translation.
     pub fn rotation(angle: T) -> Self {
```
knoellle: In terms of naming, I think `Isometry::rotation` should return a `Rotation<From, To>` and creating a pure rotation isometry from an angle should be called something like `from_angle`.

Likewise, if we were to add such functions, 
- `from_rotation` should take a `Rotation` and return `Self`
- `angle` should take `&self` and return a `T`

crates/linear_algebra/src/isometry.rs
```diff
@@ -75,32 +96,46 @@ where
...
         ))
     }
 
+    /// Creates a pure rotation isometry with zero translation.
     pub fn from_rotation(axisangle: Vector3<To, T>) -> Self {
         Self::wrap(nalgebra::Isometry3::rotation(axisangle.inner))
     }
 
+    /// Creates a pure translation isometry.
     pub fn from_translation(x: T, y: T, z: T) -> Self {
```
knoellle: How do we want to handle passing a Vector as translation vs. individual coordinates.
The parameter in `new` is also called translation, I think these should be consistent.

crates/linear_algebra/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//! ensures that only compatible operations are allowed, catching errors at compile time.
+//!
+//! ## Features
+//!
+//! - Enforces coordinate system correctness at compile time using Rust's type system.
+//! - Separates Vectors from Points, Isometries from Poses, etc.
+//! - Wraps commonly used parts of the [`nalgebra`] API to provide frame-safe abstractions.
+//! - Supports 2D and 3D geometry with extensible coordinate system tagging.
+//! - Provides clear and explicit geometric transformations.
+//! - Re-exports [`nalgebra`] so `point!` and `vector!` work without a direct dependency.
```
knoellle, line 24: Is this a "feature"?
I think its an implementation detail. It would work fine too if instead of wrapping nalgebra's macros we had copied them instead.

crates/linear_algebra/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//! ## Motivation
+//!
+//! Geometric computations often involve multiple coordinate systems. Mixing up these systems can
+//! lead to subtle and hard-to-find bugs. By encoding coordinate systems in types, this crate
+//! ensures that only compatible operations are allowed, catching errors at compile time.
+//!
+//! ## Features
+//!
+//! - Enforces coordinate system correctness at compile time using Rust's type system.
+//! - Separates Vectors from Points, Isometries from Poses, etc.
```
knoellle: The first one nalgebra already does, the second one is something we added.
Maybe that should be made clear to the reader?

Possibly also instead of talking about Poses, talk about types existing with vs. converting between coordinate systems. A potential new user probably doesn't already know how we use the term "Pose".

crates/linear_algebra/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//!
+//! ## Example
+//!
+//! ```rust
+//! use linear_algebra::{point, vector, Isometry3, Orientation3, Point3};
+//!
+//! struct Camera;
+//! struct Ground;
+//!
+//! fn ball_to_ground(
```
knoellle:
```suggestion
//! fn ball_in_ground(
```
`ball_to_ground` sounds like a transformation.

crates/linear_algebra/src/lib.rs
```diff
@@ -1,3 +1,51 @@
...
+//! - Provides clear and explicit geometric transformations.
+//! - Re-exports [`nalgebra`] so `point!` and `vector!` work without a direct dependency.
+//!
+//! ## Example
+//!
+//! ```rust
+//! use linear_algebra::{point, vector, Isometry3, Orientation3, Point3};
+//!
+//! struct Camera;
+//! struct Ground;
```
knoellle, line 32: Do we ever explicitly state that a coordinate system can be any marker type the user wants?
Here or before/after the example code block might be a good place.

crates/linear_algebra/src/pose.rs
```diff
@@ -5,28 +5,59 @@ use crate::{Framed, Isometry2, Isometry3, Orientation2, Orientation3, Point2, Po
...
             orientation.inner,
         ))
     }
 
+    /// Returns a new pose with zero translation and zero orientation.
     pub fn zero() -> Self {
         Default::default()
     }
 
+    /// Returns the pose as an isometry transform, labeling the Pose frame as `From`.
```
knoellle:
```suggestion
    /// Returns the pose as an isometry, transforming from the pose-relative `From` coordinate system to the `Frame` coordinate system the pose exists in.
```

crates/linear_algebra/src/pose.rs
```diff
@@ -57,27 +88,41 @@ where
...
 {
+    /// Create a new pose from a position and orientation.
     pub fn from_parts(position: Point3<Frame, T>, orientation: Orientation3<Frame, T>) -> Self {
         Self::wrap(nalgebra::Isometry3::from_parts(
             position.inner.into(),
             orientation.inner,
         ))
     }
+
+    /// Returns the pose as an isometry transform, labeling the Pose frame as `From`.
```
knoellle:
```suggestion
    /// Returns the pose as an isometry, transforming from the pose-relative `From` coordinate system to the `Frame` coordinate system the pose exists in.
```

knoellle (changes requested):

crates/linear_algebra/src/point.rs
schmidma: done

crates/linear_algebra/src/vector.rs
schmidma: done

crates/linear_algebra/src/isometry.rs
schmidma: Done, I renamed `from_rotation` to `from_axis_angle`. `angle()` returns `T`

crates/linear_algebra/src/isometry.rs
schmidma: The initial attempt was to align it with nalgebra... But I removed that, `from_translation` now always takes a `Vector3`

crates/linear_algebra/src/lib.rs
schmidma: rephrased some of the docstrings

crates/linear_algebra/src/lib.rs
schmidma: I hope this is clearer now. Added a small sentence on what a `Pose` is

crates/linear_algebra/src/lib.rs
schmidma: fixed

crates/linear_algebra/src/lib.rs
schmidma: added a docstring

crates/linear_algebra/src/pose.rs
schmidma: done

crates/linear_algebra/src/pose.rs
schmidma: done

crates/linear_algebra/src/point.rs
knoellle: The whole point (heh) of being able to specify the frame inside the macro was being able to omit more verbose type information elsewhere.

crates/linear_algebra/src/isometry.rs
```diff
@@ -186,3 +196,29 @@ where
...
+    struct Robot;
+    #[derive(Debug)]
+    struct Field;
+
+    #[test]
+    fn pure_rotation_isometry2_exposes_rotation_and_angle_on_self() {
+        let isometry = Isometry2::<Robot, Field>::from_angle(0.5);
+
+        assert!((isometry.angle() - 0.5).abs() < 1.0e-6);
+        assert!((isometry.rotation().angle() - 0.5).abs() < 1.0e-6);
```
knoellle, line 215: This seems like a job for `approx`

crates/linear_algebra/src/lib.rs
knoellle: I'm still not convinced the re-exporting is a feature worthy of being listed here.

PR #1905, opened 2025-07-05
deny standup after 3 tries and ouch

### LINKED ISSUE #1756: deny standup after `X` tries
After failing to standup multiple times, we could guess the robot is broken and just give up to protect robot hardware

vivi-enne:
## Why? What?

Make ouch sound and change motion command when starting 3rd stand up attempted (parameter debatable). This should be to avoid standing up too often in the competition when the robot is broken and the referee does not see it. 

Fixes #1756

## ToDo / Known Issues

## How to Test

Let robot (gently) stand up 3 or more times in a row and observe an unstiff robot and a single ouch sound. Making the robot stable again (Penalized, Walking, Stand, Initial) brings it back to normal function.

### DIFF
--- a/crates/control/src/behavior/stand_up.rs
+++ b/crates/control/src/behavior/stand_up.rs
@@ -4,7 +4,10 @@ use types::{
     world_state::WorldState,
 };
 
-pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+pub fn execute(world_state: &WorldState, maximum_standup_attempts: u32) -> Option<MotionCommand> {
+    if world_state.robot.stand_up_count >= maximum_standup_attempts {
+        return Some(MotionCommand::Unstiff);
+    }
     let kind = match world_state.robot.fall_state {
         FallState::Fallen { kind } => kind,
         FallState::StandingUp { kind, .. } => kind,
--- a/crates/control/src/behavior/stand_up.rs
+++ b/crates/control/src/behavior/stand_up.rs
@@ -5,7 +5,7 @@ use types::{
 };
 
 pub fn execute(world_state: &WorldState, maximum_standup_attempts: u32) -> Option<MotionCommand> {
-    if world_state.robot.stand_up_count >= maximum_standup_attempts {
+    if world_state.robot.stand_up_count > maximum_standup_attempts {
         return Some(MotionCommand::Unstiff);
     }
     let kind = match world_state.robot.fall_state {
--- a/crates/control/src/motion/motion_selector.rs
+++ b/crates/control/src/motion/motion_selector.rs
@@ -14,6 +14,7 @@ use types::{
 pub struct MotionSelector {
     last_motion: MotionType,
     stand_up_count: u32,
+    was_standing_up: bool,
 }
 
 #[context]
@@ -43,6 +44,7 @@ impl MotionSelector {
         Ok(Self {
             last_motion: MotionType::Unstiff,
             stand_up_count: 0,
+            was_standing_up: false,
         })
     }
 
@@ -57,8 +59,15 @@ impl MotionSelector {
             *context.has_ground_contact,
         );
 
-        let stand_up_count =
-            stand_up_counting(self.last_motion, current_motion, self.stand_up_count);
+        let is_standing_up =
+            current_motion.is_dispatching() && requested_motion.is_standup_motion();
+
+        let stand_up_count = stand_up_counting(
+            current_motion,
+            is_standing_up,
+            self.was_standing_up,
+            self.stand_up_count,
+        );
 
         if self.stand_up_count <= *context.maximum_standup_attempts
             && stand_up_count > *context.maximum_standup_attempts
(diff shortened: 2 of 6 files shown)


etc/parameters/default.json
```diff
@@ -1348,7 +1348,9 @@
     "maximum_lookaround_duration": {
       "nanos": 0,
       "secs": 6
-    }
+    },    
+    "maximum_standup_attempts": 3
+
```
knoellle, line 1353: Remove this empty line

crates/control/src/behavior/stand_up.rs
```diff
@@ -4,7 +4,10 @@ use types::{
     world_state::WorldState,
 };
 
-pub fn execute(world_state: &WorldState) -> Option<MotionCommand> {
+pub fn execute(world_state: &WorldState, maximum_standup_attempts: u32) -> Option<MotionCommand> {
+    if world_state.robot.stand_up_count >= maximum_standup_attempts {
```
knoellle, line 8: The counter increments the moment we begin the standup motion.
If `maximum_standup_attempts` is set to 1, we'll give up on the first attempt, right?
I think this should be just `>`.

crates/control/src/motion/motion_selector.rs
```diff
@@ -51,9 +57,19 @@ impl MotionSelector {
             *context.has_ground_contact,
         );
 
-        self.stand_up_count =
+        let stand_up_count =
             stand_up_counting(self.last_motion, current_motion, self.stand_up_count);
 
+        if self.stand_up_count < *context.maximum_standup_attempts
+            && stand_up_count >= *context.maximum_standup_attempts
```
knoellle, line 64: These comparisons have the same off-by-one issue as well.

knoellle (changes requested):

etc/parameters/default.json
vivi-enne: Done

crates/control/src/motion/motion_selector.rs
vivi-enne: Done

crates/control/src/motion/motion_selector.rs
```diff
@@ -51,8 +59,25 @@ impl MotionSelector {
             *context.has_ground_contact,
         );
 
-        self.stand_up_count =
-            stand_up_counting(self.last_motion, current_motion, self.stand_up_count);
+        let is_standing_up =
+            current_motion.is_dispatching() && requested_motion.is_standup_motion();
+
+        let stand_up_count = stand_up_counting(
```
schluis, line 65: I'm not a huge fan of the `ing`, I think this is quite uncommon in our codebase?
```suggestion
        let stand_up_count = stand_up_count(
```
```suggestion
        let stand_up_count = count_stand_ups(
```

crates/control/src/motion/motion_selector.rs
```diff
@@ -51,8 +59,25 @@ impl MotionSelector {
             *context.has_ground_contact,
         );
 
-        self.stand_up_count =
-            stand_up_counting(self.last_motion, current_motion, self.stand_up_count);
+        let is_standing_up =
```
schluis, line 62:
```suggestion
        let switching_to_standing_up =
```

crates/control/src/motion/motion_selector.rs
```diff
@@ -51,8 +59,25 @@ impl MotionSelector {
...
+
+        let stand_up_count = count_stand_up_attempts(
+            current_motion,
+            is_standing_up,
+            self.was_standing_up,
+            self.stand_up_count,
+        );
+
+        if self.stand_up_count <= *context.maximum_standup_attempts
+            && stand_up_count > *context.maximum_standup_attempts
```
schluis, line 73: Why?

schluis (changes requested):
I like it! :) 🚀

crates/control/src/motion/motion_selector.rs
schluis:
```suggestion
        let is_dispatching_to_standing_up =
```

crates/control/src/motion/motion_selector.rs
schluis: I'm pretty sure this should be possible.

For example like this:
```rs
let stand_up_count = count_stand_up_attempts(
    current_motion,
    is_standing_up,
    self.was_standing_up,
    self.previous_stand_up_count,
);

if self.previous_stand_up_count == *context.maximum_standup_attempts {
    context
        .hardware_interface
        .write_to_speakers(SpeakerRequest::PlaySound { sound: Sound::Ouch });
}

self.previous_stand_up_count = stand_up_count;

let dispatching_motion = if current_motion == MotionType::Dispatching {
    if requested_motion == MotionType::Unstiff {
        Some(MotionType::SitDown)
    } else {
        Some(requested_motion)
    }
} else {
    None
};

*context.stand_up_count = stand_up_count;
```

vivi-enne: Done

schluis (approved):

PR #1910, opened 2025-07-07
Kick decision step planning

### LINKED ISSUE #1186: Prefer Foot that doesn't need Zero Step
When deciding for kicks, we currently only consider our pose for deciding for a kick-decision. That means, only finding the closest kick pose to our current position (plus some obstacle avoidance stuff).

We should also check whether the kick we want to execute actually has the correct support foot in the next step. If this is not the case, if we want to kick with a left foot starting step and the next support foot is the left foot, then walking cannot execute the kick directly, but rather has to do a "zero-step".

When we anticipate that and use it in the decision, we can save a single step and be quicker in kicking the ball.

knoellle:
## Why? What?

Currently, step plans are ordered by a distance metric (also taking into account orientation).
But not all kick decisions are created equal, some may be better even if slightly further away because the foot used to kick the ball fits better to the current support foot "phase".
With this PR, the kick selector does greedy step planning for each kick decision and calculates the total duration of these steps.
This duration metric is then used instead of the distance metric for the calculating the Güte of kick decisions.

**You may want to review the first commit separately.**

Fixes #1186

## ToDo / Known Issues

- [ ] Actually use the calculated walk duration for comparing kick decisions.
- [ ] Think long and hard again about how to compare the kicking side to the current support foot
- [ ] Clean up the walking engine mess
- [ ] Check if the robot can get stuck oscillating between two kick decisions

## How to Test

A robot walking in place in front of a ball should repeatedly switch which kick decision it thinks is best depending on the current support foot.


rmburg:
we no longer perform step planning

PR #1912, opened 2025-07-09
Replace `Arc<Vec<YCbCr422>>` with `Arc<[YCbCr422]>`

oleflb:
## Why? What?

`YCbCr422` image had an unnecessary double indirection. 

Fixes #

## ToDo / Known Issues

## How to Test

* CI takes care of that

### DIFF
--- a/crates/types/src/ycbcr422_image.rs
+++ b/crates/types/src/ycbcr422_image.rs
@@ -29,7 +29,7 @@ pub struct YCbCr422Image {
     width_422: u32,
     height: u32,
     #[path_serde(leaf)]
-    buffer: Arc<Vec<YCbCr422>>,
+    buffer: Arc<[YCbCr422]>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/types/src/ycbcr422_image.rs
```diff
@@ -54,12 +54,12 @@ impl From<RgbImage> for YCbCr422Image {
...
                 [left_color, right_color].into()
             })
-            .collect();
+            .collect::<Vec<YCbCr422>>();
 
         Self {
             width_422,
             height,
-            buffer: Arc::new(data),
+            buffer: Arc::from(data),
```
schmidma, line 62: How does this behave with respect to copying? Is this just taking ownership and moving the pointers behind `Vec`?

schmidma:
This does not work, as this `From` is copying every element into a newly allocated slice.

Issue #1826, opened 2025-04-09
Reduce walking speed when hot


schluis:
Reduce the `max_step_size` in the `step_planner.rs` if the maximal joint temperature (can be found in `sensor_data`) is highter than `const TEMPERATURE_LEVEL_ONE: f32 = 76.0;`.
- Think about how this affects time to reach and talk to @knoellle about it

philipniklas-r:
Small input: maybe add a field for the maximal **leg** joint temperature. Otherwise walking slower when one arm joint is hot is not necessary ;)

Also the sensor data from the NAO also contains a status field for the temperature, e.g. **regular**, **hot**, **veryHot** or **criticallyHot**. Based on those the stiffness is reduced. This might be important, as once a joint exceeds 76°C, the status will switch to **hot**. But I think it needs to be below 70°C before switching back to **regular**, and only then the stiffness is increased again.

pejotejo:
continued in #1914

PR #1914, opened 2025-07-09
walking slower when hot

pejotejo:
## Why? What?
to increase stability when the leg joints get hot
#1826 

implement it into Rasmus' StepPlanning

## How to Test

let the Nao get hot and look if it still can run fast (it shouldn't)

### DIFF
--- a/crates/energy_optimization/src/current_minimizer.rs
+++ b/crates/energy_optimization/src/current_minimizer.rs
@@ -1,3 +1,4 @@
+use color_eyre::eyre::{ContextCompat, Result};
 use filtering::hysteresis::less_than_with_absolute_hysteresis;
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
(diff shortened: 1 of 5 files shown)


schluis (changes requested):
Nice work

oleflb (approved):

crates/energy_optimization/src/current_minimizer.rs
```diff
@@ -43,22 +44,21 @@ impl CurrentMinimizer {
         let (joint, maximal_current) = currents
             .enumerate()
             .max_by(|(_, left), (_, right)| f32::total_cmp(left, right))
-            .expect("currents must not be empty.");
+            .wrap_err("currents must not be empty.")?;
```
schmidma, line 47: This can never fail... why did you introduce a runtime error here?

schmidma (changes requested):

schluis (dismissed):

PR #1965, opened 2025-07-16
Improve Camera Calibration

oleflb:
## Why? What?

- The `correction_in_camera_{top, bottom}` parameters were redundant with `extrinsic_rotations`, therefore I removed the latter.
- Now the "Save to head" button in the semi-automatic calibration ui is disabled when the optimizer did not yet converge
- The `ground_to_field` CyclerState was only updated during Playing. Also, no `ground_to_field` was produced in the calibration primary state. This resulted in very wrong `ground_to_field` values if you don't manage to get the robot into calibration mode in the first try, since then the cyclerstate contains the ready pose

Fixes #

## ToDo / Known Issues

- [x] Manual calibration ui now shows radians instead of degrees

## How to Test

* Perform semi-automatic camera calibration and inspect field lines. Make sure to also check the transition Unstiff -> Penalized -> Playing/Ready -> Unstiff -> Penalized -> Calibration


knoellle (approved):
LGTM

PR #1976, opened 2025-07-16
Remove unused maximum_step_size

rmburg:
## Why? What?

I came across this (seemingly) unused parameter and suspect it is not there on purpose.
@julianschuler as the author of the code, can you confirm or deny this?

## ToDo / Known Issues

- [ ] await Julian's response

None

## How to Test

Stare at the code

### DIFF
--- a/crates/control/src/behavior/intercept_ball.rs
+++ b/crates/control/src/behavior/intercept_ball.rs
@@ -9,14 +9,12 @@ use types::{
     motion_command::{HeadMotion, ImageRegion, MotionCommand, OrientationMode, WalkSpeed},
     parameters::InterceptBallParameters,
     planned_path::PathSegment,
-    step::Step,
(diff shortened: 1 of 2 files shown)


crates/control/src/behavior/intercept_ball.rs
```diff
@@ -54,16 +52,6 @@ pub fn execute(
...
 
-            let Step {
-                forward,
-                left,
-                turn: _,
-            } = maximum_step_size;
-
-            if forward == 0.0 || left == 0.0 {
-                return None;
-            }
```
julianschuler, line 65: I'm not sure what the reasoning was here. 
It makes sense that we can't do intercepting steps when we can't move, though that is not something to decide at a behavior level. Resolve when read.

julianschuler (approved):
Wech damit

PR #2001, opened 2025-07-17
Build parameter tester faster

knoellle:
## Why? What?

On main, the parameter tester depends on robotics code (vision, control crate, etc.), which is not strictly necessary.
This PR makes the parameter tester depend only on the dependencies of robotics code, which reduces the amount of crates to compile from 428 to 320 and the compile time on my machine from just over a minute to 40 seconds.

Fixes #

## How to Test

```sh
time pepsi build parameter_tester
```


rmburg (approved):
schaut gut zu mir

PR #2018, opened 2025-07-17
Tune feet-detection parameter for RC25

paudar:
## Why? What?

Update Feet-Detection Parameter for rc25. Now there are less falls Positives in Top and less false negative in Bottom.

## ToDo / Known Issues

It's better than before.

Do again at next competition

## How to Test

Use the replayer or trust me.


schluis (approved):
Sehr schön 🥇

PR #2026, opened 2025-07-18
Clip goal-line targets with ball position

### LINKED ISSUE #2024: post-game task: Don't aim for the goal line when behind the ball

schmidma:
## Why? What?

Moves the goal line targets and corner kick targets such that they are relative to the robot, if the robot exceeds the opponent line.

This can lead to more jockeling, as the targets move as soon as the robot is behind the goal line. But probably still better than kicking out knowingly.

fixes #2024


rmburg (approved):
Abfahrt 🚂

Issue #1679, opened 2025-03-09
Dynamically reclaim Striker


schmidma:
A thought on advancing role assignment:

A robot could actively re-announce its striker role if the situation changed *significantly* (ball moved away, or even to another robot). Such a striker announcement allows other robots to counter-claim striker quicker, than waiting for the usual 2 second re-transmit interval.

This needs careful thoughts on handling messaging budget, but also seems like a strong improvement of quicker striker decisions for a supporter being closer after the striker had a dueling situation.

PR #2044, opened 2025-07-18
Don't trust network teamballs

paudar:
## Why? What?

Parameter change to not claime Striker depending on network balls. In the last games on the RC25 lost of faulty striker claims could be observed.

## ToDo / Known Issues

Enable again when localisation is perfect.

Dynamicly trust network balls (see #????) or dynamic reclaim (see #1679).

## How to Test

None


paudar:
Due to an failing test, this change is applied every game if necessary.

PR #2057, opened 2025-07-19
Implement Filter Recalibration

oleflb:
## Why? What?

Recalibrates the orientation filter when robot is not moving

Fixes #

## ToDo / Known Issues

## How to Test

* Check the recalibrated output


crates/control/src/orientation_filter.rs
```diff
@@ -63,64 +64,68 @@ impl OrientationFilter {
...
-                    if *remaining_cycles == 0 {
-                        let gravity = -filtered_gravity.state().inner;
-                        let up = nalgebra::Vector3::y();
-                        let orientation = UnitQuaternion::look_at_rh(&gravity, &up);
-                        self.state = State::Filtering { state: orientation };
-                    }
-                } else {
-                    self.state = State::WaitingForSteady;
+                    self.calibration_state = State::WaitingForSteady;
+                    break 'update;
```
schluis, line 90: @pejotejo findest du das leserlich? Oder lieber ein `else`?

PR #2086, opened 2025-07-20
Prevent yielding striker to differently-localized robots

knoellle:
## Why? What?

If another robot claims striker but sends a very different ball position, that most likely means one of the robots is not localized correctly. With this change, both robots will stay striker as only claims on the "same" ball are recognized.
The parameter value is just a guess, suggestions welcome.

Fixes #

## How to Test

This is currently not testable in simulator.
Deploy two robots, one localized correctly, the other one not. They should both be striker if they have different ideas on where the ball is on the field, but one should yield if they believe the ball to be in a similar location.

### DIFF
--- a/crates/types/src/ball_position.rs
+++ b/crates/types/src/ball_position.rs
@@ -1,9 +1,9 @@
-use std::time::SystemTime;
+use std::{ops::Mul, time::SystemTime};
 
 use serde::{Deserialize, Serialize};
 
 use coordinate_systems::Field;
-use linear_algebra::{Point2, Vector2};
+use linear_algebra::{Isometry2, Point2, Vector2};
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
 
 #[derive(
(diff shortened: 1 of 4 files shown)


etc/parameters/default.json
```diff
@@ -1207,8 +1207,9 @@
   "role_assignment": {
     "forced_role": null,
     "keeper_replacementkeeper_switch_time": { "nanos": 0, "secs": 12 },
+    "claim_striker_from_team_ball": false,
     "maximum_trusted_team_ball_age": { "nanos": 0, "secs": 1 },
-    "claim_striker_from_team_ball": false
+    "maximum_trusted_team_ball_distance": 50.0
```
schluis, line 1212: That seems quite high

crates/types/src/ball_position.rs
```diff
@@ -15,6 +15,31 @@ pub struct BallPosition<Frame> {
...
+    }
+}
+
+impl<From, To> Mul<BallPosition<From>> for Isometry2<From, To> {
+    type Output = BallPosition<To>;
+
+    fn mul(self, rhs: BallPosition<From>) -> Self::Output {
+        BallPosition {
+            position: self * rhs.position,
+            velocity: self * rhs.velocity,
```
schluis, line 37: It seems weird to me to also multiply the velocity. Can you elaborate on this?

schluis (changes requested):
I would prefer to go the other way around, i.e., to only send a striker message when seeing the ball instead of always sending a striker message and choosing to ignoring it, but I understand that this is difficult in the current architecture. Apart from that 🚀

etc/parameters/default.json
knoellle: Whoops, that value was for testing purposes only, not meant to be committed.
Reverted to intended value.

crates/types/src/ball_position.rs
knoellle: As discussed, not multiplying it would mean having the velocity point in an incorrect direction for the specified frame.
A velocity in Ground can have a different direction than the same velocity in Field.

PR #2089, opened 2025-07-20
Add pepsi nextest

knoellle:
## Why? What?

Extra für @schluis: 
This PR adds a new subcommand to pepsi which launches the cargo addon "nextest".

Nextest is a faster and more featured test-runner for cargo than the builtin `cargo test`.

I deliberately chose not to mimick the arguments directly because we likely won't use most of them and I think its more ergonomic this way.
For example, `pepsi nextest --profile release` executes `cargo nextest run --cargo-profile release`
The full argument set is still available by directly invoking cargo nextest instead of going through pepsi.

Fixes #

- Use nextest in CI

## How to Test

Install cargo nextest if you haven't already, then run `./pepsi nextest`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6747,7 +6747,7 @@ checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
 name = "pepsi"
-version = "6.4.1"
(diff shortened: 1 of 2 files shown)


schluis:
Why?!?!?!?!?!

rmburg (approved):
gefällt 👍

PR #2095, opened 2025-07-22
Fix autoref ball positioning in sub states

knoellle:
## Why? What?

- Fixed multiple instances of the x coordinate of the ball position being used by the autoref to determine the field side instead of the y coordinate. 
- Some cleanup, removing unnecessary `as_mut` and merging some if-lets

Fixes #

## How to Test

In the kick_in scenario around frame 6000, a kick in is called but on main the ball is place on the wrong side. This should no longer be the case.


rmburg (approved):
schaut gut zu mir :+1:

PR #2098, opened 2025-08-23
Bevyhaviorsimulator parameters

rmburg:
## Why? What?

Previously, if you connect twix to the behavior simulator and subscribe to `parameters`, you would only get `selected_frame`, `selected_robot`, and `field_dimensions`.

Now, `parameters` contains all normal parameters you would get from a real robot,
and the selected frame/robot are exposed under `simulator`. 

## ToDo / Known Issues

None

- prevent selection of non-existent robots (currently, the map panel goes blank in this case due to missing field dimensions)

## How to Test

Use twix

### DIFF
--- a/crates/bevyhavior_simulator/default.json
+++ b/crates/bevyhavior_simulator/default.json
@@ -1,21 +1,4 @@
 {
   "selected_frame": 0,
-  "selected_robot": 7,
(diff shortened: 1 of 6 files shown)


knoellle (approved):
LGTM

PR #2153, opened 2025-12-02
K1 SDK

alexschmander:
## Why? What?

This PR adds the new SDK for the Booster K1, as well as support for using it with `./pepsi build`. 
The SDK is defined as a `Containerfile`, built manually, and then pushed to [ghcr.io](https://github.com/orgs/HULKs/packages/container/package/k1sdk). In pepsi, by default podman will now be used to pull and run the SDK image. This then serves as the cross-compilation environment.    
The binary being built is a minimal `hulk_booster` crate, simply including a hello world.  

Depends on #2168

## ToDo / Known Issues

- [x] ~~Building the `ort` currently fails during the linking of the `ort-sys` crate. This is maybe due to incorrectly configured sysroots for bindgen. This still needs to be investigated...~~ Ort now builds in the sdk container. TensorRT does not yet build however. 
- [x] ~~`CompiledNN` does not have a `aarch64` target and thereby cannot be built for the K1. When setting ball detection back up, this needs to be considered.~~ Fixed by #2155

## Ideas for Next Iterations (Not This PR)

- Our ort setup does not yet support GPU Utilisation. Tracked in #2166 

## How to Test

- `./pepsi build booster`

PR #2154, opened 2025-12-02
Pepsi upload

alexschmander:
## Why? What?

This PR reenables most `pepsi` functionality for interacting with the K1, including: 
- `upload`
- `logs` (but `hulk.err` is not shown at all...)
- `shell`
- `ping`
- Maybe some others as well, but those were not tested yet.

Currently, the infrastructure on the K1 is setup manually, since we don't yet have a way of flashing it. @julianschuler please paste the documentation here, of what we did. 

The PR also includes a minimal DDS example in the `hulk_booster` binary subscribing and deserializing a Topic from the K1. 

Depends on #2153

## ToDo / Known Issues

## How to Test

- To reach the K1, currently one has to configure a route to `192.168.10.102` manually or set a static ip. 
- With the K1 running, run `./pepsi upload 192.168.10.102 --skip-os-check`

### DIFF
--- a/crates/robot/src/lib.rs
+++ b/crates/robot/src/lib.rs
@@ -46,11 +46,11 @@ impl TryFrom<String> for RobotNumber {
     }
 }
 
-pub struct Booster {
+pub struct Robot {
     pub address: Ipv4Addr,
 }
 
-impl Booster {
+impl Robot {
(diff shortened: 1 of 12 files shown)


tools/pepsi/Cargo.toml
knoellle: Don't forget to bump the pepsi version.

PR #2155, opened 2025-12-03
Removed ball-detection

NSVEGUR:
## Why? What?

Ball-detection dependency was built only for x86 architecture and was not working in arm. So Have removed the lines from vision related to compiled-nn in `Cargo.toml` and ball-detection in `lib.rs` of vision.

Use different inference engine that works on arm for vision.

## How to Test

Compile on arm


alexschmander (approved):
LGTM

PR #2161, opened 2025-12-06
Move camera viewport updating from PostUpdate to Update to fix crash

knoellle:
## Why? What?

What it says.
For some reason doing it in PostUpdate leads to a crash when making large, sudden changes to the window size.

Fixes #

## How to Test

Issue was reproducible by moving the window between monitors.


paudar (approved):
Fixed the bug and it works for me.

Issue #2175, opened 2026-01-09
Check camera angle in mujoco


knoellle:
This is the camera coordinate system of the mujoco model.

<img width="566" height="452" alt="Image" src="https://github.com/user-attachments/assets/8998acfd-1953-4c0e-930a-aeb9a7fa7386" />

Doesn't look right, investigate.
This misalignment is also reflected in the rendered images.

PR #2176, opened 2026-01-10
Feature fake

vivi-enne:
## Why? What?

We need a feature for the SLAM, therefore we decided to take the upper left corner of the goal. This PR also includes the `camera_to_world` transition, so that future features should be easily included. 

Fixes #

## ToDo / Known Issues

- For multiple features a better output type needs to be specified. Depends on the feature detection.
- Camera angle looks not as expected but is rotated a bit #2175 

## How to Test

Check in simulator: standing or walking robot should output a vector for `Control.main_outputs.feature` that matches the position of the left upper corner of the goal in the camera image. If it is not visible it should be None.


crates/bevy_mujoco/src/lib.rs
```diff
@@ -299,7 +299,7 @@ fn spawn_mujoco_scene(
                 );
             }
         });
-        if body.parent.is_none() && body.name.as_ref().is_some_and(|name| name == "Trunk") {
```
oleflb: What is the reason for this change?

crates/booster/src/lib.rs
```diff
@@ -21,6 +21,8 @@ pub struct LowState {
     pub motor_state_parallel: Vec<MotorState>,
     /// Serial structure joint feedback
     pub motor_state_serial: Vec<MotorState>,
+    /// Serial structure joint feedback
+    pub camera_to_world: [f32; 12],
```
oleflb: Please make this member strongly typed (e.g. `Isometry3<Camera, World>`). The conversion should be handled in the constructing method

tools/twix/src/panels/mujoco_simulator.rs
```diff
@@ -24,25 +27,47 @@ impl<'a> Panel<'a> for MujocoSimulatorPanel {
...
+            kinematics,
+            camera_to_world,
+        }
     }
 }
 
+fn toggle_gizmos(mut config_store: ResMut<GizmoConfigStore>) {
+    let (config, _) = config_store.config_mut::<DefaultGizmoConfigGroup>();
+    config.enabled = true;
+    config.render_layers = RenderLayers::from_layers(&[0, 1, 2, 3, 4, 5, 6, 7, 8]);
```
oleflb: Sincere question: what do `render_layers` mean in this context?

crates/control/src/feature_provider.rs
```diff
@@ -0,0 +1,66 @@
...
+            context.field_dimensions.length / 2.0,
+            context.field_dimensions.goal_inner_width / 2.0
+                + context.field_dimensions.goal_post_diameter / 2.0,
+            context.field_dimensions.goal_inner_height,
+        ];
+
+        let upper_left_goal_post_in_camera = world_to_camera * upper_left_goal_post;
+
+        let field_of_view_width_angle = 60.0_f32.to_radians();
+        let field_of_view_height_angle = 45.0_f32.to_radians();
```
oleflb: Don't we already have config for this?

oleflb (changes requested):
Looks nice overall, just some remarks

crates/booster/src/lib.rs
knoellle: I wasn't sure where to do the conversion. For the mujoco viewer in twix, we always export the mujoco types and convert in the client.

crates/control/src/sensor_data_receiver.rs
```diff
@@ -118,15 +123,28 @@ impl SensorDataReceiver {
...
 
+        let raw = low_state.camera_to_world;
+
+        let rotation = nalgebra::Rotation3::from_matrix_unchecked(
+            nalgebra::Matrix3::from_row_slice(&raw[3..]),
+        );
+        let translation =
+            nalgebra::Translation3::from(nalgebra::Vector3::from_row_slice(&raw[..3]));
+        let camera_to_world = nalgebra::Isometry3::from_parts(translation, rotation.into())
+            * UnitQuaternion::from_euler_angles(-FRAC_PI_2, FRAC_PI_2, 0.0);
```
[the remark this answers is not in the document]
knoellle: I didn't know this code was going to be merged tbh. 😅

The camera_to_world transform shouldn't live in the LowState at all, since this isn't something provided on the robot.
The clean approach would be to have yet another type in the simulator which contains all the ground truth data we want to use and request that in the hulk_mujoco hardware interface.

pejotejo:
changes with new framework

PR #2179, opened 2026-01-11
Reintroduce HSL Network

pejotejo:
## Why? What?
Reintroduces the Network Cycler and Game-Controller-State-Filter 

## ToDo / Known Issues
- Some Temporary Fake Inputs in the Filtered Game Controller State

- The Gamecotroller and the Simulator cannot be run on the Same device

## How to Test

- open a Gamecontroller on a second device (probably in the SPL_A) and the Simulator on yours and then look at the filtered_game_controller_state in twix

### DIFF
--- a/crates/argument_parsers/src/lib.rs
+++ b/crates/argument_parsers/src/lib.rs
@@ -11,7 +11,6 @@ use color_eyre::{
 use regex::Regex;
 
 use hsl_network_messages::PlayerNumber;
-use nao::{Network, SystemctlAction};
 use robot::{Network, SystemctlAction};
 
 pub const SYSTEMCTL_ACTION_POSSIBLE_VALUES: &[&str] =


etc/parameters/default.json
```diff
@@ -273,5 +269,59 @@
       "left": 0.0,
       "turn": 0.0
     }
+  },
+  "player_number": "Seven",
```
alexschmander: I think the default number should not be Seven anymore. This can be looked into further, when introducing role assignment again, but even for now I would change this to `"Three"`, since will be correct in both foundation and advanced games.

pejotejo: done

crates/hsl_network_messages/headers/RoboCupGameControlData.hpp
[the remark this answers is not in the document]
pejotejo: done

PR #2177, opened 2026-01-10
Booster Zenoh Bridge

julianschuler:
## Why? What?

This PR adds a bridge to make the booster ROS interfaces available via zenoh.

Blocked on #2154.

## ToDo / Known Issues

- Has to be started manually

## Ideas for Next Iterations (Not This PR)

- Orchestration: Automatic deployment of the binary

## How to Test

```sh
./pepsi build crates/zenoh_bridge
rsync target/aarch64-unknown-linux-gnu/debug/zenoh_bridge <booster-ip>:
# On the robot
./zenoh_bridge
```

Then, listen to the available zenoh topics, e.g. using zenoh-cli.

PR #2182, opened 2026-01-13
`hulk-booster`

### LINKED ISSUE #2103: Use Boostercamera with Framework
The first step of adapting vision would be to use our old vision with the new Camera. Probably one wants to start with only one camera.

### LINKED ISSUE #2107: Implement hardware interface for real booster hardware
Implement the hardware interface for real booster hardware using DDS.

alexschmander:
## Why? What?

This PR adds a `booster-hulk` crate. This builds the binary, that will run on the robot. The scope for this PR is adding the hardware interface, such that all relevant data from the robot is available to the robot, and the low command can be sent back to the robot. 
The hardware interface uses zenoh to subscribe and publish robot data. For the robots topics to be available in zenoh, the booster zenoh bridge from #2177 needs to be run on the robot.  

Fixes #2107
Fixes #2103

## ToDo / Known Issues

- [x] ~~The motor states are need to be handled better. The simulator sends the serial ones, while the robot sends the parallel ones.~~
- [x] ~~We should figure out, which of the image topics is the best to use for now. Using multiple, at least if we receive them in the same cycler, makes everything very slow. Or making the receiving of all topics fast works too :D~~ Every read on the hardware interface, that should not set the frequency of the cycler, is now moved into its own cycler. This makes these hardware interface calls asynchronous and not block each other. 
- [x] The twix image panel currently has hardcoded image dimensions. Making it dynamically use the dimensions from the image proved harder than I thought. Especially, since the `JpegImage` does not contain its dimensions. Tracked in #2202
- [ ] It seems, that if certain data is not available via the Zenoh topics, the framework tries to shut down, but some blocking function does not terminate (perhaps not cancelled via the token) and the framework thus does not return with a useful error. 

## How to Test

- On the robot: `~/zenoh_bridge`
- Upload the hulk_booster binary: `./pepsi upload 192.168.10.102 --skip-os-check`
- The values of the hardware interface, that are output in the robotics code as a MainOutput, should be visible in twix, when connecting to `192.168.10.102` and having a network route to the robot.

### DIFF
--- a/crates/hulk_manifest/src/lib.rs
+++ b/crates/hulk_manifest/src/lib.rs
@@ -129,16 +129,7 @@ pub fn collect_hulk_cyclers(root: impl AsRef<Path>) -> Result<Cyclers, Error> {
             CyclerManifest {
                 name: "Image",
                 kind: CyclerKind::Perception,
-                instances: vec![
-                    "Rectified",
-                    "RightRectified",
-                    "RightRaw",
-                    "CombinedRaw",
-                    "StereonetDepth",
-                    "StereonetVisual",
-                    "OriginLeft",
-                    "OriginRight",
-                ],
+                instances: vec!["Rectified", "StereonetDepth"],
                 setup_nodes: vec!["sensor_receiver::image_receiver"],
                 nodes: vec![],
                 execution_time_warning_threshold: None,
--- a/etc/parameters/default.json
+++ b/etc/parameters/default.json
@@ -377,12 +377,6 @@
   },
   "image_receiver": {
     "image_rectified": "Rectified",
-    "image_right_rectified": "RightRectified",
-    "image_right_raw": "RightRaw",
-    "image_combined_raw": "CombinedRaw",
-    "image_stereonet_depth": "StereonetDepth",
-    "image_stereonet_visual": "StereonetVisual",
-    "image_origin_left": "OriginLeft",
-    "image_origin_right": "OriginRight"
+    "image_stereonet_depth": "StereonetDepth"
   }
 }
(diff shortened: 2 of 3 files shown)


crates/booster/src/lib.rs
```diff
@@ -18,10 +19,10 @@ use pyo3::prelude::*;
 pub struct LowState {
     /// IMU feedback
     pub imu_state: ImuState,
+    /// Serial structure joint feedback
+    pub motor_state_serial: [MotorState; 22],
     /// Parallel structure joint feedback
     pub motor_state_parallel: Vec<MotorState>,
```
oleflb: also the type between serial and parallel is inconsistent

crates/hulk_booster/src/hardware_interface.rs
```diff
@@ -0,0 +1,495 @@
...
+
+use crate::HardwareInterface;
+
+#[repr(C)]
+#[derive(Deserialize)]
+struct DDSDataWrapper<'a> {
+    #[serde(with = "serde_bytes")]
+    representation_identifier: [u8; 2],
+    #[serde(with = "serde_bytes")]
+    representation_options: [u8; 2],
```
oleflb: I think using `serde_bytes` for the header (4 bytes) is a bit overkill. Copying is totally fine here

crates/hulk_booster/src/hardware_interface.rs
```diff
@@ -0,0 +1,495 @@
...
+    fn from_bytes(bytes: &'a [u8]) -> Self {
+        Self {
+            representation_identifier: bytes[0..2].try_into().unwrap(),
+            representation_options: bytes[2..4].try_into().unwrap(),
+            bytes: &bytes[4..],
+        }
+    }
+
+    fn to_bytes(&self) -> Bytes {
+        let mut bytes = Vec::new();
```
oleflb: Use `Vec::with_capacity`

crates/booster/src/lib.rs
alexschmander: Is not consistent and both serial and parallel motor states are received correctly, with the current zenoh bridge that does the additional serde.

PR #2212, opened 2026-01-29
High level interface

### LINKED ISSUE #2117: High level mode switching
We need to connect the high level interface of the booster to switch the robot into Custom mode, in which we can talk to the robot using the lowlevel interface and control joints.

julianschuler:
## Why? What?

This PR implements communicating with the high-level Booster interface without using the Booster SDK.

Fixes #2117.

Blocked on #2177.

## ToDo / Known Issues

None.

Expose more of the high-level booster interface as needed.

## How to Test

Compile, upload and restart the Zenoh bridge.

Create a request and publish it via Zenoh to `booster/high_level_interface`, e.g. using the following code:

```rust
use cdr::{CdrLe, Infinite};
use high_level_interface::{Mode, Request};
use zenoh::Config;

#[tokio::main]
async fn main() {
    let session = zenoh::open(Config::default()).await.unwrap();
    let publisher = session
        .declare_publisher("booster/high_level_interface")
        .await
        .unwrap();

    let request = Request::change_mode(Mode::Custom);
    let serialized_request = cdr::serialize::<_, _, CdrLe>(&request, Infinite).unwrap();
    publisher.put(serialized_request).await.unwrap()
}
```

### DIFF
--- a/crates/zenoh_bridge/Cargo.toml
+++ b/crates/zenoh_bridge/Cargo.toml
@@ -14,6 +14,7 @@ cdr = { workspace = true }
 color-eyre = { workspace = true }
 env_logger = { workspace = true }
 futures-util = { workspace = true }
+high_level_interface = { workspace = true }
 ros2 = { workspace = true }
 ros2-client = { workspace = true }
 serde = { workspace = true }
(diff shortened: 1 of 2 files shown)


crates/high_level_interface/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "high_level_interface"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+
+[package.metadata.pepsi]
+cross-compile = true
```
knoellle: Isn't this only needed for binary crates?

PR #2214, opened 2026-02-04
change simultor teamcom port

pejotejo:
## Why? What?
Now we can start the simulator and the gamecontroller on th

Fixes #

## ToDo / Known Issues
gamecontroller does not count down message count

## How to Test

start gc and simulator on the  same devices


BenSampaolo (approved):
Works on my machine

PR #2217, opened 2026-02-18
reintroduce node.rs

pejotejo:
## Why? What?

reintroduces node.rs and moves walk to ball into action

## ToDo / Known Issues

reintroduce actions

## How to Test

test the walk ball example, it should still work


alexschmander (approved):
LGTM!

PR #2218, opened 2026-02-20
Use correct mujoco camera fovy for camera info

alexschmander:
## Why? What?

What the title says. Previously, we used some different fovy, which, after changing to the newer fov, cause the camera projection to be completely off.

## ToDo / Known Issues

## How to Test

- Setup simulator. Use `parameters.common_motor_command.default_positions.head` to make the robot look around. The ball measurement in the map panel should remain at the same place.


BenSampaolo (changes requested):
Mini cleanup :)

PR #2222, opened 2026-02-28
Add safe mode handler

alexschmander:
## Why? What?

This PR adds a node to detect, whether it is safe to leave the `PrimaryState::Safe`. This determined by checking if the "Stand" button has been long pressed and the robots joints, linear acceleration and angular velocity is very close to that of the known prep/stand mode. The output of this node is simply a bool. This is to be used in the `world_state::primary_state_filter` as guard for any transition out of the `PrimaryState::Safe

## ToDo / Known Issues

## How to Test

- `pepsi upload <robot number> --skip-os-version`
- In twix subscribe `WorldState.main_outputs.safe_to_leave_safe_mode` either in the text or enum plot panel.
- Press the stand button once to make it stand, then long press the stand button, and see if the bool is true for a short moment.

### DIFF
--- a/crates/world_state/src/safe_mode_handler.rs
+++ b/crates/world_state/src/safe_mode_handler.rs
@@ -38,7 +38,7 @@ pub struct CycleContext {
 #[context]
 #[derive(Default)]
 pub struct MainOutputs {
-    pub safe_to_leave_safe_mode: MainOutput<bool>,
+    pub should_exit_safe_mode: MainOutput<bool>,
 }
 
 impl SafeModeHandler {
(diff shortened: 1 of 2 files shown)


crates/world_state/src/safe_mode_handler.rs
```diff
@@ -0,0 +1,147 @@
...
+                    .velocity
+                    .abs_diff_eq(&safe_motor_state.velocity, *joint_velocity_threshold)
+        })
+}
+
+fn imu_state_is_safe(
+    imu_state: &ImuState,
+    prep_mode_imu_state: &ImuState,
+    angular_velocity_threshold: &f32,
+    linear_acceleration_threshold: &f32,
```
knoellle: I would at least pass the f32s by value, maybe the Joints too.
Same above

knoellle (approved):
LGTM

PR #2224, opened 2026-02-28
Remove penalty_shot_direction from ball state

rmburg:
We're going to need a ball state later, but we're not going to implement penalty shot behavior soon. Removing this field makes it easier to implement the ball state composer.


knoellle (approved):
LGTM

PR #2226, opened 2026-02-28
Add team ball receiver

rmburg:
Not much to see here. Needed for the ball state composer


knoellle (dismissed):
SGZM aber dieser Zweig hat Konflikte.

PR #2231, opened 2026-02-28
Gammaray

knoellle:
## Why? What?

This PR allows "flashing" a new booster robot with our environment.
Unlike on the NAO, this does not actually install a new OS but only places some files and installs packages via apt.

Actions already implemented:

- changing sudoers config to allow sudo without password entry
- building and uploading our own zenoh bridge
- installing ros2dds zenoh bridge
- installing systemd service files for `hulk`, `zenoh-bridge`, `zenoh-bridge-ros2dds`
- reloading systemd user service files
- installing `hulk`, `launchHULK`
- installing `podman`
- enabling & starting services: `hulk`, `zenoh-bridge`, `zenoh-bridge-ros2dds`.
  - hulk will fail on a fresh robot, but that doesn't matter

Actions planned but not yet implemented:

- Setting up networking (static 10.1.24.XX IP, etc.)
- Installing podman image for running hulk service
- ...? I'm sure I forgot some things

Fixes #

## ToDo / Known Issues

## How to Test

`./pepsi gammaray <IP>`

### DIFF
--- a/tools/k1-setup/launchHULK
+++ b/tools/k1-setup/launchHULK
@@ -11,6 +11,7 @@ ln -sf ${INTERNAL_LOG_PATH}/hulk.err ${INTERNAL_LOG_PATH}/../hulk.err
 HOST_TEGRA_LIB_PATH="/usr/lib/aarch64-linux-gnu/tegra"
 DEVICES="--device nvidia.com/gpu=all"
 
+sudo podman rm hulk
 sudo podman run -it --rm \
   --privileged \
   --network=host \


tools/k1-setup/zenoh-bridge.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=HULK Service
```
alexschmander: Same as above

tools/k1-setup/launchHULK
```diff
@@ -0,0 +1,30 @@
...
+INTERNAL_LOG_PATH=/home/booster/hulk/logs/${CURRENT_DATE}
+
+mkdir -p ${INTERNAL_LOG_PATH}
+
+ln -sf ${INTERNAL_LOG_PATH}/hulk.out ${INTERNAL_LOG_PATH}/../hulk.out
+ln -sf ${INTERNAL_LOG_PATH}/hulk.err ${INTERNAL_LOG_PATH}/../hulk.err
+
+HOST_TEGRA_LIB_PATH="/usr/lib/aarch64-linux-gnu/tegra"
+DEVICES="--device nvidia.com/gpu=all"
+
```
alexschmander: I think we still need this:
`sudo podman rm hulk`

PR #2232, opened 2026-02-28
ort dynamic linking

knoellle:
## Why? What?

Link ort dynamically and cache tensor-rt artifacts
Should only apply when building for booster, mujoco should be untouched.

Fixes #

## ToDo / Known Issues

## How to Test

Run robotics code on the robot and in mujoco, see if the GPU is used on the robot.


alexschmander (approved):
LGTM

PR #2235, opened 2026-03-01
Gammaray, die Zweite

knoellle:
## Why? What?

Follow up for #2231

Lots of cleanup and fixes #2234

## Ideas for Next Iterations (Not This PR)

Upload and unpack podman image

## How to Test

The bugs in #2234 should be fixed, the rest is just refactoring.

PR #2237, opened 2026-03-02
Gammaray Podman

knoellle:
## Why? What?

- Based on #2235
 - Follow up for #2231

Adds an optional argument to gammaray to upload and import podman images.
This is needed to run `hulk` with GPU acceleration, see #2232.

## ToDo / Known Issues

 - [x] ~~Decide if .tar file should be deleted after upload~~. Decided against for now.

Configure static network addresses for the robot.

## How to Test

Run `./pepsi gammaray --image <path to rust-trt-inference-image.tar> <IP>` on a fresh robot, then run `./pepsi upload <IP>`.
With this PR, `hulk` should now be able to launch successfully.

### DIFF
--- a/tools/pepsi/src/gammaray.rs
+++ b/tools/pepsi/src/gammaray.rs
@@ -147,7 +147,7 @@ async fn gammaray_robot(
         .await?;
 
     if let Some(image_file) = image_file {
-        const REMOTE_IMAGE_PATH: &str = "/home/booster/.cache/hulk/inference-runtime.tar";
+        const REMOTE_IMAGE_PATH: &str = "/home/booster/.cache/hulk/runtime-container-image.tar";
         robot
             .rsync_with_robot()?
             .arg(image_file)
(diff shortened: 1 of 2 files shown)


tools/pepsi/src/gammaray.rs
```diff
@@ -126,6 +140,26 @@ async fn gammaray_robot(
...
+    if let Some(image_file) = image_file {
+        robot
+            .rsync_with_robot()?
+            .arg(image_file)
+            .arg(format!("{}:rust-trt-inference-image.tar", robot.address))
+            .rsync_with_log("uploading podman image", &progress_bar)
+            .await?;
+        robot
+            .ssh_to_robot()?
+            .arg("sudo podman load -i rust-trt-inference-image.tar")
```
alexschmander: Do we maybe want to tag this image with a custom name like `hulk-runtime-environment` and then also use this in launchHULK? Then we can upload different images, without touching the deployed files.

knoellle: The name/tag solely depends on how the archive was created.
As I understand, the load command can be used to import any number of images and tags, d!pending on whats inside the archive file.

knoellle: As discussed, this is more complicated.
Only renamed the file and moved it to `~/.cache/hulk/` for now.

PR #2240, opened 2026-03-03
reintroduce some actions

pejotejo:
## Why? What?
Reintroduces Initial, Penalized, FinishPose, StandUp, LookAround Actions and their Executions. Also add a Safe Action. Reintroduces world-state composer

## ToDo / Known Issues
more actions


alexschmander (changes requested):
A couple of comments :)

PR #2241, opened 2026-03-03
Reintroduce roles

JackyBloxx:
## Why? What?

Reintroduce Role assignment back 

Fixes #

## ToDo / Known Issues

The  the return message to the game controller need to be tested
and a proper time_to_reach_kick_position need to be introduced

## How to Test

start simulater and see if the robot have a role and change force role to a possible role and see if it changes

### DIFF
--- a/crates/world_state/src/role_assignment.rs
+++ b/crates/world_state/src/role_assignment.rs
@@ -763,7 +763,7 @@ fn keep_current_role_during_free_kicks(
     current_role: Role,
 ) -> Option<Role> {
     if let Some(FilteredGameControllerState {
-        sub_state: Some(SubState::KickIn | SubState::PushingFreeKick),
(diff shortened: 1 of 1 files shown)


crates/world_state/src/role_assignment.rs
```diff
@@ -900,81 +889,3 @@ fn pick_keeper_or_searcher(
...
-#[cfg(test)]
-mod test {
-    use proptest::prelude::*;
-
-    use super::*;
-
-    proptest! {
-        #[allow(clippy::too_many_arguments)]
-        #[test]
-        fn process_role_state_machine_should_be_idempotent_with_event_none(
```
knoellle: I think we should keep this test. Just delete all of the values that don't exist anymore.

knoellle (dismissed):
LGTM

PR #2243, opened 2026-03-03
Reintegrate look around node

BenSampaolo:
## Why? What?

This PR reintegrates the look around node which handles the look_around and the search_for_lost_ball head motions. 

## How to Test

Run the Twix simulator and position the robot in a way that the ball is not in the FOV of the robot. Make sure that the `parameters.behavior.remote_control.enable` parameter is set to `false`. 

The robot should start to move its head from side to side. 

Also you can open an Enum plot and subscribe to `Motion.additional_outputs.look_around_mode` there you see more detailed info to the current motion.


alexschmander (approved):
LGTM

PR #2247, opened 2026-03-04
Long press `Stand` framework entry

alexschmander:
## Why? What?

We need a way of interacting with the robot to safely start the logic of our robotics code, similar to pressing the chest button on the NAO. My approach for this is reusing the Stand button on the K1, after it was already pressed to essentially make pressing the button a noop for the booster software. This PR adds renames the already existing SafeModeHandler and renames it to ButtonHandler, to emit more than one Button logic. 
If the Stand Button is long pressed and released while the robot is in the known safe stand or prepare pose, the ButtonHandler emits a `Buttons::IsStandLongPressedDuringSafePose`, which is then used to transition from `PrimaryState::Safe` to `PrimaryState::Initial` essentially starting the robotics code logic. 
If the robot is in any other primary state and at any time the Stand or F1 Button is pressed, the primary state is reset to `PrimaryState::Safe`. Additionally, for debugging and testing purposes, if the `Buttons::IsStandLongPressedDuringSafePose` is again emitted while in `PrimayState::Initial`, a transition to `PrimaryState::Playing` happens.

## ToDo / Known Issues

- In hulk-mujoco, since a `Buttons::IsStandLongPressedDuringSafePose` is injected into the `world_state::primary_state_filter.rs`, the primary state transitions first to Initial and then directly to Playing. This makes testing game controller behavior difficult and this should not happen. 

## How to Test

- On the robot, upload and then press the Stand Button, while already Standing. Look at the `primary_state` or the `buttons` main outputs in twix.

### DIFF
--- a/crates/world_state/src/button_event_handler.rs
+++ b/crates/world_state/src/button_event_handler.rs
@@ -58,15 +58,14 @@ impl ButtonEventHandler {
         for (time, button_event_messages) in all_button_event_messages {
             self.most_recently_processed_button_event_message_time = time;
 
-            button_event_messages.into_iter().flatten().for_each(
-                |ButtonEventMsg { button, event }| {
-                    buttons[*button] = ButtonPressType::from_button_event_types(
-                        &self.last_button_event_types[*button],
-                        event,
-                    );
-                    self.last_button_event_types[*button] = Some(*event);
-                },
-            );
+            for button_event_message in button_event_messages.into_iter().flatten() {
+                buttons[button_event_message.button] = ButtonPressType::from_button_event_types(
+                    &self.last_button_event_types[button_event_message.button],
+                    &button_event_message.event,
+                );
+                self.last_button_event_types[button_event_message.button] =
+                    Some(button_event_message.event);
+            }
         }
 
         Ok(MainOutputs {


crates/world_state/src/button_event_handler.rs
```diff
@@ -96,13 +96,40 @@ impl SafeModeHandler {
...
+                imu_state.angular_velocity - context.prep_mode_imu_state.angular_velocity
+            });
+
+        let Some(button_event) = context
+            .maybe_button_event
+            .persistent
+            .into_iter()
+            .flat_map(|(_time, info)| info)
+            .last()
+            .flatten()
```
knoellle: As discussed, please also use temporaries and deduplicate by timestamp.
Then run the state machine for each event and keep last_event separate for each button.

crates/world_state/src/button_event_handler.rs
```diff
@@ -118,11 +145,50 @@ impl SafeModeHandler {
...
+                _,
+            ) => Some(Buttons::IsStandOrF1Pressed),
+            _ => None,
+        };
+
+        self.last_button_event = Some(button_event.clone());
 
         Ok(MainOutputs {
-            should_exit_safe_mode: should_exit_safe_mode.into(),
+            buttons: buttons.into(),
```
knoellle: As discussed, change this type to have separate fields for each button that indicate if there was a press (on button release) and if it was a long press.

PR #2249, opened 2026-03-04
Pepsi Wifi

knoellle:
## Why? What?

- Based on #2238 
This PR changes all the `pepsi wifi ...` commands to work with the new network names and to use nmcli instead of iwctl.

## How to Test

Run the `pepsi wifi` commands

### DIFF
--- a/crates/bevy_mujoco/src/lib.rs
+++ b/crates/bevy_mujoco/src/lib.rs
@@ -139,13 +139,12 @@ fn calculate_tangents(mesh: &SceneMesh) -> Vec<[f32; 4]> {
     let tangents_attribute = helper_mesh
         .remove_attribute(Mesh::ATTRIBUTE_TANGENT)
         .expect("we calculated these earlier");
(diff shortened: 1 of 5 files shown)


crates/robot/src/lib.rs
```diff
@@ -349,16 +349,16 @@ impl Robot {
...
-            .arg("station")
-            .arg("wlan0")
+            .arg("nmcli")
+            .arg("device")
+            .arg("wifi")
             .arg("show")
             .output()
             .await
-            .wrap_err("failed to execute iwctl ssh command")?;
+            .wrap_err("failed to execute nmcli ssh command")?;
```
alexschmander: The output formatting here is also a bit weird. See output: 
```
[10.1.24.42] ✔ SSID: <ssid>
Security: <security>

                                                                                                               
```
I think we should either grep the SSID value and only print that, or remove new lines and replace them with a space or something similar.

alexschmander (approved):
LGTM

PR #2252, opened 2026-03-05
High Level Interface for booster motion

alexschmander:
## Why? What?

This PR adds the some functions of the booster high level interface by way of
Team WhIRLwinds `booster_sdk` into the hardware interface.
To use the booster high level motion, while also allowing for our own motion inference during simulation with `hulk-mujoco`, a `MotionRuntimeInterface` is added for motion nodes of either motion type to only execute when appropriate.
In mujoco motion remains as it is on main. In `hulk-booster`, the `walking.rs` executes walking motions using the high level interface call `move_robot()`. `rotate_head.rs` allows for head control using the high level interface.

Depends on #2247

## ToDo / Known Issues

- There is some weird generated code clippy warning about unneccessary unwraps in replayer and imagine.

## How to Test

- Upload on a robot. Make it Stand and then Long press Stand twice, to transition the robot into `PrimaryState::Playing`. By default, currently the remote control is enabled. You can either control it through the twix remote control panel, or by turning remote control off. Then the robot will chase detected balls or Stand with QuickSearch.

### DIFF
--- a/crates/world_state/src/robot_mode_handler.rs
+++ b/crates/world_state/src/robot_mode_handler.rs
@@ -35,7 +35,7 @@ impl BoosterModeHandler {
         &mut self,
         context: CycleContext<impl HighLevelInterface + MotionRuntimeInteface + TimeInterface>,
     ) -> Result<MainOutputs> {
-        if context.hardware_interface.get_motion_runtime_type()? == MotionRuntime::Booster {
+        if context.hardware_interface.get_motion_runtime_type()? != MotionRuntime::Booster {
             return Ok(MainOutputs {
                 robot_mode: None.into(),
             });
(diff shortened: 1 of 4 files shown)


crates/world_state/src/robot_mode_handler.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    pub fn cycle(
+        &mut self,
+        context: CycleContext<impl HighLevelInterface + MotionRuntimeInteface + TimeInterface>,
+    ) -> Result<MainOutputs> {
+        if !matches!(
+            context.hardware_interface.get_motion_runtime_type()?,
+            MotionRuntime::Booster
+        ) {
```
knoellle: This would be improved by summoning more `Eq`uines.

PR #2264, opened 2026-03-07
Don't run booster jobs on borked CI runners

knoellle:
## Why? What?

For some reason, building with the booster container fails on CI runner 1 and 3. No idea why.
Instead of disabling them outright, this PR just prevents these jobs from running on these specific runners but they can still do the other jobs.

Fixes #

Find out why CI is bronek in the first place

## How to Test

See if CI runs correctly


knoellle:
Did not fix the issue, need to investigate further

PR #2265, opened 2026-03-07
Booster visual kick

alexschmander:
## Why? What?

This PR adds support for the booster visual kick. This is a kicking motion policy, that received a ball position, a target kick , a kick direction angle, current orientation angle to the field, and a kick power. The policy then does full motion walking towards the ball and kicks it in the direction of the target. 

Since this PR changes the Zenoh Bridge, after merging, all robots need to be gammarayed. 

## ToDo / Known Issues

## How to Test

- Upload on the robot. Either insert the Action::VisualKick in behavior into the actions executed in Playing, or inject the `MotionCommand::VisualKick` via `injected_motion_command`.

### DIFF
--- a/crates/world_state/src/behavior/node.rs
+++ b/crates/world_state/src/behavior/node.rs
@@ -10,7 +10,8 @@ use types::{
 };
 
 use crate::behavior::{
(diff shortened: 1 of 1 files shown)


alexschmander:
Konflikte wurden behoben. Die Ziel Koordinaten wurden invertiert.

PR #2269, opened 2026-03-08
Gammaray skip apt

knoellle:
## Why? What?

Gammaray is relatively fast except for the apt update && apt install steps.
This PR makes it skip these steps if the packages we want to install are already present.

Fixes #

## How to Test

Run gammaray on an already gammarayed robot. Should be much faster.


alexschmander (approved):
LGTM

PR #2276, opened 2026-03-08
Add node to perform zero-order hold for world_state crate

rmburg:
This is not super pretty and will be obsolete with the new framework, but it deduplicates and simplifies some nodes for now.

Open question: maybe these outputs should have a different name from the originals?


paudar:
no longer desired

PR #2293, opened 2026-03-10
Reintroduce vision cycler

rmburg:
Reintroduce the old vision crate, minus the unused bits (ball/feet detection, perspective grid).

To review, I recommend diffing with an old main (@ robocup25) worktree, like so:
```
difft --color always {../hulk-main/,}crates/vision/ | less -R
```

To Do:
- [ ] Fix deadlock in simulator


tools/twix/src/panels/image/mod.rs
rmburg: maybe these changes are not necessary at all, I only needed them for testing.

PR #2303, opened 2026-03-10
Gammaray gateway

knoellle:
## Why? What?

Set a getaway gateway so we have internet for the apt steps.

- based on #2299 

Fixes #

## How to Test

Gamer ray


alexschmander (approved):
LGTM

PR #2312, opened 2026-03-10
Rust tensorrt warmup script

oleflb:
## Why? What?

Creates a script to initialize the tensorrt engine cache without running the framework.

Fixes #

## ToDo / Known Issues

## How to Test

- Only possible on the robot by running it on the yolo onnx model

### DIFF
--- a/tools/tensorrt-compile/src/main.rs
+++ b/tools/tensorrt-compile/src/main.rs
@@ -21,7 +21,7 @@ struct CliArguments {
 }
 
 fn main() -> Result<()> {
-    const IMAGE_WIDTH: usize = 554;
+    const IMAGE_WIDTH: usize = 544;
     const IMAGE_HEIGHT: usize = 448;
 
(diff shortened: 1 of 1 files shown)


tools/tensorrt-compile/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// Path to onnx model
+    onnx_path: PathBuf,
+
+    /// Path to cache folder
+    #[arg(long, default_value = "/home/booster/.cache/hulk/tensor-rt")]
+    cache_path: PathBuf,
+}
+
+fn main() -> Result<()> {
+    const IMAGE_WIDTH: usize = 554;
```
alexschmander:
```suggestion
    const IMAGE_WIDTH: usize = 544;
```

alexschmander (changes requested):

alexschmander (approved):
LGTM

PR #2336, opened 2026-03-11
lets make Behavior not WIP again

pejotejo:
## Why? What?

Combines the Actions of #2288 and #2307 with the rolematching of  #2285 and a lot of fixes that make it work. With this  basically our whole old behavior is back main. (with some kicks and corners) 

## How to Test
probably some test games


alexschmander (approved):
LGTM

PR #2358, opened 2026-03-12
Use Ricos alternative image provider for X5 board

oleflb:
## Why? What?

Uses custom firmware for the head provided by Rico from Booster.
This allows to receive images with buttery-smooth 60fps without problems.

After merging this PR we need to reflash all heads. If we want to revert this PR, we have to reflash again.
We should make sure the implementation is correct.

Fixes #

## ToDo / Known Issues

## How to Test

Subscribe to images in twix (currently only 44 has the correct head firmware). Enjoy buttery-smooth low-latency images

### DIFF
--- a/crates/hulk_booster/src/x5_receiver/receiver.rs
+++ b/crates/hulk_booster/src/x5_receiver/receiver.rs
@@ -152,20 +152,11 @@ impl X5ReceiverTask {
         "A little-endian target architecture is required because the network byte stream represents memory dumps from a little-endian X5 board."
     );
     async fn receive_frame(&mut self) -> io::Result<X5CameraFrame> {
-        let mut header_bytes = [0u8; 21];
+        const { assert!(size_of::<X5CameraFrameHeader>() == 21) }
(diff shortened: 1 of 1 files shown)


crates/sensor_receiver/src/image_receiver.rs
alexschmander: I think you can remove this whole node and cycler. Since we dont actually receive the rectified image. This was previously only used to debug and compare different images. We now know which images we want, and thus we do not need these Image receiver Cyclers anymore.

crates/types/src/parameters.rs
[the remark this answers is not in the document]
oleflb: done

crates/sensor_receiver/src/image_receiver.rs
oleflb: done

oleflb:
> I'm a bit sad, that this is not using zenoh yet ;)

I can understand that. I think a better solution for the future is letting a zenoh node run in the head. But we should check with booster what their plans are for the future of the software on the head

PR #2359, opened 2026-03-12
Cubature KF

schmidma:
## Why? What?

Fixes some bugs in the pose filter (localization)
https://kalman-filter.com/cubature-kalman-filter/
https://groups.seas.harvard.edu/courses/cs281/papers/unscented.pdf

## ToDo / Known Issues

none

none

## How to Test

loca on the field?

### DIFF
--- a/crates/filtering/src/pose_filter.rs
+++ b/crates/filtering/src/pose_filter.rs
@@ -373,7 +373,7 @@ mod tests {
     use std::f32::consts::PI;
 
     use approx::assert_relative_eq;
-    use nalgebra::{matrix, vector};
+    use nalgebra::{Matrix3, matrix, vector};
 
     use super::{CUBATURE_POINT_COUNT, PoseFilter, sample_cubature_points, wrap_angle};
     use types::multivariate_normal_distribution::MultivariateNormalDistribution;
@@ -390,12 +390,7 @@ mod tests {
             covariance: initial_covariance,
         };
 
-        state
-            .predict(
-                |state| state,
-                matrix![0.0, 0.0, 0.0; 0.0, 0.0, 0.0; 0.0, 0.0, 0.0],
-            )
-            .unwrap();
+        state.predict(|state| state, Matrix3::zeros()).unwrap();
 
         assert_relative_eq!(state.covariance, initial_covariance, epsilon = 1.0e-5);
     }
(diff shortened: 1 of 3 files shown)


crates/filtering/src/pose_filter.rs
```diff
@@ -303,5 +365,196 @@ fn cross_covariance_from_2d_translation_sigma_points(
...
+        ];
+        let mut state = MultivariateNormalDistribution {
+            mean: vector![1.0, -0.5, 0.3],
+            covariance: initial_covariance,
+        };
+
+        state
+            .predict(
+                |state| state,
+                matrix![0.0, 0.0, 0.0; 0.0, 0.0, 0.0; 0.0, 0.0, 0.0],
```
oleflb: use `Matrix3::zeros()` here

crates/world_state/src/localization.rs
```diff
@@ -911,6 +909,29 @@ fn predict(
...
+    let translation_process_noise = rotation_to_field.matrix()
+        * Matrix2::from_diagonal(&translation_noise_in_odometry_frame)
+        * rotation_to_field.matrix().transpose();
+
+    let mut process_noise = Matrix3::zeros();
+    process_noise
+        .fixed_view_mut::<2, 2>(0, 0)
+        .copy_from(&translation_process_noise);
+    process_noise[(2, 2)] =
+        current_odometry_to_last_odometry.rotation.angle().abs() * odometry_noise.z;
```
oleflb, line 937: Why multiply with the angle here?

oleflb (changes requested):

crates/filtering/src/pose_filter.rs
schmidma: done

crates/world_state/src/localization.rs
schmidma: In the same spirit as I multiply the process noise for x/y with the amount of translation. The larger the movement of this prediction step, the larger the uncertainty of the process. Noise scales with larger movement.
Thus a multiplication of `noise[(2, 2)]` is the diagonal component of the theta orientation angle and this scales with the rotational size of odometry movement (the larger the rotation, the larger the noise)

oleflb (approved):

PR #2382, opened 2026-03-13
Reintroduce active vision

alexschmander:
## Why? What?

We currently do not look at the ball in Ready or Set. This PR reintroduces our old active vision. 

## ToDo / Known Issues

## How to Test

- Upload on a robot.
- Let it walk in as a Keeper or Striker.
- After reaching the position, it should do active vision.


pejotejo (approved):
looks even better

PR #2384, opened 2026-03-13
Replace odometer orientation with IMU yaw

knoellle:
## Why? What?

IMU is not being reset and doesn't stop updating arbitrarily. Thus, it does not suffer from the same race-condition issues with resetting and jumping as the odometry theta.

## How to Test

Check if loca jumps still when transitioning between safe and walk when the robot turned in between. It should no longer do that.


alexschmander (approved):
LGTM

PR #2422, opened 2026-04-09
Remote workspace fixes

knoellle:
## Why? What?

1. Preserve relative working directory for remote workspace
  Previously, the command would always be executed in the root of the repository. Now if you run the remote workspace script from within e.g. `tools/machine-learning/whatever`, the command will also run in this directory on the remote.. Note that this behavior can still be overriden by use of the `--cd` option.
2. Allow directories for `--return-file` option.
  Apparently rsync's `--archive` only implies `--recursive` when not using `--files-from`, so recursion needs to be enabled manually here.

          --archive, -a
              This is equivalent to -rlptgoD.  It is a quick way of saying you want recursion and want to preserve almost everything.
              [...]
              The only exception to the above equivalence is when --files-from is specified, in which case -r is not implied.
3. Use `info=progress2` for rsync when returning files.

## ToDo / Known Issues

## How to Test

1.
```sh
cd tools/machine-learning
../../scripts/remote_workspace pwd
# should print    ~/hulk/tools/machine-learning/multi-task-yolo
# was previously  ~/hulk/
```
2.
```sh
scripts/remote_workspace --return-file a mkdir -p a/b/c
tree a
# a
# └── b
#    └── c

# Previously only an empty directory `a` would be created
```
3. Look at rsync output when testing 2.

### DIFF
--- a/scripts/remote_workspace
+++ b/scripts/remote_workspace
@@ -4,6 +4,7 @@
 set -e
 
 # cd to repository base directory for better referencing
+INVOCATION_DIR=`pwd -P`
 BASEDIR=`cd $(dirname $0); pwd -P`
 BASEDIR=${BASEDIR%/*}
 cd "$BASEDIR"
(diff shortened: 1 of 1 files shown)


BenSampaolo (approved):
Even works on my machine

PR #2423, opened 2026-04-10
Multi-task detection with single task training

alexschmander:
## Why? What?

*Describe here why you created this PR and what it does*

- Depends on #2280 
- Fixes #2413 
- Fixes #2415 
- Fixes #2417 
- Fixes #2418

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

- A full guide on how to reproduce the results of the pipeline can be found under `tools/machine-learning/multi-task-yolo/REPRODUCE.md`

PR #2424, opened 2026-04-10
Hydra object + pose detection inference

alexschmander:
## Why? What?

This PR implements the inference of the hydra object + pose detection, multi-task model, trained with the pipeline from #2423.

## How to Test

- Upload to a robot. 
- Open twix.
- In the image panel, enable object and pose detection overlays.


crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+    fn aliases(self) -> &'static [&'static str] {
+        match self {
+            Self::ObjectDetection => &[
+                "network_detections_0",
+                "detection_output",
+                "network_detections",
+            ],
+            Self::PoseDetection => &["network_detections_1", "pose_output"],
+        }
+    }
```
oleflb: why do you need aliases? If really needed, prefer an associated constant here. This should also allow you to omit 'static

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    fn expected_columns(self) -> usize {
+        match self {
+            Self::ObjectDetection => DETECTION_OUTPUT_COLUMNS,
+            Self::PoseDetection => POSE_OUTPUT_COLUMNS,
```
oleflb: maybe also an assoc. const?

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+            let values = squeeze_output_tensor(name, tensor)?;
+            Ok(ModelOutput {
+                name: name.to_string(),
+                values,
+            })
+        })
+        .collect()
+}
+
+fn squeeze_output_tensor(output_name: &str, tensor: ArrayViewD<'_, f32>) -> Result<Array2<f32>> {
```
oleflb: Function name does not reflect what happens here. A squeeze should never clone data, which is what you do here.

oleflb: And even reverses an axis?

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+            } else if shape[2] == 1 {
+                tensor
+                    .index_axis(Axis(2), 0)
+                    .into_dimensionality::<Ix2>()
+                    .map(|array| array.reversed_axes().to_owned())
+                    .wrap_err_with(|| {
+                        format!(
+                            "failed to interpret output `{output_name}` trailing axis as rank-2 tensor"
+                        )
+                    })
```
oleflb: make these match arms with guards

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+            } else {
+                bail!(
+                    "unsupported shape for output `{output_name}`: expected [1, rows, cols] or [cols, rows, 1], got {shape:?}"
+                );
+            }
+        }
+        rank => {
+            bail!("unsupported rank for output `{output_name}`: expected rank 2 or 3, got {rank}")
+        }
```
oleflb: As currently written, the outputs have semantic meaning (certain axes are removed, others reversed) but that meaning cannot be inferred from reading the code. Either you write comments about the contents of the array or you need a typed representation of your outputs (which maybe also require explaining comments)

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+        if let Some(output) = outputs.iter().find(|output| output.name == *alias) {
+            ensure_output_columns(task, output)?;
+            return Ok(output.values.view());
+        }
+    }
+
+    let matching_outputs = outputs
+        .iter()
+        .filter(|output| output.values.ncols() == task.expected_columns())
+        .collect::<Vec<_>>();
```
oleflb: this seems dangerous? Why don't you find by named key?

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+fn select_task_output<'a>(
+    task: TaskHead,
+    outputs: &'a [ModelOutput],
+) -> Result<ArrayView2<'a, f32>> {
+    for alias in task.aliases() {
+        if let Some(output) = outputs.iter().find(|output| output.name == *alias) {
+            ensure_output_columns(task, output)?;
+            return Ok(output.values.view());
+        }
+    }
```
oleflb: I think you should rework your export script. This looks like madness :D

crates/hydra/src/inference.rs
```diff
@@ -0,0 +1,437 @@
...
+        .filter_map(|row| {
+            let confidence = row[4usize];
+            if confidence < confidence_threshold {
+                return None;
+            }
+
+            let class_id = row[5usize] as usize;
+            let label = RobocupObjectLabel::from_index(class_id);
+
+            Some(Object {
```
oleflb: impl From<[DType]> for Object

oleflb (changes requested):
I am not sure about cycler name. But since we won't have cyclers for much longer I won't be pedantic here

crates/hydra/src/inference.rs
alexschmander: Removed. Outputs are now indexed with a single fixed name for that task

crates/hydra/src/inference.rs
alexschmander: I have cleaned this up a lot

crates/hydra/src/inference.rs
alexschmander: Did not do this. You decide if this is really necessary

crates/hydra/src/inference.rs
alexschmander: Moved to a much simpler check of output shape

crates/hydra/src/inference.rs
alexschmander: See above

crates/hydra/src/inference.rs
alexschmander: See above

crates/hydra/src/inference.rs
alexschmander: See above

crates/hydra/src/inference.rs
alexschmander: Did this for both `Object` and `Pose`

oleflb (approved):

PR #2430, opened 2026-04-16
fix: look_at while kicking

pejotejo:
## Why? What?
In look_at we did not match the VisualKick motion, therefore never looked at the ball during kicking.  

## How to Test

Let the robot kick and check if it looks at the ball.


alexschmander (approved):
LGTM

PR #2431, opened 2026-04-16
Add seeed orin to team.toml for pepsi

alexschmander:
Adds an entry for the local Orin NX developer board to the team.toml, for pepsi integration.


knoellle (approved):
LGTM

Issue #2343, opened 2026-03-11
reintroduce intercept ball action


pejotejo:
in our old behavior we had this action, we probably want to have it again

PR #2435, opened 2026-04-18
Playing tree

pejotejo:
## Why? What?
Adds simple behavior for us to at least try to shoot goals. It simply adds the visual kick action back. But mainly it give the playing tree a simple structure.

## Ideas for Next Iterations (Not This PR)

Build the better subtrees.

## How to Test

Try to play with the robot and look at the behavior panel in twix while doing so.

PR #2436, opened 2026-04-18
Striker Subtree

pejotejo:
depends on  #2435 

## Why? What?
Adds a more complex striker tree with switching between walking and kicking and switching beween the different kicking policies. The current tree should prevent instant switches between the different policies. This should hopefully stabilize its motions. Also its adds an interceptmotion.

## ToDo / Known Issues

think about how we want to split up the kicking motion with kickingpower and target

handle headmotion seperately (currently working on it)

## How to Test

Play with the robot and look at the behavior tree.


crates/types/src/motion_type.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[default]
+    Prepare,
+    Stand,
+    StandUp,
+    Kick,
+    Walk,
+}
+
+impl MotionType {
+    pub fn is_stable(&self) -> bool {
```
paudar: unused function?

crates/world_state/src/behavior/tree.rs
```diff
@@ -29,50 +40,120 @@ pub fn create_tree() -> Node<Blackboard> {
...
+        ),
+        action!(stand)
+    )
+}
+
+fn kick_alternatives_subtree() -> Node<Blackboard> {
+    selection!(
+        sequence!(
+            condition!(is_last_motion_type, MotionType::Walk),
+            action!(set_is_alternative),
```
paudar: In my view, the `set_is_alternative` function is not relevant

crates/types/src/motion_type.rs
pejotejo: was from the old behavior, is now gone

crates/world_state/src/behavior/tree.rs
pejotejo: is removed

paudar (dismissed):
LGTM

pejotejo:
closes #2343

paudar (approved):
LGTM

PR #2437, opened 2026-04-18
Improved Behavior Tree Panel

pejotejo:
depends on
#2435 
#2436 

## Why? What?
Improves behaviortree panel in twix. Now subtrees can be collapsed and there are different views for different user preferences. Also there were minor quality improvements like animations, correct resetting and a checkbox to flip the graph.

- keyboard support
## How to Test

upload the behavior tree to the robot and look at the behaviortree panel in twix


tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        self.update_trace_strokes();
+
+        let (response, mut painter) = TwixPainter::<World>::allocate(
+            ui,
+            vector![25.0, 25.0],
+            point![0.0, 0.0],
+            Orientation::LeftHanded,
+        );
+
+        let mut focus_transform = None;
```
knoellle: Replace
```rs
let mut x = None;
if let Some(y) = z {
  x = Some(f(y));
}
```
with
```rs
let x = z.map(f);
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        if let Some(root_node) = self
+            .circle_nodes
+            .iter()
+            .find(|node| node.id == "root")
+            .or_else(|| self.circle_nodes.iter().find(|node| node.id == "cfg_root"))
```
knoellle: Why can the tree have two differently named root nodes?
When searching the repo for "cfg_root", I don't find any place where such an id is assigned.

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        {
+            let root_pixel = painter.transform_world_to_pixel(root_node.target_position);
```
knoellle:
```suggestion
            let root_in_pixel = painter.transform_world_to_pixel(root_node.target_position);
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                    VIEW_SWITCH_FOCUS_SCALE,
+                )
+                .framed_transform(),
+            );
+        }
+
+        if self.pending_view_switch_focus {
+            if let Some(transform) = focus_transform {
+                self.zoom_and_pan.transformation = transform;
+            }
```
knoellle: Instead of overwriting the struct member, can you add a function to the `ZoomAndPan` type which centers the view on a `world`-position with a given scale?

tools/twix/src/panels/image/overlays/object_detection.rs
```diff
@@ -58,7 +58,7 @@ fn paint_bounding_boxes(
         painter.floating_text(
             bounding_box.area.max,
             Align2::RIGHT_TOP,
-            detection.label.into(),
+            format!("{:?}", detection.label),
```
knoellle: Why does a label need to be Debug printed?
Why this change?

tools/twix/src/panels/behavior_tree/graph.rs
```diff
@@ -0,0 +1,205 @@
+use coordinate_systems::World;
+use eframe::egui::{Align2, Color32, FontId, Response, Stroke};
+use linear_algebra::{Point2, distance_squared};
+
+use crate::twix_painter::TwixPainter;
+
+const COLLIISION_LIMIT: usize = 50;
```
knoellle:
```suggestion
const COLLISION_LIMIT: usize = 50;
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+            for node in &mut self.circle_nodes {
+                let layout_position = node.position;
+                node.target_position = layout_position;
+
+                if let Some(old_node) = old_nodes.get(&node.id) {
+                    node.position = old_node.position;
+                    node.opacity = old_node.opacity;
+                } else {
+                    if node.id == "cfg_root" {
+                        node.position = layout_position;
```
knoellle: This is a no-op. `layout_position` already is a copy of `node.position`.

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        let mut path = Vec::new();
+        collect_statuses_by_id(&trace, &mut path, &mut statuses);
+
+        if let Some(root_status) = statuses.get("root").cloned() {
+            statuses.insert("cfg_root".to_string(), root_status);
+        }
+
+        for node in &mut self.circle_nodes {
+            if let Some(status) = statuses.get(&node.id) {
+                node.stroke = eframe::egui::Stroke::new(0.1, status_color(status));
```
knoellle:
```suggestion
                node.stroke = Stroke::new(0.1, status_color(status));
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                    ui.selectable_value(
+                        &mut self.view_mode,
+                        LayoutViewMode::Tree,
+                        "Phillips' View",
+                    );
+                    ui.selectable_value(
+                        &mut self.view_mode,
+                        LayoutViewMode::SequenceChains,
+                        "Johannes' View",
+                    );
```
knoellle: You already have a function generating labels for view modes.
```suggestion
                    for mode in [LayoutViewMode::Tree, LayoutViewMode::SequenceChains] {
                        ui.selectable_value(&mut self.view_mode, mode, mode.label());
                    }
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                .rev()
+                .find(|node| {
+                    node.is_subtree
+                        && node.contains(painter.transform_pixel_to_world(pointer_position))
+                })
+                .map(|node| node.id.clone())
+        {
+            if self.collapsed_subtrees.contains(&clicked_subtree_id) {
+                self.collapsed_subtrees.remove(&clicked_subtree_id);
+                self.opening_subtree_origin = Some(clicked_subtree_id.clone());
```
knoellle: As far as I can see, this is the only location where you set the `opening_subtree_origin` to anything other than None. This member is only read inside the `rebuild_layout` function which you call a few lines further down.
Furthermore, the `rebuild_layout` function also sets this member to None.

My question is: why is this a member and not a parameter to the `rebuild_layout` function?

tools/twix/src/panels/behavior_tree/graph.rs
```diff
@@ -0,0 +1,205 @@
...
+        let mut did_resolve_any = false;
+
+        for i in 0..nodes.len() {
+            for j in (i + 1)..nodes.len() {
+                if !nodes[i].is_visible || !nodes[j].is_visible {
+                    continue;
+                }
+
+                let to_b = nodes[j].position - nodes[i].position;
+                let distance_squared = to_b.inner.norm_squared();
```
knoellle: Our `linear_algebra::Vector2` also implements `norm_squared`.
```suggestion
                let distance_squared = to_b.norm_squared();
```

tools/twix/src/panels/behavior_tree/graph.rs
```diff
@@ -0,0 +1,205 @@
...
+                let radius = nodes[i].radius + nodes[j].radius;
+                let minimal_distance = radius + nodes[i].stroke.width + nodes[j].stroke.width;
```
knoellle: This does not consider the slightly larger radius of the nodes with an additional subtree circle.
Is that intentional?

tools/twix/src/panels/behavior_tree/graph.rs
```diff
@@ -0,0 +1,205 @@
...
+
+                    if a_dragging && !b_dragging {
+                        nodes[j].position += direction * overlap;
+                    } else if !a_dragging && b_dragging {
+                        nodes[i].position -= direction * overlap;
+                    } else if !a_dragging && !b_dragging {
+                        let half_overlap = overlap * 0.5;
+                        nodes[i].position -= direction * half_overlap;
+                        nodes[j].position += direction * half_overlap;
+                    }
```
knoellle, line 183: You set the position here not the target position, causing the nodes to animate back towards the colliding position over and over again.

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+            } else {
+                node.position = node.target_position;
+            }
+
+            node.opacity = (node.opacity + ENTER_FADE_STEP).min(1.0);
+        }
+
+        if !self.exiting_nodes.is_empty() {
+            let mut remaining_exiting_nodes = Vec::with_capacity(self.exiting_nodes.len());
+            for mut node in self.exiting_nodes.drain(..) {
```
knoellle: Consider using [`retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) instead of creating a new vector.

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                node.position += delta * LAYOUT_ANIMATION_FACTOR;
+                any_animating = true;
+            } else {
+                node.position = node.target_position;
+            }
+
+            node.opacity = (node.opacity + ENTER_FADE_STEP).min(1.0);
+        }
+
+        if !self.exiting_nodes.is_empty() {
```
knoellle: This check is unnecessary.

tools/twix/src/panels/behavior_tree/model.rs
```diff
@@ -0,0 +1,193 @@
...
+    desired_names: &HashSet<&str>,
+    path: &mut Vec<usize>,
+    collapsed_ids: &mut HashSet<String>,
+) {
+    if node_trace.name.starts_with(SUBTREE_PREFIX) {
+        let raw_name = node_trace
+            .name
+            .strip_prefix(SUBTREE_PREFIX)
+            .unwrap_or(node_trace.name.as_str());
+        let subtree_name = raw_name.split(':').next().unwrap_or(raw_name);
```
knoellle: Prefer `split_once`
```suggestion
        let subtree_name = raw_name.split_once(':').unwrap_or(raw_name);
```

tools/twix/src/panels/behavior_tree/control_flow_layout.rs
```diff
@@ -0,0 +1,286 @@
...
+        ControlFlowNodeKind, control_flow_node_kind, display_name_for_node, node_id_from_path,
+        normalized_name_and_subtree_flag,
+    },
+};
+
+const NODE_RADIUS: f32 = 2.0;
+const CONTROL_FLOW_X_SPACING: f32 = 4.4;
+const CONTROL_FLOW_Y_SPACING: f32 = 7.0;
+const CONTROL_FLOW_BRANCH_GAP: f32 = 0.6;
+const HIDE_SEQUENCE_NODES_IN_CONTROL_FLOW_VIEW: bool = true;
```
knoellle: Is this something you intend to change at some point?
It seems weird to me to have a const parameter like this. I think you should commit to either variation or make it actually switchable.

knoellle (changes requested):
Some preliminary comments

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: was a relict when i tried to remove the selection nodes, then the root node would have been gone, but this led to other problems, so i reverted it and must have mist it.  is removed

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/image/overlays/object_detection.rs
pejotejo: was some llm mistake, changed it back

tools/twix/src/panels/behavior_tree/graph.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: removed it

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: sorry was a relict from when i tried something out, it's removed now

tools/twix/src/panels/behavior_tree/graph.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/graph.rs
pejotejo: no, is fixed

tools/twix/src/panels/behavior_tree/graph.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/mod.rs
pejotejo: removed

tools/twix/src/panels/behavior_tree/model.rs
pejotejo: done

tools/twix/src/panels/behavior_tree/control_flow_layout.rs
pejotejo: removed it

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn animate_layout(&mut self) -> bool {
+        let mut any_animating = false;
+
+        for node in &mut self.circle_nodes {
+            if node.is_dragging {
+                continue;
+            }
+
+            let delta = node.target_position - node.position;
+            if delta.inner.norm_squared() > LAYOUT_ANIMATION_EPSILON * LAYOUT_ANIMATION_EPSILON {
```
knoellle:
```suggestion
            if delta.norm_squared() > LAYOUT_ANIMATION_EPSILON * LAYOUT_ANIMATION_EPSILON {
```

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+            } else {
+                node.position = node.target_position;
+            }
+
+            node.opacity = (node.opacity + ENTER_FADE_STEP).min(1.0);
+        }
+
+        self.exiting_nodes.retain_mut(|node| {
+            let delta = node.target_position - node.position;
+            if delta.inner.norm_squared() > LAYOUT_ANIMATION_EPSILON * LAYOUT_ANIMATION_EPSILON {
```
knoellle:
```suggestion
            if delta.norm_squared() > LAYOUT_ANIMATION_EPSILON * LAYOUT_ANIMATION_EPSILON {
```

tools/twix/src/twix_painter.rs
```diff
@@ -424,14 +424,13 @@ impl<World> TwixPainter<World> {
     pub fn floating_text(
         &self,
         position: Point2<World>,
-        align: Align2,
-        text: impl ToString,
+        align: eframe::emath::Align2,
+        text: String,
```
knoellle, line 428: Why this change?
`self.painter.text` also takes an `impl ToString`, so we can just forward it and leave it to the user.

tools/twix/src/twix_painter.rs
```diff
@@ -424,14 +424,13 @@ impl<World> TwixPainter<World> {
     pub fn floating_text(
         &self,
         position: Point2<World>,
-        align: Align2,
-        text: impl ToString,
+        align: eframe::emath::Align2,
```
knoellle, line 427: Y u do dis?

tools/twix/src/panels/behavior_tree/mod.rs
```diff
@@ -183,27 +169,23 @@ impl BehaviorTreePanel {
...
-
-                node.opacity = (node.opacity - EXIT_FADE_STEP).max(0.0);
-                if node.opacity > 0.0 {
-                    remaining_exiting_nodes.push(node);
-                    any_animating = true;
-                }
+        self.exiting_nodes.retain_mut(|node| {
+            let delta = node.target_position - node.position;
+            if delta.inner.norm_squared() > LAYOUT_ANIMATION_EPSILON * LAYOUT_ANIMATION_EPSILON {
+                node.position += delta * LAYOUT_ANIMATION_FACTOR;
```
knoellle: This makes the animation speed depend on the user's framerate. Consider using egui's frametime (delta time) to fix this.

tools/twix/src/twix_painter.rs
pejotejo: i reverted the changes to how it was before, i did not change the painter compared to main

tools/twix/src/twix_painter.rs
pejotejo: same as above

knoellle (changes requested):
LGTM but is missing a twix version bump

knoellle (approved):
LGTM

PR #2442, opened 2026-04-19
Add Stop Button in Twix

pejotejo:
## Why? What?
Adds a stop button in twix which sets the primary state to safe until the stand or f1 button is pressed. With this we can then stop the robot remotely during emergencies or scary behavior. 

- one maybe could go into dampening instead, since this would not harm the robot as much when it falls
- 
## How to Test

1. put the robot into any primary state but safe
2. put the robot in a safe position 
3. press stop

### DIFF
--- a/crates/world_state/src/primary_state_filter.rs
+++ b/crates/world_state/src/primary_state_filter.rs
@@ -16,6 +16,7 @@ use types::{
 #[derive(Deserialize, Serialize)]
 pub struct PrimaryStateFilter {
     last_primary_state: PrimaryState,
+    last_safe_pose: u32,
 }
 
 #[context]
@@ -29,6 +30,7 @@ pub struct CycleContext {
     is_safe_pose: Input<bool, "is_safe_pose">,
 
     injected_primary_state: Parameter<Option<PrimaryState>, "injected_primary_state?">,
+    injected_safe_pose: Parameter<u32, "injected_safe_pose">,
     player_number: Parameter<PlayerNumber, "player_number">,
     recorded_primary_states: Parameter<HashSet<PrimaryState>, "recorded_primary_states">,
 
@@ -45,6 +47,7 @@ impl PrimaryStateFilter {
     pub fn new(_context: CreationContext) -> Result<Self> {
         Ok(Self {
             last_primary_state: PrimaryState::Safe,
+            last_safe_pose: 0,
         })
     }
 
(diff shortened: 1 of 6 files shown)


crates/world_state/src/robot_mode_handler.rs
```diff
@@ -52,6 +61,26 @@ impl BoosterModeHandler {
             });
         }
 
+        if self.button_pressed ^ *context.force_stop_robot {
```
alexschmander:
```suggestion
        if self.button_pressed != *context.force_stop_robot {
```

crates/world_state/src/robot_mode_handler.rs
```diff
@@ -52,6 +61,26 @@ impl BoosterModeHandler {
...
+            ) {
+                self.button_pressed = !self.button_pressed;
+            }
+
+            change_mode(&context, RobotMode::Damping);
+            return Ok(MainOutputs {
+                robot_mode: Some(RobotMode::Damping).into(),
+            });
+        }
+
```
alexschmander: As discussed:
```suggestion
        let is_local_stop_toggle_short_press =
            matches!(context.buttons.f1, Some(ButtonPressType::Short))
                || matches!(context.buttons.stand, Some(ButtonPressType::Short));

        if is_local_stop_toggle_short_press {
            self.local_stop_toggle = !self.local_stop_toggle;
        }

        let should_enter_damping_mode = self.local_stop_toggle != *context.remote_stop_toggle;
        if should_enter_damping_mode {
            change_mode(&context, RobotMode::Damping);
            return Ok(MainOutputs {
                robot_mode: Some(RobotMode::Damping).into(),
            });
        }
```

crates/world_state/src/primary_state_filter.rs
```diff
@@ -16,6 +16,7 @@ use types::{
 #[derive(Deserialize, Serialize)]
 pub struct PrimaryStateFilter {
     last_primary_state: PrimaryState,
+    last_safe_pose: u32,
```
alexschmander: This is unused

alexschmander (changes requested):

alexschmander (dismissed):
LGTM

alexschmander:
This branch now has conflicts...

alexschmander (changes requested):
Please also bump twix version

alexschmander (approved):
LGTM

PR #2280, opened 2026-03-09
Introduce 2 service approach

oleflb:
## Why? What?

The motivation is to be able to easily run commands inside the hulk container, which has special mounts, libraries and environment variables.
With the `launchHULK` script, this required some shenanigans in order to launch a shell.

This PR introduces a 2 service approach:
- One `hulk-runtime` service, which spawns the container and does nothing (sleeps)
- The `hulk` service, which has a dependency on `hulk-runtime` and execs into the container to run the `hulk` binary.

Therefore it is now very simple for users to run custom commands (or a shell) in the container by just running `sudo podman exec -it hulk bash`

In this PR I also fixed logging such that the journal can be used again.

Fixes #

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

*Describe how to test your changes. (For the reviewer)*

PR #2443, opened 2026-04-22
Create log directory before writing to it

knoellle:
## Why? What?

Fixes an issue introduced in #2280. Since that was merged, the log directory with the timestamp was created after the `tee`s tried to write to it, causing them to emit an error message but no log files being created.

- Fixes  #

## How to Test

- Gammaray a robot with the new launch-hulk script
- Restart hulk service and see the logs being created correctly


alexschmander (approved):
LGTM

PR #2445, opened 2026-04-22
Pepsi klotz zeig pfade

knoellle:
## Why? What?

- Based on #2443
- Fixes another issue caused by #2280

## How to Test

`pepsi klotz zeig` should work again.


alexschmander (approved):
LGTM

PR #2446, opened 2026-04-22
Query booster's user id instead of hardcoding it

knoellle:
## Why? What?

What it says.
The `booster` user has a different user ID on PeterFox than on the robots.

- Fixes #

## How to Test

gammaray then upload to peter fox.


alexschmander (approved):
LGTM

PR #2449, opened 2026-04-23
Bump K1 firmware to 1.5.3

alexschmander:
## Why? What?

Update hulk.toml os_version to 1.5.3 that matches the currently installed firmware on both robots.


knoellle (approved):
LGTM

PR #2451, opened 2026-04-23
Use Heatmap in Searcher Subtree

pejotejo:
depends on #2436 

## Why? What?
improves the searcher subtree: When we now loose a ball we first go look a the warmest position on the heatmap before doing leuchtturm. We also look a potential ball hypothesis when loosing the ball. 

- introduce different ballsearch for the other roles

## How to Test

In Playing let the robot see the ball, remove it. The robot should now walk to the former location of the ball and do the leuchtturm at some point. You can also look at the map panel in twix with the ball search heatmap overlay.


JackyBloxx (approved):
lgtm

PR #2452, opened 2026-04-23
Port validator to consistent naming and single model evaluation

alexschmander:
## Why? What?

Ports the current multi-task pipelines to a new consistent model naming scheme.

## ToDo

- [ ] Port export script


knoellle (approved):
LGTM

PR #2466, opened 2026-04-26
Update Konerl training

BenSampaolo:
## Why? What?

Updated the training to the current state.

Walking and kicking

## How to Test

Start a training using the train.sh script or run 
`uv run play Mjlab-Standing-K1 --agent zero` 
to see that the simulation is working. In case you have a training you can run 
`uv run play Mjlab-Standing-K1 --checkpoint-file your/path/to/the/checkpoint.pt`


alexschmander (approved):
Very nice work! LGTM 🚀

PR #2476, opened 2026-04-29
reintroduce remote control

pejotejo:
what the title says


crates/world_state/src/behavior/tree.rs
BenSampaolo, line 1: Why remove the injected motion command?

BenSampaolo (approved):
LGTM!

PR #2478, opened 2026-04-29
messages in new behavior

pejotejo:
## Why? What?
This pr moves player-messages from roleassignment to our new behavior. It also introduces a separate message receiver and removes roleassignment. Now we can start implementing behavior with multiple robots and also start using voronoi for positioning.

## How to Test

Take to robots and look at the playerstates output in twix.


crates/world_state/src/world_state_composer.rs
```diff
@@ -42,10 +41,10 @@ pub struct CycleContext {
     obstacles: Input<Vec<Obstacle>, "obstacles">,
     position_of_interest: Input<Point2<Ground>, "position_of_interest">,
     primary_state: Input<PrimaryState, "primary_state">,
-    role: Input<Role, "role">,
     rule_ball: Input<Option<BallState>, "rule_ball_state?">,
     rule_obstacles: Input<Vec<RuleObstacle>, "rule_obstacles">,
     suggested_search_position: Input<Option<Point2<Field>>, "suggested_search_position?">,
+    player_states: Input<Vec<PlayerState>, "player_states">,
```
alexschmander: Why didn't you use the `Players<PlayerState>` struct?

crates/types/src/world_state.rs
```diff
@@ -118,5 +120,12 @@ pub struct RobotState {
...
     pub player_number: PlayerNumber,
     pub primary_state: PrimaryState,
-    pub role: Role,
+}
+
+#[derive(
+    Clone, Debug, Default, Serialize, Deserialize, PathSerialize, PathDeserialize, PathIntrospect,
+)]
+pub struct PlayerState {
+    pub player_number: PlayerNumber,
```
alexschmander: When using the `Players` struct, this will not be needed anymore.

crates/types/src/world_state.rs
```diff
@@ -118,5 +120,12 @@ pub struct RobotState {
...
+    pub pose: Pose2<Field>,
```
alexschmander, line 137: Do you think adding the ball from the message to this state also makes sense? Then the team ball receiver, could be changed to a team ball filter, only choosing the best team ball.

crates/hsl_network_messages/src/lib.rs
```diff
@@ -18,16 +18,16 @@ pub use game_controller_state_message::{
...
 
-#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(
+    Clone, Copy, Debug, Deserialize, Serialize, PathDeserialize, PathIntrospect, PathSerialize,
+)]
 pub enum HulkMessage {
-    Striker(StrikerMessage),
-    Loser(LoserMessage),
-    VisualReferee(VisualRefereeMessage),
+    Base(BaseMessage),
```
alexschmander: I suggest changing the name something like `StateMessage`.

crates/world_state/src/behavior/node.rs
```diff
@@ -40,6 +43,8 @@ pub struct Behavior {
     pub tree: Node<Blackboard>,
     #[serde(skip, default = "create_static_layout_default")]
     pub static_layout: NodeTrace,
+    pub last_system_time_transmitted_game_controller_return_message: Option<SystemTime>,
+    pub last_transmitted_hsl_message: Option<SystemTime>,
```
alexschmander:
```suggestion
    pub last_sent_game_controller_return_message_time: Option<SystemTime>,
    pub last_sent_hsl_message_time: Option<SystemTime>,
```

crates/world_state/src/behavior/send_message.rs
```diff
@@ -0,0 +1,139 @@
...
+    ) -> bool {
+        is_cooldown_elapsed(
+            now,
+            self.last_transmitted_hsl_message,
+            hsl_network_parameters.hsl_base_message_send_interval,
+        )
+    }
+}
+
+fn ground_to_field_or_initial_pose(world_state: &WorldState) -> Isometry2<Ground, Field> {
```
alexschmander: I think this function is a bit redundant. Naming this expression is not that useful imo, and makes the other functions less readable.

crates/world_state/src/behavior/send_message.rs
```diff
@@ -0,0 +1,139 @@
...
+use coordinate_systems::{Field, Ground};
+use hardware::NetworkInterface;
+use hsl_network_messages::{BaseMessage, GameControllerReturnMessage, HulkMessage};
+use linear_algebra::Isometry2;
+use types::{messages::OutgoingMessage, parameters::HslNetworkParameters, world_state::WorldState};
+
+use crate::behavior::node::Behavior;
+
+impl Behavior {
+    pub fn try_sending_game_controller_return_message(
```
alexschmander: Since this is not an overloaded function, like it is the case for `from` and `try_from`, using `try_...` here is unnecessary imo. 
```suggestion
    pub fn send_game_controller_return_message(
```

crates/world_state/src/behavior/send_message.rs
```diff
@@ -0,0 +1,139 @@
...
+        hsl_network_parameters: &HslNetworkParameters,
+    ) -> bool {
+        is_cooldown_elapsed(
+            now,
+            self.last_system_time_transmitted_game_controller_return_message,
+            hsl_network_parameters.game_controller_return_message_interval,
+        )
+    }
+
+    pub fn try_sending_base_message(
```
alexschmander: See above
```suggestion
    pub fn send_base_message(
```

crates/world_state/src/behavior/send_message.rs
```diff
@@ -0,0 +1,139 @@
...
+        remaining_amount_of_messages: Option<&u16>,
+        last_sent_message: &mut AdditionalOutput<HulkMessage>,
+        hardware: &Arc<impl NetworkInterface>,
+    ) -> Result<()> {
+        let now = world_state.now;
+
+        if !self.is_base_message_cooldown_elapsed(now, hsl_network_parameters) {
+            return Ok(());
+        }
+        if remaining_amount_of_messages.is_some_and(|remaining_amount_of_messages| {
```
alexschmander: In the case that we have a None for `remaining_amount_of_messages`, this always returns false, and the early return is skipped, allowing message sending. Is this the intended behavior?

crates/hsl_network_messages/src/lib.rs
```diff
@@ -39,16 +39,21 @@ pub struct StrikerMessage {
...
+    Default,
+    Deserialize,
+    Serialize,
+    PathDeserialize,
+    PathIntrospect,
+    PathSerialize,
+)]
+pub struct BaseMessage {
     pub player_number: PlayerNumber,
     pub pose: Pose2<Field>,
```
alexschmander, line 55: If you add the ball_position to the player_state, you could also just directly send the player_state struct along with the player_number in the base/state message.

crates/world_state/src/player_states_receiver.rs
```diff
@@ -0,0 +1,88 @@
...
+use linear_algebra::Isometry2;
+use serde::{Deserialize, Serialize};
+use types::{
+    ball_position::BallPosition, cycle_time::CycleTime, messages::IncomingMessage,
+    parameters::HslNetworkParameters, world_state::PlayerState,
+};
+
+#[derive(Serialize, Deserialize)]
+pub struct PlayerStatesReceiver {
+    pub last_player_states: Vec<PlayerState>,
```
alexschmander: `Players` struct here as well.

crates/world_state/src/player_states_receiver.rs
```diff
@@ -0,0 +1,88 @@
...
+                        .iter()
+                        .position(|player| player.player_number == base_message.player_number)
+                    {
+                        player_states[index].pose = base_message.pose;
+                    } else {
+                        player_states.push(PlayerState {
+                            player_number: base_message.player_number,
+                            pose: base_message.pose,
+                        });
+                    }
```
alexschmander: This will be drastically simplified, if you are using the `Players` struct.

alexschmander (changes requested):
A couple of things :) Otherwise works!

crates/world_state/src/world_state_composer.rs
pejotejo: good question, now it is

crates/types/src/world_state.rs
pejotejo: is removed

crates/hsl_network_messages/src/lib.rs
pejotejo: done

crates/types/src/world_state.rs
pejotejo: sounds good, done

crates/world_state/src/player_states_receiver.rs
pejotejo: done

crates/world_state/src/player_states_receiver.rs
pejotejo: done

crates/hsl_network_messages/src/lib.rs
pejotejo: i cant do that if i turn the teamballreceiver into a filter, since this needs a different ballposition with last seen and this is best calculated when the message is received

crates/world_state/src/behavior/send_message.rs
pejotejo: your right, i'm supprised that never happened with the old behavior

crates/world_state/src/behavior/send_message.rs
pejotejo: done

crates/world_state/src/behavior/send_message.rs
pejotejo: done

crates/world_state/src/behavior/send_message.rs
pejotejo: this was a relic from when i removed the initial pose, it's now gone

crates/world_state/src/behavior/node.rs
pejotejo: done

crates/types/src/parameters.rs
```diff
@@ -259,7 +258,7 @@ pub struct HslNetworkParameters {
     pub remaining_amount_of_messages_to_stop_sending: u16,
     pub silence_interval_between_messages: Duration,
     pub hsl_striker_message_receive_timeout: Duration,
-    pub hsl_striker_message_send_interval: Duration,
+    pub hsl_base_message_send_interval: Duration,
```
alexschmander: There still are a couple of references to `base`

crates/world_state/src/behavior/send_message.rs
```diff
@@ -0,0 +1,134 @@
...
+        hsl_network_parameters: &HslNetworkParameters,
+    ) -> bool {
+        is_cooldown_elapsed(
+            now,
+            self.last_sent_game_controller_return_message_time,
+            hsl_network_parameters.game_controller_return_message_interval,
+        )
+    }
+
+    pub fn send_base_message(
```
alexschmander: Here as well: using the base message name

crates/hsl_network_messages/src/lib.rs
alexschmander: No i just mean, to use the PlayerState as the container, in which the pose and ball_position are packaged in the message. The type definitions of the pose and ball_positions, in this `StateMessage` are identical, to those in `PlayerState`.

alexschmander (changes requested):

crates/hsl_network_messages/src/lib.rs
pejotejo: The two `BallPosition` are different from another, on contains last_seen as a SystemTime and the age as a duration. We cant transmit the system time, but if we give the playerstate as output the age would not increase. So I can't use the Playerstate in the messages if it is the output.

crates/types/src/parameters.rs
pejotejo: renamed

crates/world_state/src/behavior/send_message.rs
pejotejo: renamed

crates/types/src/parameters.rs
alexschmander: No you did not...

crates/world_state/src/behavior/send_message.rs
alexschmander: And here as well

alexschmander:
Also, the branch has conflicts  again

alexschmander (changes requested):
Also one thing I noticed: The HSL Network cycler does not start cycling until the game controller is available in the network. This also means, that testing this feature requires the gamecontroller to be available during startup of the framework. 
Is this intended?

crates/types/src/parameters.rs
pejotejo: sorry, forgot to push the commit, now it should be

crates/world_state/src/behavior/send_message.rs
pejotejo: same here

alexschmander (approved):
LGTM

PR #2481, opened 2026-05-02
integrate ros z

schmidma:
Introduces the initial ros-z stack: Zenoh-native Rust middleware plus supporting crates for CDR serialization, protocol/key expressions, schema/type info, derives, stream helpers, and CLI inspection commands.

Suggested review path: start with `crates/ros-z/README.md` and `crates/ros-z/src/lib.rs`, then review by crate boundary. For behavior, focus on pub/sub, services, parameters, graph discovery, QoS/time, dynamic messages, and the tests next to each crate.


pejotejo (approved):
Image

PR #2484, opened 2026-05-05
Multi task yolo/fix training weight reuse

alexschmander:
## Why? What?

Fixed the loading and reuse of models generated by the `src/validator/validation.py`. Currently, the saving is done after the validation call. This changes the pytorch architecture for inference optimization, causing the saved model to be different architecturally, than the original model, resulting in the model weights not correctly being transferred during training. 
Now the model is saved before the validation call.


knoellle (approved):
LGTM

PR #2485, opened 2026-05-06
Add Walk to Kickoff Behavior

### LINKED ISSUE #2447: Einlaufbaum
Implementation of the Ready Subtree

paudar:
## Why? What?

The behaviour-tree is extended to handle the game-state ready. The players walk to their Kick-off poses, to be rule conform and have a better start of the game. Player 3 is placed depending on the kicking team, as he is sometimes allowed to be in the centre circle.

- Fixes #2447

- choose the pose dynamicly

## How to Test

Use a robot and a gamecontroller at a field.

### DIFF
--- a/crates/types/src/parameters.rs
+++ b/crates/types/src/parameters.rs
@@ -82,6 +82,7 @@ pub struct BehaviorParameters {
     pub path_planning: PathPlanningParameters,
     pub remote_control: RemoteControlParameters,
     pub role_positions: RolePositionsParameters,
+    pub kickoff_positions: KickOffPositionsParameters,
     pub search: SearchParameters,
     pub look_action: LookActionParameters,
     pub walk_and_stand: WalkAndStandParameters,
@@ -157,6 +158,25 @@ pub struct RolePositionsParameters {
     PathDeserialize,
     PathIntrospect,
     ros_z::Message,
+)]
+pub struct KickOffPositionsParameters {
+    pub one: Point2<Field>,
+    pub two: Point2<Field>,
+    pub three: Point2<Field>,
+    pub four: Point2<Field>,
+    pub five: Point2<Field>,
+}
+
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    Default,
+    Deserialize,
+    Serialize,
+    PathSerialize,
+    PathDeserialize,
+    PathIntrospect,
 )]
 pub struct SearchParameters {
     pub position_reached_distance: f32,
--- a/crates/types/src/parameters.rs
+++ b/crates/types/src/parameters.rs
@@ -177,6 +177,7 @@ pub struct KickOffPositionsParameters {
     PathSerialize,
     PathDeserialize,
     PathIntrospect,
+    ros_z::Message,
 )]
 pub struct SearchParameters {
     pub position_reached_distance: f32,
--- a/crates/types/src/parameters.rs
+++ b/crates/types/src/parameters.rs
@@ -9,7 +9,7 @@ use coordinate_systems::{Camera, Field, Ground, NormalizedPixel, Pixel, Robot};
 use linear_algebra::{Point2, Vector2, Vector3};
 use path_serde::{PathDeserialize, PathIntrospect, PathSerialize};
 
-use crate::{motion_command::MotionCommand, step::Step};
+use crate::{motion_command::MotionCommand, players::Players, step::Step};
 
 #[derive(
     Clone,
@@ -82,7 +82,7 @@ pub struct BehaviorParameters {
     pub path_planning: PathPlanningParameters,
     pub remote_control: RemoteControlParameters,
     pub role_positions: RolePositionsParameters,
-    pub kickoff_positions: KickOffPositionsParameters,
+    pub kickoff_positions: Players<KickOffPose>,
     pub search: SearchParameters,
     pub look_action: LookActionParameters,
     pub walk_and_stand: WalkAndStandParameters,
@@ -159,12 +159,9 @@ pub struct RolePositionsParameters {
     PathIntrospect,
     ros_z::Message,
 )]
-pub struct KickOffPositionsParameters {
-    pub one: Point2<Field>,
-    pub two: Point2<Field>,
-    pub three: Point2<Field>,
-    pub four: Point2<Field>,
-    pub five: Point2<Field>,
+pub struct KickOffPose {
+    pub position: Point2<Field>,
+    pub rotation: f32,
 }
 
 #[derive(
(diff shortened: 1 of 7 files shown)


crates/world_state/src/behavior/condition.rs
```diff
@@ -89,3 +92,13 @@ pub fn has_hypothetical_ball_position(blackboard: &mut Blackboard) -> bool {
...
+
+pub fn hulks_is_kicking_team(blackboard: &mut Blackboard) -> bool {
+    matches!(
+        blackboard.world_state.filtered_game_controller_state,
+        Some(FilteredGameControllerState {
+            kicking_team: Some(Team::Hulks),
+            ..
+        })
+    )
+}
```
pejotejo, line 141: this function now exists twice, you probably want to remove the one in substates, sice it is now not only related to substates anymore

crates/world_state/src/behavior/walk.rs
```diff
@@ -198,3 +200,38 @@ pub fn walk_instead_of_kicking(blackboard: &mut Blackboard) -> Status {
         Status::Failure
     }
 }
+
+pub fn walk_to_kickoff_pose(blackboard: &mut Blackboard) -> Status {
+    let ground_to_field = match blackboard.world_state.robot.ground_to_field {
+        Some(transform) => transform,
+        None => return Status::Failure,
+    };
```
pejotejo: in all other nodes we used if let, for consistency it would be nice if you could also do that

crates/types/src/parameters.rs
```diff
@@ -127,6 +128,26 @@ pub struct RolePositionsParameters {
...
+    PathIntrospect,
+    ros_z::Message,
+)]
+pub struct KickOffPositionsParameters {
+    pub one: Point2<Field>,
+    pub two: Point2<Field>,
+    pub three: Point2<Field>,
+    pub four: Point2<Field>,
+    pub five: Point2<Field>,
+}
```
pejotejo: You can use the Players struct instead. An example for that you can see with the initial_poses in the `localisation.rs`. Then we probably don't need a new Parameter Type.

crates/types/src/parameters.rs
```diff
@@ -127,6 +128,26 @@ pub struct RolePositionsParameters {
...
+    Default,
+    Deserialize,
+    Serialize,
+    PathSerialize,
+    PathDeserialize,
+    PathIntrospect,
+    ros_z::Message,
+)]
+pub struct KickOffPositionsParameters {
+    pub one: Point2<Field>,
```
pejotejo: You should use a Pose2 instead, such that the robots look in the right direction and not just where they looked at last.

crates/world_state/src/behavior/walk.rs
```diff
@@ -198,3 +200,38 @@ pub fn walk_instead_of_kicking(blackboard: &mut Blackboard) -> Status {
...
+        blackboard.world_state.robot.player_number,
+        hulks_is_kicking_team(blackboard),
+    ) {
+        (PlayerNumber::One, _) => blackboard.parameters.kickoff_positions.one,
+        (PlayerNumber::Two, _) => blackboard.parameters.kickoff_positions.two,
+        (PlayerNumber::Three, false) => blackboard.parameters.kickoff_positions.three,
+        (PlayerNumber::Three, true) => point!(0.0, -0.5),
+        (PlayerNumber::Four, _) => blackboard.parameters.kickoff_positions.four,
+        (PlayerNumber::Five, _) => blackboard.parameters.kickoff_positions.five,
+    };
```
pejotejo: with a Players struct one can use the robots player number its index to get the right case

crates/world_state/src/behavior/condition.rs
paudar: Done

crates/types/src/parameters.rs
paudar: Yes, that looks much cleaner. I still want a new parameter (see next comment).

crates/types/src/parameters.rs
paudar: I tried to use a `Pose2` but the deserialization is kind of painful, therefore i still had a new parameter with the fields `position` and `rotation`, also to make the default.json more self explaining.

crates/world_state/src/behavior/walk.rs
```diff
@@ -203,6 +205,41 @@ pub fn walk_to_block_position(blackboard: &mut Blackboard) -> Status {
...
+            target_position = point!(0.0, -0.5);
+        }
+
+        let tragtet_in_ground = field_to_ground * target_position;
+
+        walk_to(
+            blackboard,
+            Pose2::from_parts(
+                tragtet_in_ground,
+                Orientation2::new(blackboard.parameters.kickoff_positions[player_number].rotation),
```
pejotejo: die orientation muss auch in ground umgerechnet werden, dann kannst du erst die pose berechnen und dann transformieren

etc/parameters/default.json
```diff
@@ -465,6 +465,13 @@
       "striker_distance_to_non_free_center_circle": 0.4,
       "striker_kickoff_position": [-0.21, 0.03]
     },
+    "kickoff_positions": {
+      "one": { "position": [-3.5, 1], "rotation": 0 },
```
pejotejo: why is player one further away from the y axis, than player two?

crates/world_state/src/behavior/walk.rs
```diff
@@ -203,6 +205,43 @@ pub fn walk_to_block_position(blackboard: &mut Blackboard) -> Status {
...
+    if let (Some(ground_to_field), player_number) = (
+        blackboard.world_state.robot.ground_to_field,
+        blackboard.world_state.robot.player_number,
+    ) {
+        let field_to_ground = ground_to_field.inverse();
+
+        let mut target_position = blackboard.parameters.kickoff_positions[player_number].position;
+
+        if hulks_is_kicking_team(blackboard) && player_number == PlayerNumber::Three {
+            target_position = point!(-0.5, 0.0);
```
pejotejo: is it possble to put this also somewhere a parameter?

etc/parameters/default.json
```diff
@@ -465,6 +465,13 @@
...
+      "two": { "position": [-3, 0.5], "rotation": 0 },
+      "three": { "position": [-2, 0], "rotation": 0 },
+      "four": { "position": [-2.5, -1.5], "rotation": 0 },
+      "five": { "position": [-2.5, 1.5], "rotation": 0 }
```
pejotejo: it would be nice if those two would be angled at the ball

etc/parameters/default.json
paudar: No reason. Is updated to be 0.5 for both

crates/world_state/src/behavior/walk.rs
paudar: Done

etc/parameters/default.json
paudar: Done

pejotejo (approved):
lgtm

PR #2488, opened 2026-05-07
Change PlayerState in player_state_reciver into option

JackyBloxx:
## Why? What?

What the title says.
Changed so if no message is recived it is none and not default.

- Fixes #


pejotejo (approved):
lgtm

PR #2489, opened 2026-05-08
reset player_state when penalyzed

JackyBloxx:
## Why? What?

resets the playerstate for a robot when penalized to None
- Fixes #

## How to Test

Start two Robots that sends mesages then look into playerstates. When a robot gets penalized the other robot should reset that robots playerstates to none.


pejotejo (approved):
lgtm

PR #2486, opened 2026-05-07
Pose detection image overlay

alexschmander:
## Why? What?

Adds a pose detection image overlay in twix visualizing the detected poses output by `Hydra`.

## How to Test

- Upload to a robot
- `./twix`
- Open image panel and enable image overlay `Pose Detection`

Issue #2414, opened 2026-03-29
Root: Multi-task feature detection


alexschmander:
# Thesis plan 

## Levels of polish

1. Bullet points: 
    - [x] **DEADLINE 2025-05-27**
2. Basic red thread:  
    - [x] Depends on #2540 2026-05-25 @alexschmander 
    - [x] **DEADLINE 2025-05-31**   
3. Informationally complete: 
    - [x] Depends on #2498 ... @knoellle 
    - [x] **DEADLINE 2025-06-07** 
4. Polished phrasing: 
    - [x] **DEADLINE 2025-06-10**

PR #2490, opened 2026-05-08
Update hydra model for better goal post detection

alexschmander:
## Why? What?

Updates the yolo model doing object and pose detection. The changes are mainly in now using a segmentation backbone instead of the object detection backbone. Additionally, the finetuning training of both heads reused the weights from the base model heads instead of training from randomly initialize weights. Lastly, the `Person` label and associated data from removed from the object dataset in order to improve the class balance and we still have the `Person` detection via the pose detection head with both bounding boxes and keypoints.   

Uses results from: 
- #2414 
- #2484 

## How to Test

- Upload to robot. 
- Open twix and the image, enable the overlays.
- #2486 adds a pose detection image overlay to actually visualize the poses.


alexschmander:
Was merged accidentally in #2499

PR #2491, opened 2026-05-08
Pose detections in obstacle filter

alexschmander:
## Why? What?

To improve the goalpost accuracy #2490 removed the redundant Person label from the object detection head of hydra. Since we currently rely on the detected person objects to spawn person obstacles, this causes the obstacle filter to simply not spawn any person obstacles anymore. Luckily, the pose detection head already gives us person detection, even ones with keypoints. 
This PR adds these detected poses to the obstacle filter. If the feet keypoint are above a confidence threshold, they are used to determine the person obstacles position. If they are below, the bounding box` confidence is checked and if that is high enough, the position of the obstacle is determined from that, just like all other object detected obstacle. 
This PR also tunes the person obstacle measurement noise way down, since the detections are really consistent and have near zero false positive rates. This massively improves the responsiveness of person obstacles.

- Depends on #2490 

## How to Test

- Upload to the robot.
- Open twix
- Look at the obstacles in the map obstacles overlay


pejotejo:
one still needs to tune the parameters, else it looks good

pejotejo (approved):

PR #2495, opened 2026-05-10
Handle Substates

### LINKED ISSUE #2471: Add Substate handling

pejotejo:
## Why? What?

Adds substate handling to out behavior tree. While testing the robot always got stuck in spawned obstacles, therefore the robot now leaves obstacles in a spline to the first segment, before continuing its path. In addition I also introduced the fov decay on the Heatmap we developed in #1977 and also modified the way ruleballs are spawned on the heatmap. 
 
- closes #1977 
- closes #2471 

- better kick targets, especially in corner kicks
- behavior for other "roles"

## How to Test
 Take the robot and the gamecontroller and test the different substates. Too deal with the bad loca you can look at the map panel in twix to see where the robot thinks it is.


crates/world_state/src/behavior/condition.rs
```diff
@@ -51,6 +52,42 @@ pub fn is_close_to_ball(blackboard: &mut Blackboard) -> bool {
...
+        );
+
+        let direction = (goal_position - target_kick_position).normalize();
+        let robot_facing_direction = vector!(1.0, 0.0);
+        let is_aligned =
+            direction.angle(&robot_facing_direction) < parameters.alignment_angle_threshold;
+
+        blackboard.direction_difference = direction.angle(&robot_facing_direction);
+
+        is_close_and_aligned = is_close && (is_aligned || blackboard.last_close_enough_to_kick);
```
paudar, line 84: Why should `is_close_and_aligned` be true, if the robot is no longer aligned (but if it was aligned before)?

pejotejo: this condition is used for switching from walking to kicking in the substates. Once we were alligned we switch to the visual kick policy, which does something that would instantly set the allignment to false. To stop it from switching back we ignore the alignment from that point on to the point where we are not close to the ball anymore. In my opinion this is ok, since this whole thing should just position us more precise and controlled for kicking and when we then start kicking we should just behave like the normal kicking again and go on with it .

paudar (approved):
Lgtm

PR #2496, opened 2026-05-11
Require ros-z attachment metadata

schmidma:
## Motivation
ros-z messages need source timestamps for stream alignment, replay ordering, and latency handling. Letting publishers disable attachments creates a protocol state where typed ros-z consumers have to guess or degrade when metadata is missing.

## Changes
This removes the publisher attachment opt-out and makes typed pub/sub and service paths require valid attachment metadata. Typed receivers now fail with clear errors when raw Zenoh traffic reaches ros-z without the expected metadata.

## Review
Focus on API semantics, raw Zenoh boundaries, service callbacks, and transient-local replay.
- Does it make sense that typed ros-z rejects attachmentless samples, requests, and responses?

### DIFF
--- a/crates/ros-z-streams/src/announce.rs
+++ b/crates/ros-z-streams/src/announce.rs
@@ -229,6 +229,6 @@ mod tests {
                 .await
                 .expect("timeout waiting for payload")
                 .expect("receive payload");
-        assert_eq!(received_payload.publication_id(), Some(publication_id));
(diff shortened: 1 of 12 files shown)


alexschmander:
The tests on this PR are not green @schmidma

schmidma:
All checks successful now :)

alexschmander (approved):
LGTM

PR #2499, opened 2026-05-12
Tune obstacle filter

alexschmander:
## Why? What?

Tunes the obstacle filter parameters by reducing the measurement noise of persons, robots, and goal posts, since these are detected quite reliably now. 
This improves the responsiveness of the filter by a lot.
Also changed the `obstacle_merge_distance` to `0`, since we ran into an issue, where the smaller goal post hypothesis are merged with the larger person hypothesis, but retaining the smaller goal post radius. This change effectively disables the removing of hypothesis by merging.   

## How to Test

- Setup a robot, upload
- In twix, in the map panel look at the obstacles.


pejotejo (approved):
lgtm

PR #2500, opened 2026-05-13
refactor: remove enclave from ros-z

schmidma:
## Summary

Remove unused enclave state from ros-z protocol entities, context/node builders, and graph APIs.

## Motivation

ros-z does not serialize or use enclave in its native liveliness format, and enclave is not a requirement. Keeping it in public APIs suggests support for a concept the runtime does not implement.

## Reviewer Notes

This is a breaking API cleanup by design. The native `@ros_z/<zid>/<nid>/<eid>/<kind>/<ns>/<name>` liveliness format stays unchanged, covered by a new protocol test.


knoellle (approved):
LGTM

PR #2501, opened 2026-05-13
Drop `no_std` mode from `ros-z-protocol`

schmidma:
# Description

`ros-z-protocol` now requires `std`, matching its actual dependency requirements and removing an unsupported `no_std` configuration path. This simplifies the crate feature setup and lets workspace consumers depend on `ros-z-protocol` without explicitly enabling `std`.

Treat this as an intentional feature/API cleanup.


knoellle (approved):
LGTM

PR #2502, opened 2026-05-13
Require strict ros-z endpoint metadata

schmidma:
## Summary

- Stack on #2509 
- Require ros-z endpoint liveliness keys and runtime endpoint records to carry explicit schema hashes.
- Update graph, discovery, CLI, typed builders, dynamic endpoints, and hulk_ros_z callers to use strict type metadata.
- Remove placeholder and missing-hash handling from endpoint parsing and display paths.

## Reviewer Notes

- This is an intentional internal API break: `TypeInfo::new(name, hash)` is now the constructor.
- Dynamic publishers no longer manufacture missing hashes; callers must supply valid schema metadata.
- There is and was no path where this could be missing, this was a compatability left-over of upstream ros-z, as for example ROS2 Humble does not expose hashes on the liveliness wire.
- Please bear with the current error-handling noise in this PR; #2503 follows up to clean that up.

PR #2503, opened 2026-05-13
Improve ros-z error UX and preserve error sources

schmidma:
# Description

This PR reworks ros-z error handling so failures are easier to trace and compatible with downstream `eyre`/`color-eyre` reporting. The main motivation is to stop collapsing important runtime failures into strings, while treating invalid static type metadata as programmer errors, not runtime errors.

## Overview:

- Based on #2509 
- Based on #2502 
- Based on #2496 
- Based on #2501 : Drops `no_std` support for `ros-z-protocol` as a prerequisite
- Moves protocol errors to typed `std::error::Error` values.
- Adds a top-level `ros_z::Error` boundary with source-preserving variants.
- Makes wire encoding fallible and routes pub/sub/service runtime failures through typed errors.
- Preserves source chains for dynamic schema discovery, parameter operations, CDR decode errors, SHM, config, and downstream binary errors.
- Makes static typed metadata APIs infallible: `Message::schema()`, `Message::schema_hash()`, `Message::type_info()`, and `ServiceTypeInfo::service_type_info()`.
- Updates downstream crates and tests to avoid source-erasing string conversions.


crates/ros-z-protocol/src/format.rs
```diff
@@ -19,6 +20,13 @@ pub const EMPTY_PLACEHOLDER: &str = "%";
 
 const ESCAPE_CHAR: char = '%';
 
+fn key_expr(expression: String) -> Result<zenoh::key_expr::KeyExpr<'static>> {
+    expression
+        .clone()
+        .try_into()
+        .map_err(|source| ProtocolError::InvalidKeyExpression { expression, source })
+}
```
schmidma, line 28: If someone has a nice suggestion to avoid this `clone()`... Please let me know

crates/ros-z/src/dynamic/error.rs
alexschmander, line 1: Don't know where else to put this...  
As discussed, the issue that the execution stalls/freezes when a parameter cannot be properly deserialized remains.

crates/ros-z/README.md
```diff
@@ -26,7 +26,7 @@ C/C++ runtime dependencies.
 ```rust
 use ros_z::prelude::*;
 
-# async fn demo() -> zenoh::Result<()> {
+# async fn demo() -> ros_z::Result<()> {
 let context = ContextBuilder::default().build().await?;
 let node = context.create_node("talker").build().await?;
 let publisher = node.publisher::<String>("/chatter")?.build().await?;
```
alexschmander, line 32: Not really related to the changes by this PR, but why are there #s here in the code block?

crates/ros-z/src/parameter/remote/services.rs
```diff
@@ -421,15 +416,21 @@ where
...
 fn reply<T>(query: &Query, response: &T)
 where
     T: Message,
     for<'a> <T as Message>::Codec: WireEncoder<Input<'a> = &'a T>,
 {
-    let bytes = <<T as Message>::Codec as WireEncoder>::serialize(response);
+    let bytes = match <<T as Message>::Codec as WireEncoder>::serialize(response) {
+        Ok(bytes) => bytes,
+        Err(error) => {
+            tracing::warn!(error = ?error, "[PARAM] Failed to serialize parameter reply");
```
alexschmander, line 430: Why is this function infallible, if this is a failure case. Why do you just log it, instead of making the function fallible?

crates/ros-z/src/service.rs
```diff
@@ -298,9 +387,34 @@ where
...
+                } else {
+                    if let Some(remaining) = timeout.checked_sub(started_at.elapsed()) {
+                        tokio::time::sleep(remaining).await;
+                    }
+                    return Err(self.timeout_error(timeout));
+                }
+            }
+            Ok(Err(error)) => return Err(error),
+            Err(_) => return Err(self.timeout_error(timeout)),
+        };
```
alexschmander, line 334: I don't understand this section for handling the timeout. Why is there a sleep in the when handling the error, when afterwards it just emits an error anyway.

crates/ros-z/src/service.rs
```diff
@@ -23,6 +23,12 @@ use crate::{
     time::Clock,
 };
 
+fn zenoh_query_timeout_after(user_timeout: Duration) -> Duration {
+    user_timeout
+        .checked_add(Duration::from_secs(1))
+        .unwrap_or(user_timeout)
+}
```
alexschmander: Why is there always a second added to the user timeout?

alexschmander (changes requested):
A couple of things, questions and that parameter deserialize freeze. Otherwise the code looks fine and the errors look really good.

crates/ros-z/src/dynamic/error.rs
schmidma: yes, I'm on it. Not this PR

crates/ros-z/README.md
schmidma: The idea was to copy the rustdoc style of omitting lines that are not important to the actual demonstration. But I see your point that this might be more confusing than just keeping them. Future issue/cleanup

crates/ros-z/src/parameter/remote/services.rs
schmidma: Because this is the function that should reply to the service query... and if serializing this exact reply fails, what can we do?!

crates/ros-z/src/service.rs
schmidma: I know, weird complex behavior, but this is making sure, that if zenoh exits early, we do not return until the timeout actually happened. This is not needed complexity. I remove that.

crates/ros-z/src/service.rs
schmidma: also removed

crates/ros-z/README.md
alexschmander: Please open a corresponding issue

crates/ros-z/src/dynamic/error.rs
alexschmander: Is this already tracked somewhere?

alexschmander (approved):
LGTM

PR #2506, opened 2026-05-13
feat(ros-z-debug): add debug adapter crate

schmidma:
## Summary
- Add `ros-z-debug`, a read-only subscription manager crate for typed and dynamic debug subscriptions.
- Retain latest samples and optional source-time windows with status snapshots, events, and topic/type metadata.
- Harden the debug API by validating target namespaces through existing `ros-z` topic qualification, preserving topic validation sources, rejecting private selectors without node context, pruning dead registry refs, returning empty inverted history windows, and reporting receive error source chains with topic/type context.
- Make dynamic schema discovery timeout configurable through `ManagerOptions`.
- Clarify that dynamic debug subscriptions use currently visible publishers; the discovery timeout applies to schema service queries, not waiting for future publishers.

## Scope
- Introduces the first usable debug subscription layer for Twix-facing work.
- Does not replace direct `ros-z` subscribers for normal application receive loops.
- Keeps reconnect orchestration, runtime namespace retargeting, graph-driven topic discovery, pending dynamic subscriptions, and stricter global topic-name validation out of the initial API contract.
- Keeps the timestamp index private to `ros-z-debug`; it does not promote a new `ros_z::time_index` public API.

## Future Work
- Pending dynamic subscriptions that wait for publisher/schema availability instead of failing during build.
- Runtime target-namespace retargeting for switching inspected robots.
- Reconnect behavior that rebuilds managed subscriptions after connection changes.
- Graph-driven topic discovery and projection for Twix topic lists.
- A Twix backend/facade layer that uses `ros-z-debug` for read-only robot data.

## Test Plan
- [x] `cargo fmt --check`
- [x] `cargo nextest run -p ros-z-debug`
- [x] `cargo clippy -p ros-z-debug --all-targets --locked -- -D warnings`
- [x] `cargo test --doc -p ros-z-debug`

### DIFF
--- a/crates/ros-z-debug/Cargo.toml
+++ b/crates/ros-z-debug/Cargo.toml
@@ -11,8 +11,9 @@ parking_lot = { workspace = true }
 ros-z = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt", "sync"] }
+tokio = { workspace = true, features = ["macros", "rt"] }
 tokio-util = { workspace = true }
 
 [dev-dependencies]
 ros-z-schema = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "time"] }
--- a/crates/ros-z-debug/Cargo.toml
+++ b/crates/ros-z-debug/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "ros-z-debug"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+
+[dependencies]
+arc-swap = { workspace = true }
+parking_lot = { workspace = true }
+ros-z = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt", "sync"] }
+tokio-util = { workspace = true }
+
+[dev-dependencies]
+ros-z-schema = { workspace = true }
--- a/crates/ros-z-debug/src/manager.rs
+++ b/crates/ros-z-debug/src/manager.rs
@@ -1,0 +1,536 @@
+use std::{error::Error as _, fmt::Write as _, marker::PhantomData, sync::Arc, time::Duration};
+
+use parking_lot::Mutex;
+use ros_z::{Message, dynamic::DynamicPayload, node::Node};
+use tokio_util::sync::CancellationToken;
+
+use crate::{
+    JsonRenderPolicy, JsonSubscriptionHandle, Result, RetentionPolicy, SampleMetadata,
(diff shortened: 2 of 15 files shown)


crates/ros-z-debug/src/manager.rs
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    async fn build_dynamic_payload(self) -> Result<SubscriptionHandle<DynamicPayload>> {
+        let retention = self.retention;
+        let requested_topic = TopicSelector::new(self.topic)?;
+        let resolved_topic = requested_topic.resolve(self.manager.target_namespace())?;
+        let subscriber = self
+            .manager
+            .node()
+            .dynamic_subscriber_auto(&resolved_topic, Duration::from_secs(5))
```
paudar: What do you think about defining the timeout as a parameter in the `ManagerOptions` struct?

crates/ros-z-debug/tests/integration.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    publisher
+        .publish(&"hello".to_string())
+        .await
+        .expect("publish should work");
+
+    let deadline = tokio::time::Instant::now() + Duration::from_secs(2);
+    loop {
+        if let Some(record) = handle.latest() {
+            assert_eq!(record.value, "hello");
```
paudar, line 79: Maybe make the test more powerful by sending more than one sample, so only it shows that actually the last sample is stored. (Or is that to much?)

crates/ros-z-debug/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+arc-swap = { workspace = true }
+parking_lot = { workspace = true }
+ros-z = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt", "sync"] }
+tokio-util = { workspace = true }
+
+[dev-dependencies]
+ros-z-schema = { workspace = true }
```
paudar: I could not find a reason why to include the `"sync"` feature, but missing tokio with the features `["rt-multi-thread", "time"]` in the dev dependencies for the tests.

crates/ros-z-debug/src/manager.rs
schmidma: Added ManagerOptions::schema_discovery_timeout and use it for dynamic schema discovery.

crates/ros-z-debug/Cargo.toml
schmidma: Removed the unused sync feature from the normal dependency and moved rt-multi-thread/time to dev-dependencies for tests.

crates/ros-z-debug/tests/integration.rs
schmidma: Updated the integration test to publish a second sample and assert latest() advances to it.

paudar (approved):
LGTM

PR #2509, opened 2026-05-13
refactor: make ros-z endpoint factories fallible

schmidma:
## Summary

- Currently, builders have a weird intermediate failed state, storing an error string.

This PR changes that:

- Make ros-z publisher, subscriber, cache, service, and dynamic endpoint factories return `Result<Builder>`.
- Remove deferred schema-error state from endpoint builders so builders only represent valid endpoint metadata.
- Migrate workspace call sites, examples, docs, and tests to handle factory-time schema/type-info errors.

## Reviewer Notes

- This PR leaves some error-handling noise in place. #2503 follows up to clean that up, so please focus this review on builder invariants and call-site handling.

## How to test (ci does that as well)

- `cargo fmt --all`
- `cargo check --workspace --all-targets`
- `cargo nextest run -p ros-z -p ros-z-cli -p ros-z-streams`

### DIFF
--- a/crates/ros-z/src/node.rs
+++ b/crates/ros-z/src/node.rs
@@ -214,6 +214,26 @@ fn typed_message_type_info<T: Message>(schema: &Schema) -> TypeInfo {
     TypeInfo::with_hash(&T::type_name(), ros_z_schema::compute_hash(schema.as_ref()))
 }
 
+fn message_schema_build_error(
+    endpoint_kind: &str,
+    name: &str,
+    error: impl std::fmt::Display,
+) -> zenoh::Error {
+    zenoh::Error::from(format!(
+        "failed to build message schema for {endpoint_kind} '{name}': {error}"
+    ))
+}
+
+fn service_type_info_build_error(
+    endpoint_kind: &str,
+    name: &str,
+    error: impl std::fmt::Display,
+) -> zenoh::Error {
+    zenoh::Error::from(format!(
+        "failed to build service type info for {endpoint_kind} '{name}': {error}"
+    ))
+}
+
 fn registerable_schema_root(
     root_name: &str,
     schema: &ros_z_schema::SchemaBundle,
@@ -252,33 +272,25 @@ impl Node {
     /// - Absolute topics (starting with '/') are used as-is
     /// - Private topics (starting with '~') are expanded to `/<namespace>/<node_name>/<topic>`
     /// - Relative topics are expanded to `/<namespace>/<topic>`
-    pub fn publisher<T>(&self, topic: &str) -> PublisherBuilder<T>
+    pub fn publisher<T>(&self, topic: &str) -> Result<PublisherBuilder<T>>
     where
         T: Message,
         T::Codec: WireEncoder,
     {
         debug!("[NOD] Creating publisher: topic={}", topic);
-        match T::schema().map(Arc::new) {
-            Ok(schema) => {
-                let type_name = T::type_name();
-                let type_info = typed_message_type_info::<T>(&schema);
-                let builder = self.publisher_impl::<T, T::Codec>(topic, Some(type_info));
-                match registerable_schema_root(&type_name, schema.as_ref()).and_then(
-                    |should_register| {
-                        if should_register {
-                            self.register_schema_with_service(&type_name, Arc::clone(&schema))?;
-                        }
-                        Ok(())
-                    },
-                ) {
-                    Ok(()) => builder,
-                    Err(error) => builder.schema_error(error),
-                }
-            }
-            Err(error) => self
-                .publisher_impl::<T, T::Codec>(topic, None)
-                .schema_error(error),
+
+        let schema = Arc::new(
+            T::schema().map_err(|error| message_schema_build_error("publisher", topic, error))?,
+        );
+        let type_name = T::type_name();
+        let should_register = registerable_schema_root(&type_name, schema.as_ref())
+            .map_err(|error| message_schema_build_error("publisher", topic, error))?;
+        if should_register {
+            self.register_schema_with_service(&type_name, Arc::clone(&schema))
+                .map_err(|error| message_schema_build_error("publisher", topic, error))?;
         }
+
+        Ok(self.publisher_impl::<T, T::Codec>(topic, Some(typed_message_type_info::<T>(&schema))))
     }
 
     fn publisher_impl<T, S>(
@@ -307,7 +319,6 @@ impl Node {
             attachment: true,
             shm_config: self.shm_config.clone(),
             dyn_schema: None,
-            schema_error: None,
             _phantom_data: Default::default(),
         }
     }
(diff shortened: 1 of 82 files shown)


crates/ros-z/src/node.rs
```diff
@@ -485,18 +491,18 @@ impl Node {
     /// - Absolute service names (starting with '/') are used as-is
     /// - Private service names (starting with '~') are expanded to `/<namespace>/<node_name>/<service>`
     /// - Relative service names are expanded to `/<namespace>/<service>`
-    pub fn create_service_client<T>(&self, name: &str) -> ServiceClientBuilder<T>
+    pub fn create_service_client<T>(&self, name: &str) -> Result<ServiceClientBuilder<T>>
```
knoellle, line 494: The pattern of `Fn() -> Result<Builder>` seems weird to me.
Why is there validation happening before the `build` call?

crates/ros-z/src/node.rs
```diff
@@ -485,18 +491,18 @@ impl Node {
...
     {
         debug!("[NOD] Creating service client: name={}", name);
-        match T::service_type_info().and_then(|type_info| {
-            Self::validate_service_message_schemas::<T>()?;
-            Ok(type_info)
-        }) {
-            Ok(type_info) => self.create_client_impl(name, Some(type_info)),
-            Err(error) => self.create_client_impl(name, None).schema_error(error),
-        }
+        let type_info = T::service_type_info()
```
knoellle, line 499: Why does this function return a result? As far as I can tell, all implementations of this trait are either not fallible or their only error path consists of checking if a static string is non-empty.

crates/ros-z/src/node.rs
```diff
@@ -409,15 +418,13 @@ impl Node {
...
-        let sub_builder = match T::schema().and_then(|_| T::type_info()) {
-            Ok(type_info) => {
-                self.subscriber_impl::<T, <T as Message>::Codec>(topic, Some(type_info))
-            }
-            Err(error) => self
-                .subscriber_impl::<T, <T as Message>::Codec>(topic, None)
-                .schema_error(error),
-        };
-        CacheBuilder::new(sub_builder, capacity)
+        T::schema().map_err(|error| message_schema_build_error("cache", topic, error))?;
```
knoellle, line 421: Why is `T::schema` fallible?
Since `T` implements message, it should always be possible to create a schema for it, no?

crates/ros-z/src/node.rs
```diff
@@ -631,37 +634,27 @@ impl Node {
...
-        match build_error {
-            Some(error) => builder.schema_error(error),
-            None => builder,
+        if type_info.hash.is_none() {
+            type_info.hash = Some(ros_z_schema::compute_hash(schema.as_ref()));
         }
+
+        let should_register = registerable_schema_root(&type_info.name, schema.as_ref())
+            .map_err(|error| message_schema_build_error("publisher", topic, error))?;
+        if should_register {
```
knoellle, line 652: If I see this correctly, `registerable_schema_root` is incapable of returning `Ok(false)`, only `Ok(true)` or `Err(..)`, so this check is redundant. Maybe change the return type?

knoellle (changes requested):
I don't really understand the point of this PR.
Why not just store the raw parameters in the builder and do all validation in the build function?

schmidma:
Thank you for the ideas and review. I agree to all of your points. But I see all of them as future work. There is a lot to clean up and reduce layers of abstractions and unnecessary error states, as I treat many things in our version of ros-z as required, and not backwards compatible optional features. This makes many callsites currently look weird, as I did not change API when removing features. I tried to keep the PR as small as possible. First removing this weird internal failed state, whether the error is returned once, or multiple times, can be changed in a follow-up PR. Also the other functions I did not semantically touch and change. This PR should be a first step of cleaning that up.
If you agree, I'd like to track your suggestions in follow-up issues, and small iterative PRs. What do you think?
I could alternatively collapse this with the next 5 PRs (some of them already open), and then many things look less weird. But I think this defeats the purpose of small reviewable atomic PRs.

crates/ros-z/src/node.rs
schmidma: No, the API contract of providing a schema is fallible itself.
Concrete failure cases include:
- Message::type_name() returns an empty or invalid schema name.
- A manual MessageSchema impl creates conflicting definitions for the same name.
- A schema references a named definition that is missing.
- A schema contains unreachable definitions.
- A manual enum schema has no variants.
- Duplicate field or variant names appear in manually-built schema definitions.

crates/ros-z/src/node.rs
schmidma: Cleaned up in #2503

crates/ros-z/src/node.rs
schmidma: Currently true. Yes. But I consider this also a valid failure case.
Additionally, currently https://github.com/HULKs/hulk/blob/cb20898b23f70d71e64afcf23891f9860da87e7e/crates/ros-z-schema/src/hash.rs#L45 contains a panic, which is replaced with an error in #2503 , making `compute_hash` and thus callsites like `service_type_info()` fallible

crates/ros-z/src/node.rs
knoellle: I get that creating a schema object manually can return a runtime error. But this trait function isn't expected to have parameters or any other side effects, right? So, apart from allocations, the function could even be static. I don't see why that kind of function should return a runtime error type.
I think both derived and manual implementations of this trait should fallibly create their message schema, then `expect` it. Runtime panics can then easily be avoided by adding a test that simply calls the schema function once.
When I compile the program once, the function will either never fail or it will fail every time.

crates/ros-z/src/node.rs
schmidma: See my general comment: I agree and would like to bump this to future work...

crates/ros-z/src/node.rs
schmidma: Okay, I can see that. How do you imagine this "test" for successful schema generation? Is there another `try_schema` function that actually serves the Result? And who is responsible of creating and calling these functions? Would you then also make the `schema()` function internally expect, such that the API changes to an infallible surface? Or should every caller take care of `.expect`ing the schema?

This also sounds like a followup issue btw ;)

knoellle: - The test just calls the function. If it panics, the test fails :) Can derive macros generate tests? 🤔
- I'm not sure when a `try_*` variant would be useful on a (hopefully) deterministic function if you also have a unit test
- Yes, I think `schema` should not return a result. This forces manual implementers using fallible APIs internally to ensure that their implementation is correct. `Result` should be reserved for errors resulting from incorrect usage or unexpected environmental conditions, not for internal programming errors IMHO.

schmidma: hmmm... But then there is this implicit assumption, that everyone implements a proper test case for the fallible schema... This could still lead to errors at runtime. I first got your suggestion as a test once during runtime...
So you'd completely drop error handling at runtime and panic.

Generating test cases with derive macros sounds really weird to me...

oleflb: I completely agree with @knoellle. Testing is a separate topic, since we can write tests for both panics and Results.
In our context, a runtime error isn't better. It just adds verbosity to the API for users who expect every type to have a valid schema.

The optimal solution would be checking the schema at compile-time using a `const` function. But I also see this is probably out-of-scope for now.

schmidma: I implemented the discussed infallible `Message::schema()` in #2503

crates/ros-z/src/node.rs
schmidma: Now, as discussed, this is strictly infallible and panicking. Tests or compile-time tests are future iteration

crates/ros-z/src/node.rs
schmidma: Tracked in https://github.com/HULKs/hulk/issues/2513

knoellle (approved):
LGTM

PR #2511, opened 2026-05-14
expose zenoh session from ros-z context

schmidma:
What it says... allows users to interact with the raw zenoh session

cleanup along the way: `zenoh::Session` is already cheaply clonable, we do not need an additional `Arc`


alexschmander (approved):
LGTM

Issue #2426, opened 2026-04-15
Leaving the City


pejotejo:
This issue tracks the nodes, that still need to be ported.

## WorldState
- [ ] active_vision
- [x] ball_filter
- [ ] ball_state_composer
- [ ] behavior
- [x] button_event_handler
- [x] camera_matrix_calculator
- [ ] fake_odometry
- [ ] game_controller_filter
- [ ] game_controller_state_filter
- [x] ground_provider
- [ ] kinematics_provider
- [ ] led_handler
- [ ] localization
- [ ] obstacle_filter
- [ ] obstacle_receiver
- [x] primary_state_filter
- [x] robot_mode_handler
- [ ] rule_obstacle_composer
- [ ] safe_pose_checker
- [ ] search_suggestor
- [ ] team_ball_receiver
- [ ] time_to_reach_kick_position
- [ ] world_state_composer
- [ ] world_to_field_provider
## Object Detection
- [x] object_detection
## Vision
- [ ] field_border_detection
- [ ] image_segmenter
- [ ] line_detection
- [ ] segment_filter
## Network:
- [ ] network
## Motion:
- [ ] look_at
- [ ] head_motion
- [ ] look_around
- [ ] rotate_head
- [x] kick
- [x] walking
- [ ] stand_up
- [ ] command_sender
- [ ] ~motor_commands_collector~
## Additionally:
- [ ] falldownstate
- [ ] ButtonEvent
- [ ] Odometry

PR #2421, opened 2026-04-02
Kinematics based support foot estimator

alexschmander:
## Why? What?

We currently always calculate the ground_to_robot through the kinematics of the left foot, even when the right foot is the only support foot. This PR adds a kinematics based support foot estimator, to allow the ground provider to calculate the ground_to_robot based on the correct support foot.  

- Fixes #2156 

## How to Test

- Upload to the robot. 
- In twix look at `WorldState.main_outputs.support_foot`
- Let the robot walk

PR #2519, opened 2026-05-15
Port camera matrix calculator and intermediate nodes

alexschmander:
## Why? What?

Ports the nodes `kinematics_provider`,  `support_foot_estimator`, `ground_provider` and `camera_matrix_calculator`. 

- Depends on #2421 
- Part of #2426 

## ToDo / Known Issues

- [x] Use caches to match inputs
- [x] Support foot estimator
- [x] Camera matrix calculator

## How to Test

- Upload
- Use the twix ros z twix to view the output in the text panel or in rosz cli

### DIFF
--- a/crates/nodes/kinematics_provider/src/lib.rs
+++ b/crates/nodes/kinematics_provider/src/lib.rs
@@ -1,9 +1,28 @@
-use std::{future::pending, sync::Arc};
+use std::sync::Arc;
 
 use color_eyre::Result;
 
-use booster::MotorState;
-use kinematics::{joints::Joints, robot_kinematics::RobotKinematics};
+use booster::{JointsMotorState, MotorState};
+use kinematics::{
+    forward::{
+        head_to_neck, left_ankle_to_left_tibia, left_foot_to_left_ankle,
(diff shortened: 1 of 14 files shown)


crates/nodes/kinematics_provider/src/lib.rs
```diff
@@ -12,20 +31,133 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
+    let robot_kinematics_pub = node
         .publisher::<RobotKinematics>("robot_kinematics")
         .into_eyre()?
         .build()
         .await
         .into_eyre()?;
 
-    pending::<()>().await;
+    loop {
+        tokio::select! {
```
oleflb: single arm in select

oleflb (changes requested):
There are a few more pattern matching things which could be improved but that's only aesthetics for now. Therefore only one comment

crates/nodes/kinematics_provider/src/lib.rs
alexschmander: Done

oleflb (approved):

PR #2521, opened 2026-05-16
Future Queue/Map

oleflb:
## Why? What?

__While reviewing, only consider the new state. The old state contains a lot of slop.__
__Disclaimer: Documentation is AI generated__

Implements a FutureQueue/Map for the new framework.
There are 3 components to this:
- An `AnnouncingPublisher`: this is a variant of the publisher in addition to sending data on `/some/topic` also sends an announcement on `/some/topic/annoucement`. That announcement can be sent earlier than the actual data and carries the source timestamp of the data. It implements a RAII pattern to prevent misuse
- A `FutureQueue` is the complement of the `AnnouncingPublisher` and can understand the announcements it sends. Its purpose is to guarantee in-order delivery for the same topic.
- A `FutureMap` is what we referred to as future queues in our old framework. It fuses multiple `FutureQueue`s and on `recv` it returns an object carrying `persistent` and `temporary` data. Their meaning is the same as before, only the timestamp in the `BTreeMap` now is that from the announcement and the format is a bit different since multiple inputs can be received at once.

## ToDo / Known Issues

- A cache which uses the safe time known by the future map to automatically size itself.

## How to Test

- Please take a good look at the public API
- Look at the integration tests


crates/ros-z-streams/src/announce.rs
```diff
@@ -1,44 +1,43 @@
...
-/// Publisher that emits lightweight announcements before payload transmission.
-pub struct AnnouncingPublisher<T>
-where
-    T: Message,
-    for<'a> T::Codec: WireEncoder<Input<'a> = &'a T>,
-{
+/// Publisher that emits announcements before sending payload data.
+///
+/// This publisher coordinates with its corresponding data channel to ensure that
+/// timestamps are announced before heavy payloads are transmitted. This two-phase
```
knoellle: I'm a bit confused by the emphasis on lightweight and heavy payloads.
Sure, a larger message size may lead to an increased delay between the data timestamp and when the message is received, but that should be negligible compared to the processing delay we're trying to combat with this feature.

crates/ros-z-streams/src/announce.rs
```diff
@@ -47,188 +46,75 @@ where
...
+/// The announcement and payload must be published in order—the announcement first,
+/// then the payload via this handle. If this handle is dropped without publishing,
+/// an error is logged.
 #[must_use = "data must be published before the announcement is dropped"]
-pub struct PendingAnnouncement<'a, T>
-where
-    T: Message,
-    for<'b> T::Codec: WireEncoder<Input<'b> = &'b T>,
-{
+pub struct PendingAnnouncement<'a, T: Message> {
```
knoellle, line 96: This type is not publicly accessible despite being part of the public API. Consider pub-using it in `lib.rs`

crates/ros-z-streams/src/announce.rs
```diff
@@ -47,188 +46,75 @@ where
...
+    ///
+    /// This completes the announcement-payload cycle by sending the actual data.
+    /// The payload is published with the same message ID as the prior announcement,
+    /// allowing receivers to correlate them and maintain temporal ordering.
     pub async fn publish(mut self, value: &T) -> ZResult<()> {
-        let data = self.data.take().expect("pending announcement data missing");
+        let data = self.data.take().expect("can only publish once");
         data.publish(value).await?;
-        self.completed = true;
         Ok(())
```
knoellle:
```rust
    data.publish(value).await
```

crates/ros-z-streams/src/future_queue.rs
```diff
@@ -1,299 +1,153 @@
...
+///
+/// This struct maintains two collections:
+/// - `inflight`: A set of announced timestamps awaiting corresponding data
+/// - `pending_data`: A queue of received messages awaiting matching announcements
+///
+/// The queue enforces a strict ordering: data is only released when both the
+/// announcement has arrived and the data payload has been received, ensuring
+/// temporal consistency even with variable network delays and out-of-order delivery.
+#[derive(Debug)]
+pub struct FutureQueueSubscriber<T: Message> {
```
knoellle, line 26: This type lacks an interface to receive announcements to propagate them.

Consider the following processing chain with future queues in between:
A -> B -> C -> D
Lets say B has 100 ms processing delay. When A publishes new data at time `t`, B announces timestamp `t` to C but C doesn't do anything yet.
Only once B publishes does C announce to D with timestamp `t` at time `t + 100 ms + network delay`.
If I understand correctly, the safety delay is only meant to compensate the network delay part, so D has no idea it has to buffer its other inputs received between `t` and `t + 100 ms + network delay`

knoellle (commented):
Partial review

crates/ros-z-streams/src/future_map.rs
```diff
@@ -194,148 +231,88 @@ impl<'a> FutureMapBuilder<'a, ()> {
...
-                                        return Ok(self.finalize_receive(stream_states, global_safe_time));
-                                    }
+            async fn receive_event(
+                &mut self,
+                buffer: &mut BTreeMap<Time, Self::Output>,
+            ) -> ZResult<GroupEvent> {
+                tokio::select! {
+                    $(
+                        result = self.$index.recv() => {
+                            // Match the QueueEvent here!
```
knoellle: What's the point of this comment?

crates/ros-z-streams/src/future_map.rs
```diff
@@ -194,148 +231,88 @@ impl<'a> FutureMapBuilder<'a, ()> {
...
+                            match result? {
+                                QueueEvent::Data(time, value) => {
+                                    let entry = buffer.entry(time).or_default();
+                                    entry.$index = Some(value);
```
knoellle, line 262: This allows silently overwriting the same entry. Is that intended behavior?

crates/ros-z-streams/src/future_queue.rs
```diff
@@ -1,299 +1,153 @@
...
+                announcement.time.min(transit_boundary)
+            })
     }
 
-    /// Return current stream state without consuming events.
-    pub fn current_state(&self) -> QueueState {
-        self.queue_state(None)
+    /// Returns the transit lag for this stream.
+    pub(crate) fn safety_lag(&self) -> Duration {
+        self.transit_lag
```
knoellle: Why does the accessor function have a different name than the field?

crates/ros-z-streams/src/announce.rs
oleflb: done

crates/ros-z-streams/src/announce.rs
oleflb: done

crates/ros-z-streams/src/future_queue.rs
oleflb: postponed

crates/ros-z-streams/src/future_map.rs
oleflb: Of course this is debatable. But I think it makes sense that data may change if two data events are received for the same timestamp

crates/ros-z-streams/src/future_queue.rs
oleflb: missed during rename

crates/ros-z-streams/src/announce.rs
oleflb: done

knoellle (approved):
Lets Gamble, Try Merge

PR #2522, opened 2026-05-17
Cleanup parameters

alexschmander:
## Why? What?

Cleans up the current state of parameters. Removes unnecessary Parameter wrapper structs in nodes where only  a single other Parameter struct is contained inside it. Also moves the shared parameters `player_number` and `field_dimensions` to a node supplying them as transient local topics, allowing a single source of truth for these. 

## How to Test


pejotejo (approved):
lgtm

PR #2523, opened 2026-05-17
Port network and game controller filters

alexschmander:
## Why? What?

This PR ports the `message_filter`, `game_controller_filter` and `game_controller_state_filter`.

- Part of #2426
- Depends on #2522  
- Depends on #2521 
- Depends on #2534 
- Depends on #2535 

## Discussion

- [x] Do the inputs of message filter and game controller state filter need to be future queues/maps?

## How to Test

- Upload
- Use the twix ros z twix to view the output in the text panel or in rosz cli

### DIFF
--- a/crates/nodes/game_controller_filter/src/lib.rs
+++ b/crates/nodes/game_controller_filter/src/lib.rs
@@ -1,14 +1,11 @@
 use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
 
 use color_eyre::Result;
-use ros_z_streams::CreateFutureMapBuilder;
+use ros_z_streams::{CreateAnnouncingPublisher, CreateFutureMapBuilder};
 use serde::{Deserialize, Serialize};
 
 use hsl_network_messages::GameControllerStateMessage;
-use ros_z::{
-    prelude::*,
-    time::{Clock, Time},
-};
+use ros_z::{prelude::*, time::Time};
 use types::{
     field_dimensions::GlobalFieldSide,
     game_controller_state::GameControllerState,
@@ -39,8 +36,7 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
         .build()
         .await?;
     let game_controller_state_pub = node
-        .publisher::<Option<GameControllerState>>("game_controller_state")?
-        .build()
+        .announcing_publisher::<Option<GameControllerState>>("game_controller_state")
         .await?;
     let game_controller_address_pub = node
         .publisher::<Option<SocketAddr>>("game_controller_address")?
@@ -55,7 +51,7 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
 
         let network_message = network_message_sub.recv().await?;
 
-        for (_, source_address, message) in
+        for (time, source_address, message) in
             network_message
                 .persistent
                 .iter()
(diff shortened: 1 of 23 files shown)


crates/nodes/game_controller_filter/src/lib.rs
```diff
@@ -22,25 +21,134 @@ pub struct Parameters {
 pub async fn run(ctx: Arc<Context>) -> Result<()> {
     let node = ctx.create_node("game_controller_filter").build().await?;
 
-    let _parameters = node.bind_parameter_as::<Parameters>("game_controller_filter")?;
-    let _network_message_sub = node
-        .subscriber::<IncomingMessage>("filtered_message")?
+    let parameters = node.bind_parameter_as::<Parameters>("game_controller_filter")?;
+    let mut network_message_sub = node
+        .create_future_map_builder()
+        .create_future_subscriber::<IncomingMessage>("filtered_message", Duration::from_millis(1))
```
oleflb: Why would you want a `FutureMap` with a single input? I don't think that is required here unless you assume out of order delivery. In that case you could also use the `FutureQueue` directly (built for a single subscription). Although currently it will also tell you announcements, so maybe you would have to rename the current `recv` to `recv_with_announcements` and create a second `recv` which only returns data

crates/nodes/game_controller_filter/src/lib.rs
```diff
@@ -22,25 +21,134 @@ pub struct Parameters {
...
+            let last_address = game_controller_filter
+                .last_contact
+                .iter()
+                .max_by_key(|(_address, time)| *time)
+                .map(|(address, _time)| *address);
+
+            pending_annoucement
+                .publish(&game_controller_filter.game_controller_state)
+                .await?;
+            game_controller_address_pub.publish(&last_address).await?;
```
oleflb: Are you really sure a future queue is needed for this?

crates/nodes/game_controller_state_filter/src/lib.rs
```diff
@@ -45,16 +59,491 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
         .build()
         .await?;
 
-    let _whistle_in_set_ball_position_pub = node
-        .publisher::<Point2<Field>>("whistle_in_set_ball_position")?
+    let mut game_controller_state_sub = node
```
oleflb: Same here, future map with one topic doesn't really make sense

crates/nodes/game_controller_state_filter/src/lib.rs
```diff
@@ -45,16 +59,491 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
+                if let Some(actual_ball_state) = current_ball_state {
+                    latest_ball_state = Some((received_source_time, actual_ball_state));
+                }
+            }
+            received_field_dimensions = field_dimensions_sub.recv() => {
+                field_dimensions = Some(received_field_dimensions?)
+            }
+            received_player_number = player_number_sub.recv() => {
+                player_number = Some(received_player_number?)
+            }
```
oleflb: I feel like we could greatly simplify this if we had a watch receiver equivalent for a channel. Then you would not need to have all these mutable variables flying around

oleflb (changes requested):

crates/nodes/game_controller_state_filter/src/lib.rs
alexschmander: Implemented as Caches

crates/nodes/game_controller_state_filter/src/lib.rs
alexschmander: Removed

crates/nodes/game_controller_filter/src/lib.rs
alexschmander: Removed

crates/nodes/game_controller_filter/src/lib.rs
alexschmander: Removed

oleflb (dismissed):

oleflb (approved):

PR #2524, opened 2026-05-17
Port led handler

alexschmander:
## Why? What?

What the title says.

- Part of #2426 
- Depends on #2522 
- Depends on #2521 
- Depends on #2535 
- Depends on #2523 
- Depends on #2527 

## How to Test

- Upload
- Change the primary state and see if the led on the K1 changes.


pejotejo (approved):
leuchtet good to me

PR #2526, opened 2026-05-18
reject ambiguous schema

schmidma:
Based on #2503

Rejects ambiguous schema candidates during schema discovery, also tries to clean up some of the code to sligthly increase code quality

Testing should mostly be done by the unit tests, but in theory, you may start two publishers with the same topic name but different types, then in a third terminal, run `rosz echo /my_topic`, this should fail now


alexschmander (approved):
LGTM and wörks

PR #2527, opened 2026-05-19
Port primary state filter

alexschmander:
## Why? What?

*Describe here why you created this PR and what it does*

- Part of #2426 
- Depends on #2522 
- Depends on #2521 
- Depends on #2535 
- Depends on #2523 

## ToDo / Known Issues

- [x] `safe_pose_checker`
- [x] `primary_state_filter`
- [x] ~~Has very weird inconsitency issues with the future queue. A Buttons press sometimes shows up in the temporaries, but then not in the persistents. It seems like in the mean time, another None value with the same timestamp shows up and takes the place of the original Some value... This leads to the buttons only flakily working. As a workaround, I also use the temporaries, and keep track of the most recently processed time stamp. Any values older than that timestamp are filtered out.~~ Not using future queues here anymore.

## How to Test

- Upload
- Open twix ros z or rosz
- Press buttons on the robot, receive game controller message by starting the game controller and/or inject a primary state via the parameter. All of these should change the published primary state.

### DIFF
--- a/crates/nodes/button_event_bridge/src/lib.rs
+++ b/crates/nodes/button_event_bridge/src/lib.rs
@@ -1,9 +1,13 @@
 use std::sync::Arc;
 
-use color_eyre::{Result, eyre::Context as _};
+use color_eyre::{
+    Result,
+    eyre::{Context as _, eyre},
+};
 
 use booster::ButtonEventMsg;
 use ros_z::prelude::*;
+use ros_z_streams::CreateAnnouncingPublisher;
 
 pub async fn run(ctx: Arc<Context>) -> Result<()> {
     let node = ctx.create_node("button_event_bridge").build().await?;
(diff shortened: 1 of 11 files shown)


crates/nodes/button_event_bridge/src/lib.rs
```diff
@@ -15,23 +19,22 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
                     .wrap_err("deserialization failed")?;
 
-                button_event_message_pub
+                pending_accouncement
                     .publish(&button_event)
-                    .await?;
-
-
+                    .await
+                    ?;
```
oleflb: ih baeaeaeh

oleflb (changes requested):
generally looks good, but there are some rebase errors in here

crates/nodes/button_event_handler/src/lib.rs
```diff
@@ -1,23 +1,58 @@
...
-        .await?;
-    let _buttons_pub = node
-        .publisher::<Buttons<Option<ButtonPressType>>>("buttons")?
-        .build()
+    let mut button_event_future_map = node
+        .create_future_map_builder()
+        .create_future_subscriber::<ButtonEventMsg>(
+            "inputs/button_event_message",
+            Duration::from_millis(1),
+        )
```
oleflb: single subscriber

oleflb (dismissed):

crates/nodes/button_event_handler/src/lib.rs
alexschmander: Changed to normal sub

crates/nodes/button_event_bridge/src/lib.rs
alexschmander: Fixed

pejotejo (approved):
Alex macht mich sonst fertig

PR #2529, opened 2026-05-20
Update behavior and search suggestor parameters

alexschmander:
What the title says. Fixes the current main state, such that hulk_ros_z works again.


pejotejo (approved):
lgtm

PR #2532, opened 2026-05-22
Remote Controlled Kick

pejotejo:
## Why? What?
Adds a simple kick that can be triggered via the remote control in twix. It tries to shoot 3m in front of it when triggered. It is build for demos, but the setting the target right in front can be used for kick testing in the future.
 
## How to Test

1. deploy it to a robot
2. open twix
3. activate the remote_control.enable parameter (optionally disable allow_schlong)
5. enable the controller and press a
6. profit


BenSampaolo (approved):
LGTM!

PR #2535, opened 2026-05-23
Timestamped kinematics

alexschmander:
## Why? What?

This PR adds time stamps to outputs that are derived from imu and motor states. Nodes using these outputs have their caches built with these time stamps, to allow for matching of different kinematics output to the source data timestamp.  

- Depends on #2534 

## How to Test


oleflb (changes requested):
Use wrapper struct for time

alexschmander:
@oleflb added time wrapper

oleflb (approved):

PR #2538, opened 2026-05-24
Port `robot_mode_handler`

### LINKED ISSUE #2528: Port Robot_Mode_Handler
Port the `robot_mode_handler` to the new framework.

paudar:
## Why? What?

Port `robot_mode_handler` to the new framework. 

- Part of #2426 
- Depends on #2527 
- Fixes #2528

## ToDo / Known Issues

- test how often high-level_commands are send

## How to Test

- Uplad
- Press Buttons and change the gamestate 
- observe messages send

### DIFF
--- a/crates/nodes/robot_mode_handler/Cargo.toml
+++ b/crates/nodes/robot_mode_handler/Cargo.toml
@@ -10,5 +10,5 @@ booster_sdk_interface = { workspace = true }
 color-eyre = { workspace = true }
 ros-z = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
-tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }  # TODO: what features are required?
+tokio = { workspace = true }
 types = { workspace = true }
--- a/crates/nodes/robot_mode_handler/Cargo.toml
+++ b/crates/nodes/robot_mode_handler/Cargo.toml
@@ -10,4 +10,5 @@ booster_sdk_interface = { workspace = true }
 color-eyre = { workspace = true }
 ros-z = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }  # TODO: what features are required?
 types = { workspace = true }
--- a/crates/nodes/robot_mode_handler/src/lib.rs
+++ b/crates/nodes/robot_mode_handler/src/lib.rs
@@ -1,9 +1,12 @@
-use std::{future::pending, sync::Arc, time::Duration};
+use std::{
+    sync::Arc,
+    time::{Duration, SystemTime},
+};
 
 use color_eyre::Result;
 use serde::{Deserialize, Serialize};
 
-use booster_sdk_interface::GetRobotMode;
+use booster_sdk_interface::{GetRobotMode, GetRobotModeRequest, HighLevelCommand, RobotMode};
 use ros_z::prelude::*;
 use types::{
     buttons::{ButtonPressType, Buttons},
@@ -11,7 +14,6 @@ use types::{
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, Message)]
-#[serde(deny_unknown_fields)]
 pub struct Parameters {
     pub wait_before_prepare: Duration,
     pub remote_stop_toggle: bool,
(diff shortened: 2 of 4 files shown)


crates/nodes/robot_mode_handler/Cargo.toml
```diff
@@ -10,4 +10,5 @@ booster_sdk_interface = { workspace = true }
 color-eyre = { workspace = true }
 ros-z = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }  # TODO: what features are required?
```
alexschmander: This crate does not require any non-default tokio features. Remove all please

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
+    let mut last_primary_state = PrimaryState::default();
+    let mut last_primary_state_change_time = SystemTime::UNIX_EPOCH;
+
+    loop {
+        let parameters = parameters.snapshot().typed().clone();
+
+        let robot_mode = get_robot_mode_client
+            .call_async(&GetRobotModeRequest {})
+            .await?
+            .robot_mode;
```
alexschmander: Calling this before the select statement, makes every branch of the select wait for this call. But only the primary state branch actually needs the robot mode. Please move this call into the primary state branch.

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
         .build()
         .await?;
 
-    pending::<()>().await;
+    let mut local_stop_toggle = false;
+    let mut last_primary_state = PrimaryState::default();
+    let mut last_primary_state_change_time = SystemTime::UNIX_EPOCH;
+
+    loop {
+        let parameters = parameters.snapshot().typed().clone();
```
alexschmander:
```suggestion
        let parameters_snapshot = parameters.snapshot();
        let parameters = parameters_snapshot.typed();
```

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
+
+                let should_enter_damping_mode = local_stop_toggle != parameters.remote_stop_toggle;
+
+                if should_enter_damping_mode && is_local_stop_toggle_short_press {
+                    local_stop_toggle = !local_stop_toggle;
+                }
+
+                if should_enter_damping_mode {
+                    change_mode(&high_level_command_pub, RobotMode::Damping).await;
+
```
alexschmander: Remove newline

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
+                }
+            }
+            primary_state = primary_state_sub.recv() => {
+                let primary_state = primary_state?;
+
+                if primary_state != last_primary_state {
+                    last_primary_state_change_time = SystemTime::now();
```
alexschmander: `SystemTime::now()` should not be used in the robotics code! Use `node.clock().now()`.

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
+                    last_primary_state = primary_state;
+                }
+
+                let time_since_primary_state_change = SystemTime::now()
```
alexschmander: Here as well

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -20,21 +22,97 @@ pub struct Parameters {
...
+
+                }
+            }
+            primary_state = primary_state_sub.recv() => {
+                let primary_state = primary_state?;
+
+                if primary_state != last_primary_state {
+                    last_primary_state_change_time = SystemTime::now();
+                    last_primary_state = primary_state;
+                }
```
alexschmander, line 77: The problem with always sending high level commands could be solved by only processing the primary state, if it has changed.
```suggestion
                if primary_state != last_primary_state {
                    last_primary_state_change_time = SystemTime::now();
                    last_primary_state = primary_state;
                } else {
                    continue;
                }
```

alexschmander: This has the problem, that the switch to prepare logic will not work anymore. 
I think this can be managed by checking for both that the primary state has changed and you are still inside the wait_for_prepare duration.

alexschmander (changes requested):
A couple of improvements, otherwise works! :)

crates/nodes/robot_mode_handler/src/lib.rs
paudar: Done 👍🏻

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -68,20 +61,26 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
+                } else {
+                    continue;
                 }
 
-                let time_since_primary_state_change = SystemTime::now()
-                    .duration_since(last_primary_state_change_time)
-                    .expect("time ran backwards");
+                let time_since_primary_state_change = node.clock().now()
+                    .duration_since(last_primary_state_change_time);
+                    // TODO: find eqivalent for: .expect("time ran backwards");
```
alexschmander: This is not needed with `ros_z::Time`. Remove this comment

crates/nodes/robot_mode_handler/src/lib.rs
alexschmander: > This has the problem, that the switch to prepare logic will not work anymore. I think this can be managed by checking for both that the primary state has changed and you are still inside the wait_for_prepare duration.

What about this?

alexschmander (changes requested):
Still two things

crates/nodes/robot_mode_handler/src/lib.rs
paudar: Good to know, done

crates/nodes/robot_mode_handler/src/lib.rs
paudar: I have added the change but still have to test it :)

paudar: I have decided to not change the `RobotMode` x seconds after changing to Finished or Penalized. Now the robot-handler has to press a button before lifting the robot.

My tests looked very good

alexschmander (approved):
LGTM

PR #2542, opened 2026-05-27
Add closest to ball logic

JackyBloxx:
## Why? What?

adds logic for the closest to ball calculation in the behavior
- Fixes #

## How to Test

In playing with at least two Robots: If one robot sees a ball is should only engage with the ball if it is the closest to it and the other should walk to its calculatetd centroid position.

### DIFF
--- a/crates/world_state/src/behavior/condition.rs
+++ b/crates/world_state/src/behavior/condition.rs
@@ -4,6 +4,7 @@ use linear_algebra::{point, vector};
 use types::{
     filtered_game_controller_state::FilteredGameControllerState, primary_state::PrimaryState,
 };
+use voronoi::Ownership;
 
 use crate::behavior::node::Blackboard;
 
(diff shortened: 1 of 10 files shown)


crates/world_state/src/behavior/condition.rs
```diff
@@ -88,9 +88,17 @@ pub fn is_close_to_ball_aligned(blackboard: &mut Blackboard) -> bool {
...
+        && grid.ownership_at(ball.position)
+            == Some(voronoi::Ownership::Robot(
+                blackboard.world_state.robot.player_number,
+            ))
+    {
+        true
+    } else {
+        false
+    }
 }
```
pejotejo, line 149: this would probably lead to rapid striker switching when the robots are close together or the ball is moved alongside the border. You probably want to have some kind of hysteresis. Maybe look into the old time to riech for inspiration.

crates/world_state/src/behavior/tree.rs
```diff
@@ -76,7 +76,11 @@ fn playing_subtree() -> Node<Blackboard> {
             negation!(condition!(has_ball_position)),
             subtree!(search_subtree)
         ),
-        sequence!(condition!(is_closest_to_ball), subtree!(striker_subtree)),
+        sequence!(
+            action!(calculate_voronoi_grid),
+            condition!(is_closest_to_ball),
```
pejotejo, line 80: wouldn't this lead to the behavior, that if it cant calculate the voronoi grid, it trys to walk to the voronoi position? this would be bad imo. maybe switch striker and supporter subtree and negate the closest to ball condition. Then it would default to striker if it cant calculate the grid.

crates/world_state/src/behavior/condition.rs
JackyBloxx: done

crates/world_state/src/behavior/tree.rs
JackyBloxx: done

crates/world_state/src/behavior/condition.rs
pejotejo: wouldnt a distance based hysteresis be better?

pejotejo (approved):
lgtm

PR #2547, opened 2026-05-27
Transient local primary state

alexschmander:
## Why? What?

Change the Qos of the Primary State publisher and subsribers to transient local, since it is an event triggered output and this allow nodes to always have an initial value to go off from.


pejotejo (approved):
lgtm

PR #2549, opened 2026-05-29
fix(hulk_ros_z): surface startup failures

schmidma:
## Summary
- surface monitor task errors and panics from hulk_ros_z startup
- skip graceful ROS-Z context shutdown on failed startup/runtime paths
- bound Tokio runtime shutdown so blocking tasks cannot hang process exit

## Test Plan
- cargo fmt --all -- --check
- cargo nextest run -p hulk_ros_z
- cargo build -p hulk_ros_z
- timeout -v --kill-after=2s 8s target/debug/hulk_ros_z --robot alpha --location lab --parameter-root etc/parameters/ros_z/ --router tcp/10.4.24.15:7448


alexschmander (approved):
Very nice! LGTM

PR #2550, opened 2026-05-29
Port localization to ros-z

schmidma:
## Summary
- Ported localization runtime to ros-z with `FutureMap` inputs and optional `ground_to_field` output.
- Moved the old localization algorithm into the localization node while preserving ordered sensor and perception input handling.
- Consumes the plain filtered game-controller state schema now available on `main`.

## Test Plan
- [x] `cargo nextest run`
- [x] `cargo test --doc`

### DIFF
--- a/crates/nodes/localization/src/algorithm.rs
+++ b/crates/nodes/localization/src/algorithm.rs
@@ -15,7 +15,6 @@ use geometry::line_segment::LineSegment;
 use hsl_network_messages::{GamePhase, Penalty, PlayerNumber, SubState, Team};
 use linear_algebra::{IntoTransform, Isometry2, Point2, Pose2, distance, point};
 use nalgebra::{Matrix2, Matrix3, Rotation2, Vector2, Vector3, matrix};
-use ordered_float::NotNan;
 use serde::{Deserialize, Serialize};
 use types::{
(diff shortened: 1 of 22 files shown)


crates/nodes/localization/src/algorithm.rs
```diff
@@ -0,0 +1,1830 @@
...
+        let odometer_with_imu_yaw = current_odometer.map(|Odometer { x, y, theta: _ }| Odometer {
+            x,
+            y,
+            theta: imu_state.roll_pitch_yaw.z(),
+        });
+        let current_odometry_to_last_odometry = match (self.last_odometer, odometer_with_imu_yaw) {
+            (Some(last), Some(latest)) => odometry_delta(last, latest),
+            _ => Default::default(),
+        };
+        self.last_odometer = odometer_with_imu_yaw;
```
paudar, line 257: According to my findings, incomplete batches aren't currently being handled properly here; if a `last_...` state is maintained until the next value arrives, the whole thing works much better.

paudar (commented):
I am aware that the branch is still a WIP, but as we would like to use it in a test match today, I have already had a look at the PR. 

I have only found one bug:

paudar:
closed in favor of #2622

Issue #2553, opened 2026-05-29
Rewrite obstacle filter by splitting for humanoids and stationary objects


alexschmander:
Our current obstacle filter has shoehorned moving objects like people and robots. These do not fit the filter type, since it a basic positional kalman filter. This should be split up into separate filters, one handling moving humanoid entities like humans and robots and one for handling stationary objects like goalposts and field features. After they are filtered, they may be classified as an object, depending on the object type. Additionally, these filters could feature instance tracking.

PR #2554, opened 2026-05-29
Port obstacle filter

alexschmander:
## Why? What?

Ports the obstacle filter as it is semantically implemented currently. This does swap the old obstacle receiver out for the player state receiver, since that does the same but more. In that vein, it also ports the player state receiver. 

- #2553 

## How to Test

- Upload and look at the outputs of the player state receiver and the obstacle filter


crates/nodes/obstacle_filter/src/lib.rs
```diff
@@ -1,76 +1,761 @@
...
+        tokio::select! {
+            received_detections = detections.recv() => {
+                let parameters_snapshot = parameters.snapshot();
+                let parameters = parameters_snapshot.typed();
+                let item = received_detections?;
+
+                for (detection_time, (detected_objects, detected_poses)) in item.persistent {
+                    let Some(field_dimensions) = field_dimensions_cache.get_latest() else {
+                        continue;
+                    };
```
pejotejo: this can be called outside the for loop

alexschmander: Done

crates/nodes/detection/src/lib.rs
```diff
@@ -159,8 +161,11 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
         non_maximum_suppression_duration_pub
             .publish(&non_maximum_suppression_duration)
             .await?;
-        detected_objects_pub.publish(&detected_objects).await?;
-        detected_poses_pub.publish(&detected_poses).await?;
+        let detected_objects_pending = detected_objects_pub.announce(image_time).await?;
+        let detected_poses_pending = detected_poses_pub.announce(image_time).await?;
```
pejotejo: didnt you mention that anouncements should be as early as possible in the function, at best before the compute? This should be moved up then

crates/nodes/player_state_receiver/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+pub async fn run(ctx: Arc<Context>) -> Result<()> {
+    let node = ctx.create_node("player_state_receiver").build().await?;
+    let filtered_game_controller_state_sub = node
+        .subscriber::<TimeWrapper<FilteredGameControllerState>>("filtered_game_controller_state")?
+        .build()
+        .await?;
+    let filtered_message_sub = node
+        .subscriber::<TimeWrapper<IncomingMessage>>("filtered_message")?
```
pejotejo, line 23: i'm not sure with the behavior of tokio select in this case, but if we get two messeges directly after oneanother, couldnt this lead to messages not being processed?

crates/nodes/detection/src/lib.rs
alexschmander: Done

crates/nodes/player_state_receiver/src/lib.rs
alexschmander: These are buffered by zenoh. I think this should not be an issue.

pejotejo (approved):
lgtm

PR #2555, opened 2026-05-29
Cache entry extractor

alexschmander:
## Why? What?

During development of #2554, it came up, that being able to flatten or map structured subscribed types to one or many different time stamped entries in the cache would be nice. This PR implements a `.with_stamped_entries(|value| impl IntoIterator<(Time, Vec<EntryType>)>)` on the cache builder. 

## How to Test

- `pepsi nextest --remote -p ros-z`

### DIFF
--- a/crates/ros-z/src/cache.rs
+++ b/crates/ros-z/src/cache.rs
@@ -3,9 +3,9 @@
 //! [`Cache`](crate::cache::Cache) retains a sliding window of received messages
 //! and lets callers query them by time.
 //!
-//! # Stamp strategies
+//! # Cache population strategies
 //!
-//! Two indexing strategies are available, selected at build time:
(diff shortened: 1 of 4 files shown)


schmidma:
I'm concerned that this changes the semantics of `Cache<T>`. So far, a cache entry represents one received message, with `with_stamp` only changing how that message is indexed. `with_entry_extractor` changes both cardinality and type, so one received `T` can become zero or many cached `U`s that were never published as framework messages.

If these entries represent independent state updates, I think the publisher or an adapter node should publish them as separate messages. That keeps graph visibility, QoS, metadata, ordering, and the cache abstraction aligned.

If we need this for cache-local projections over batched messages, I'd prefer naming/documentation that makes that explicit instead of treating it as another cache population strategy.

PR #2557, opened 2026-05-30
Add goalkeeper subtree

### LINKED ISSUE #2556: implement `goalkeeper_subtree`
Implement a Behavior for our goalkeeper

paudar:
## Why? What?

Adds an initial goalkeeper behavior tree with a dedicated goalkeeper policy outside the generic top-level tree.

- Introduces goalkeeper-specific behavior for:
  - intercepting shots on goal
  - clearing slow balls in the own danger area
  - active defensive positioning near the own goal
  - default goalkeeper positioning
  - first sub-state handling
  - optional last-man-standing behavior
- Moves striker behavior into its own `striker` module so it can be reused by the goalkeeper in last-man-standing mode.
- Adds goalkeeper tuning parameters for positioning, clearing thresholds, active defense, and last-man-standing behavior.
- Refactors VisualKick target setup so normal kicks, remote kick targets, and goalkeeper clear kicks share the same target application logic.

Fixes #2556

## ToDo / Known Issues

- Kick target for interception is not smartly chosen. 
- tune `select_goalkeeper_clear_kick_target` for schlong kicks

- Implement replacement goalkeeper role selection.
- Improve goalkeeper set-play positioning for corners, free kicks, and penalties.
- Add smarter goalkeeper clear targets based on opponent positions.
- Tune goalkeeper positioning parameters on robot/simulation.

## How to Test

Go on the field with a ball and a GameController. Make sure that the player number of the robot is equal to the parameter `goal_keeper_number`.
Or run the behavior simulator with a scenario that has at least one player with payer number "One".


crates/types/src/parameters.rs
```diff
@@ -146,6 +147,13 @@ pub struct RolePositionsParameters {
     pub striker_supporter_minimum_x: f32,
     pub keeper_x_offset: f32,
     pub keeper_passive_distance: f32,
+    pub keeper_clear_ball_maximum_velocity: f32,
+    pub keeper_clear_ball_maximum_robot_distance: f32,
+    pub keeper_visual_kick_hold_duration: Duration,
+    pub keeper_active_defense_x_offset: f32,
+    pub keeper_active_defense_center_forward_offset: f32,
+    pub keeper_active_defense_line_width_half: f32,
+    pub keeper_active_defense_maximum_robot_distance: f32,
```
pejotejo: I think the rest of this parameter struct is not used anymore, so it can be removed if im not mistaken. I also would prefer having the keeper related parameters in an own `KeeperParameters` struct, since there are not only positioning parameters in there

crates/world_state/src/behavior/condition.rs
```diff
@@ -21,20 +24,142 @@ pub fn is_ball_interception_candidate(blackboard: &mut Blackboard) -> bool {
...
+        false
+    }
+}
+
+fn ball_interception_velocity_in_field(blackboard: &Blackboard) -> Option<Vector2<Field>> {
+    let ball = blackboard.ball.as_ref()?;
+    let ground_to_field = blackboard.world_state.robot.ground_to_field?;
+
+    Some(ground_to_field * ball.velocity)
+}
```
pejotejo: it triggers a bit having a function that is not a condition in this file (also all conditions were alphabetically ordered). So either move the ball_velocity function to one end of the file or just remove it

crates/world_state/src/behavior/condition.rs
```diff
@@ -21,20 +24,142 @@ pub fn is_ball_interception_candidate(blackboard: &mut Blackboard) -> bool {
...
+    blackboard.ball.as_ref().is_some_and(|ball| {
+        let own_goal_x = -blackboard.field_dimensions.length / 2.0;
+        let maximum_ball_x =
+            own_goal_x + blackboard.parameters.role_positions.keeper_passive_distance;
+
+        ball.position.x() < maximum_ball_x
+    })
+}
+
+pub fn is_ball_in_own_danger_area(blackboard: &mut Blackboard) -> bool {
```
pejotejo: rename this to own_penalty_area

crates/world_state/src/behavior/kick.rs
```diff
@@ -31,6 +32,40 @@ pub fn kick_subtree() -> Node<Blackboard> {
     )
 }
 
+pub fn goalkeeper_clear_kick_subtree() -> Node<Blackboard> {
```
pejotejo: same here, put it to the goalkeeper and not to the kick, else the kick gets really big

crates/world_state/src/behavior/kick.rs
```diff
@@ -31,6 +32,40 @@ pub fn kick_subtree() -> Node<Blackboard> {
...
+        selection!(
+            sequence!(
+                action!(kick),
+                action!(intercept),
+                action!(use_kick_power, KickPower::Rumpelstilzchen),
+            ),
+            sequence!(
+                condition!(is_goalkeeper_visual_kick_hold_active),
+                subtree!(goalkeeper_clear_kick_action_subtree),
+            ),
```
pejotejo: this part is normally not reached, since the actions before should not fail

crates/world_state/src/behavior/condition.rs
```diff
@@ -21,20 +24,142 @@ pub fn is_ball_interception_candidate(blackboard: &mut Blackboard) -> bool {
...
+pub fn is_ball_suspiciously_behind_own_goal(blackboard: &mut Blackboard) -> bool {
+    blackboard.ball.as_ref().is_some_and(|ball| {
+        let field_dimensions = blackboard.field_dimensions;
+        let own_goal_x = -field_dimensions.length / 2.0;
+
+        ball.position.x() < own_goal_x - field_dimensions.goal_depth
+    })
+}
+
+pub fn is_goalkeeper_clear_candidate(blackboard: &mut Blackboard) -> bool {
```
pejotejo: please rename, clear_candidate is a weird name (also all the other clear functions)

crates/world_state/src/behavior/goalkeeper.rs
```diff
@@ -0,0 +1,141 @@
...
+        sequence!(
+            condition!(is_sub_state, SubState::PenaltyKick),
+            negation!(condition!(hulks_is_kicking_team)),
+            subtree!(goalkeeper_penalty_position_subtree)
+        ),
+        sequence!(
+            condition!(hulks_is_kicking_team),
+            condition!(is_sub_state, SubState::GoalKick),
+            subtree!(goalkeeper_clear_or_walk_to_ball_subtree)
+        ),
```
pejotejo: this is in conflict with the striker behavior, the striker would also try to kick this ball, since it is the striker and has a right to do so. This would probably also apply for any other kick situation. Somehow the goal needs to claim striker for these. Maybe talk to phillip to this, he is currently doing the separation part between field players. Maybe the striker subtree in your goalie subtree then is redundant. Same for the substates that kick the ball away.

crates/world_state/src/behavior/condition.rs
```diff
@@ -21,20 +24,142 @@ pub fn is_ball_interception_candidate(blackboard: &mut Blackboard) -> bool {
...
-            && ball_is_moving
             && ball_is_moving_towards_robot
-            && ball_is_moving_towards_own_half
+            && is_interception_velocity_towards_own_half(blackboard, ball_in_field_velocity)
+    } else {
+        false
+    }
+}
+
+pub fn is_goalkeeper_interception_candidate(blackboard: &mut Blackboard) -> bool {
```
pejotejo: move all the goal keeper specific conditions into the goalkeeper.rs

crates/world_state/src/behavior/kick.rs
```diff
@@ -31,6 +32,40 @@ pub fn kick_subtree() -> Node<Blackboard> {
...
+        ),
+        subtree!(kick_alternatives_subtree),
+    )
+}
+
+fn goalkeeper_clear_kick_action_subtree() -> Node<Blackboard> {
```
pejotejo: subtrees are great to remove code duplication, but they make it just unecessarily harder to understand whats happening and also if the subtree is changed at some point could have some implications on does not consider when changing, i would not do it to much down. Twenty subtrees in im not sure what is happening around this.

crates/world_state/src/behavior/kick.rs
pejotejo: also isnt the second half of the subtree just triggered one branch further left in the goalkeeper_clear_kick_subtree

crates/world_state/src/behavior/goalkeeper.rs
```diff
@@ -0,0 +1,141 @@
...
+        sequence!(
+            negation!(condition!(hulks_is_kicking_team)),
+            selection!(
+                condition!(is_sub_state, SubState::ThrowIn),
+                condition!(is_sub_state, SubState::IndirectFreeKick),
+                condition!(is_sub_state, SubState::DirectFreeKick),
+            ),
+            condition!(is_ball_near_own_goal),
+            subtree!(goalkeeper_active_defense_position_subtree)
+        ),
```
pejotejo: you can combine all of the conditions by putting the condition sequences into a select macro which is in sequence with the subtree:
```
sequence!(
      select!(
            sequence!(conditions1....),
            sequence!(conditions2....)
       ),
       subtree!(...)
```

crates/world_state/src/behavior/goalkeeper.rs
```diff
@@ -0,0 +1,141 @@
...
+                subtree!(goalkeeper_intercept_kick_subtree)
+            ),
+            sequence!(
+                condition!(is_goalkeeper_visual_kick_hold_active),
+                subtree!(goalkeeper_clear_kick_subtree)
+            ),
+            sequence!(
+                condition!(is_goalkeeper_clear_candidate),
+                subtree!(goalkeeper_clear_kick_subtree)
+            ),
```
pejotejo: you can combine all of the conditions by putting the conditions into a select macro which is in sequence with the subtree:
```
sequence!(
      select!(
            condition!(....),
            condition!(....)
       ),
       subtree!(...)
```

crates/world_state/src/behavior/goalkeeper.rs
```diff
@@ -0,0 +1,141 @@
...
+            sequence!(
+                condition!(is_ball_suspiciously_behind_own_goal),
+                subtree!(goalkeeper_clear_or_walk_to_ball_subtree)
+            ),
```
pejotejo: what case do you want to cover with this

crates/world_state/src/behavior/goalkeeper.rs
```diff
@@ -0,0 +1,141 @@
...
+    selection!(
+        switch_motion_type(
+            MotionType::Walk,
+            sequence!(
+                action!(set_goalkeeper_active_defense_position),
+                action!(walk_to_block_position)
+            ),
+            Node::Failure,
+        ),
+        subtree!(goalkeeper_default_position_subtree),
```
pejotejo: this is unnecessary since if this subtree is not triggered it defaults to the same

crates/world_state/src/behavior/condition.rs
```diff
@@ -21,20 +24,142 @@ pub fn is_ball_interception_candidate(blackboard: &mut Blackboard) -> bool {
...
+        let ball_in_ground = ground_to_field.inverse() * ball.position;
+
+        ball.velocity.norm() < parameters.keeper_clear_ball_maximum_velocity
+            && ball_in_ground.coords().norm() < parameters.keeper_clear_ball_maximum_robot_distance
     } else {
         false
     }
 }
 
+pub fn is_goalkeeper_visual_kick_hold_active(blackboard: &mut Blackboard) -> bool {
```
pejotejo: why do we need this separately, cant you use for the motionswitch function the  switch_motion_type function?  Wouldnt this also mean that we always stay in kick mode when the ball is in the penalty area. You know that we ignore all obstacles in there.

pejotejo: the name is confusing

crates/types/src/parameters.rs
paudar: Done

crates/world_state/src/behavior/condition.rs
paudar: Done

crates/world_state/src/behavior/condition.rs
paudar: Done

crates/world_state/src/behavior/kick.rs
paudar: Done

crates/world_state/src/behavior/kick.rs
paudar: I had problems with a policy switching, if the ball got lost on the side of the image.
The next condition returns false, if the ball is not visible (for now. i think about how to change it).
For now i remove it.

crates/world_state/src/behavior/goalkeeper.rs
paudar: now irrelevant

crates/world_state/src/behavior/goalkeeper.rs
paudar: I had the problem with de localisation. The robot was directly outside the goal area and would not go the the ball, as it was not in the penalty area (for its point of view)

crates/world_state/src/behavior/goalkeeper.rs
paudar: is removed

crates/world_state/src/behavior/condition.rs
paudar: Done

crates/world_state/src/behavior/condition.rs
paudar: Done

crates/world_state/src/behavior/goalkeeper.rs
paudar: Done

crates/world_state/src/behavior/kick.rs
paudar: Done

crates/world_state/src/behavior/condition.rs
paudar: now irrelevant

pejotejo (approved):
logotomie

PR #2558, opened 2026-05-30
port ball filter

pejotejo:
## Why? What?
Ports the ballfilter to the new framework. Also moved the filter content to the filter.rs. Now the real ballfilter exists in the filter.rs.
For the node to work there was also a change in the future queue necessary where I'm not sure if this is correct for a right implementation of a future queue.  

## ToDo / Known Issues

Probably the real future queue problems, but I'm already over the limit of my skills and my prompt engineering.

## How to Test

Upload to the robot and look at the published data in twix.


oleflb:
Not sure whether the ball filter really belongs into the filter crate. In my understanding that crate has low level filters that can be reused in several contexts. The ball filter is more high level and can only be used for one thing. So maybe this should actually go to the ball filter node crate or keep in a distinct crate.

crates/bevyhavior_simulator/src/robot.rs
```diff
@@ -28,6 +28,7 @@ use linear_algebra::{
 };
 use parameters::directory::deserialize;
 use projection::intrinsic::Intrinsic;
+use ros_z::time::Time as RosTime;
```
oleflb: Why RosTime?

crates/bevyhavior_simulator/src/robot.rs
```diff
@@ -380,21 +382,21 @@ pub fn cycle_robots(
                 ball.state.as_ref().map(|ball| BallPosition {
                     position: robot.ground_to_field().inverse() * ball.position,
                     velocity: robot.ground_to_field().inverse() * ball.velocity,
-                    last_seen: now,
+                    last_seen: now_ros,
```
oleflb, line 384: This should be the timestamp of the last image where this ball was seen

crates/nodes/ball_filter/src/lib.rs
```diff
@@ -1,208 +1,524 @@
...
+        .await?;
 
-    pub fn decay_hypotheses(&mut self, decay_factor_criterion: impl Fn(&BallHypothesis) -> f32) {
-        for hypothesis in self.hypotheses.iter_mut() {
-            let decay_factor = decay_factor_criterion(hypothesis);
-            hypothesis.validity *= decay_factor;
-        }
-    }
+    let mut ball_filter = BallFilter::default();
+    let mut last_odometer = None;
```
oleflb, line 94: Can't you just add current_to_last_odometry as output to the odometer node?

oleflb (dismissed):
Short sbahn review, not complete

pejotejo:
> Not sure whether the ball filter really belongs into the filter crate. In my understanding that crate has low level filters that can be reused in several contexts. The ball filter is more high level and can only be used for one thing. So maybe this should actually go to the ball filter node crate or keep in a distinct crate.

i'm a bit confused, i did not move anything in any other crate, i just put the ball filter itself into a separate file into. So im not sure with what you are not happy with.

crates/bevyhavior_simulator/src/robot.rs
pejotejo: there allready exists a time in this crate, i just wrote the import in the line now

crates/bevyhavior_simulator/src/robot.rs
pejotejo: these are just some type fixes for the ballposition type since rosz does  not support SystemTime, i did not change anything in terms of contents

crates/nodes/ball_filter/src/lib.rs
pejotejo: but cant it happen that one published output is not read and skipped, or some timing is wrong. Besides that should this be part of the port?

crates/nodes/ball_filter/src/lib.rs
```diff
@@ -1,213 +1,528 @@
...
+        NotNan::new(-mahalanobis_distance).expect("mahalanobis distance is NaN")
+    })
+}
+
+fn project_detected_balls(
+    detections: Option<&[Object<RobocupObjectLabel>]>,
+    camera_matrix: Option<&CameraMatrix>,
+    parameters: &BallFilterParameters,
+    ball_radius: f32,
+) -> Vec<BallPercept> {
```
paudar: I think the function should return an `Option<Vec<BallPercept>>` so that the case when a valid frame has no ball detections and when detections/camera matrix are missing are handled differently (missing data should not be negative evidence, but an empty frame should)

crates/nodes/ball_filter/src/lib.rs
```diff
@@ -1,213 +1,528 @@
...
-        for hypothesis in self.hypotheses.iter_mut() {
-            hypothesis.predict(
-                delta_time,
-                last_to_current_odometry,
-                velocity_decay,
-                moving_process_noise,
-                resting_process_noise,
-                log_likelihood_of_zero_velocity_threshold,
-            )
+        let future_map_item = future_map.recv().await?;
```
paudar: If `field_dimensions` is not available (like at startup), the loop immediately continues before reaching any await. Maybe get the `field_dimensions` after the await of the future_map_item assignment.

crates/nodes/ball_filter/src/lib.rs
```diff
@@ -1,213 +1,528 @@
...
+    let filtered_balls_in_image_pub = node
+        .publisher::<Vec<Circle<Pixel>>>("ball_filter/filtered_balls_in_image")?
+        .build()
+        .await?;
+    let ball_percepts_pub = node
+        .publisher::<Vec<BallPercept>>("ball_filter/ball_percepts")?
+        .build()
+        .await?;
+    let ball_position_pub = node
+        .publisher::<Option<BallPosition<Ground>>>("ball_filter/ball_position")?
```
paudar, line 83: Could you update the subscriber of the `search_suggestor` and `ball_state_composer` to use the same topic name and type.

crates/nodes/ball_filter/src/lib.rs
```diff
@@ -1,213 +1,528 @@
...
+        let covariance = hypothesis.position_covariance();
+
+        let mahalanobis_distance = residual.dot(
+            &covariance
+                .cholesky()
+                .expect("covariance not invertible")
+                .solve(&residual),
+        );
+
+        NotNan::new(-mahalanobis_distance).expect("mahalanobis distance is NaN")
```
paudar, line 363: I would like to see a bit more error handling to avoid a panic at runtime (e.g. from bad parameters, degenerate projection, or numerical instability).

crates/nodes/image_receiver/src/lib.rs
```diff
@@ -25,7 +25,7 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
         .build()
         .await?;
     let right_image_pub = node
-        .publisher::<Image>("inputs/right_image")?
+        .publisher::<TimeWrapper<Image>>("inputs/right_image")?
```
paudar, line 28: Why do you change the image publisher in this PR?

crates/bevyhavior_simulator/src/robot.rs
```diff
@@ -359,6 +359,7 @@ pub fn cycle_robots(
 ) {
     let messages_sent_last_cycle = take(&mut messages.messages);
     let now = SystemTime::UNIX_EPOCH + time.elapsed();
+    let now_ros = rosz::Time::from_wallclock(now);
```
paudar: did you mean `ros_z::time::Time::from_wallclock(now)` ?

crates/path_serde/Cargo.toml
```diff
@@ -11,6 +11,7 @@ nalgebra = { workspace = true }
 ndarray = { workspace = true }
 num-traits = { workspace = true }
 path_serde_derive = { workspace = true }
+ros-z = { workspace = true }
```
paudar: I am not so happy with this. `path_serde` is a low-level serialization crate, but this PR adds a dependency on ros-z only to mark `ros_z::time::Time` as unsupported.

crates/nodes/ball_filter/src/lib.rs
pejotejo: field dimensions is a parameter provided by the global parameter provider. We always have it, else something would be very broken.

crates/nodes/image_receiver/src/lib.rs
pejotejo: because it triggered me that one image has it and the other one does not and i was already doing the other one. And if we ever use it, it should have this wrapper. The other change got lost in the rebase.

crates/bevyhavior_simulator/src/robot.rs
pejotejo: done

crates/nodes/ball_filter/src/lib.rs
pejotejo: done

crates/path_serde/Cargo.toml
pejotejo: done, but now last_seen skips path_serde

crates/nodes/ball_filter/src/lib.rs
pejotejo: i did not change the logic of the ball filter, imo this should be a separate pr

crates/nodes/ball_filter/src/lib.rs
pejotejo: done

crates/bevyhavior_simulator/src/robot.rs
paudar: Since the bevyhavior_simulator is a bigger task, it is fine here ...

crates/nodes/ball_filter/src/lib.rs
paudar: to be complete: please also change the type

crates/nodes/ball_filter/src/lib.rs
paudar: Ok, but if it panicks i know why ;)

crates/nodes/image_receiver/src/lib.rs
paudar: Ok. Fine.

crates/nodes/ball_filter/src/lib.rs
paudar: I agree, that this might be problematic with the new async environment

crates/nodes/ball_filter/src/lib.rs
paudar: was done

paudar (approved):
LGTM

PR #2559, opened 2026-05-30
Validate non-crossed yolo models

alexschmander:
## Why? What?

Add `validate-original` support to validate non-crossed yolo models.


knoellle (approved):
LGTM

PR #2560, opened 2026-05-31
Wrap node async in box

knoellle:
## Why? What?

Speeds up hot rebuild for hulk_ros_z massively.
On main changing a node and then hot rebuilding takes ~28 seconds on our remote compiler.
With this PR, this is reduced to ~3 seconds.

- Fixes #pain

## How to Test

```bash
pepsi build crates/hulk_ros_z/Cargo.toml
# change a file
pepsi build crates/hulk_ros_z/Cargo.toml
```


alexschmander (approved):
LGTM

PR #2561, opened 2026-05-31
Activate mkdocs ci job

alexschmander:
## Why? What?

What the title says. We had failed doc publishing docs, because our PR ci does not check docs anymore. This PR enables the CI again.


oleflb (commented):
Wasn't the failing CI caused by rustdoc?

PR #2563, opened 2026-05-31
feat(ros-z): warn on subscriber queue loss

schmidma:
## Summary
- Emit a structured warning when a bounded subscriber queue drops the oldest sample.
- Include subscriber kind, topic, node namespace/name, message type, queue capacity, drop policy, and total dropped samples in the warning fields.
- Keep queue loss as logging-only UX for now instead of exposing a public stats API.

## Test Plan
- [x] `cargo fmt --check`
- [x] `cargo nextest run -p ros-z queue`
- [x] `cargo nextest run -p ros-z pubsub`
- [x] `cargo clippy -p ros-z --all-targets --locked -- -D warnings`
- [x] `cargo test --doc -p ros-z`


paudar (approved):
LGTM

PR #2564, opened 2026-05-31
keep unlimited history in future map subscribers

rmburg:
## Why? What?

Fixes future maps breaking with slow subscribers and/or large differences in publisher frequency. Not the prettiest fix, a proper one needs some more thought to work.

- work without unbounded buffers

## How to Test

Not sure where this bug occurred before, I have a test case which is fixed by these changes.

### DIFF
--- a/crates/ros-z-streams/src/future_queue.rs
+++ b/crates/ros-z-streams/src/future_queue.rs
@@ -7,6 +7,7 @@ use ros_z::{
     Message, Result,
     node::Node,
     pubsub::{Received, Subscriber},
+    qos::{QosHistory, QosProfile},
     time::Time,
 };
 use tokio::select;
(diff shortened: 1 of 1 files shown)


oleflb (commented):
Please open an issue for improving this in the future.

Can you please confirm nothing is needed on the sender side?

rmburg:
> Please open an issue for improving this in the future.

Now tracked by #2581.

> Can you please confirm nothing is needed on the sender side?

This PR fixes my artificial reproduction of the bug without any changes to the producer.
@pejotejo wanted to test this on his PR to confirm that this also fixes it there.

pejotejo:
i tested it with my pr, and it worked

oleflb (approved):

oleflb:
@rmburg merge when ready

PR #2568, opened 2026-06-03
add rosz doctor pubsub diagnostics

schmidma:
## Summary
- add `rosz doctor [--settle-timeout SECONDS]` as a bounded whole-graph pub/sub snapshot linter
- support text output and the existing global `--json` output mode
- report dangling publishers/subscribers as warnings
- report type/schema mismatches and QoS incompatibilities as errors
- return exit code `1` only when error-level findings exist
- use graph revision changes for CLI-owned graph settling before collecting diagnostics

## Review response
- keep the intentional `Graph` to `GraphView` diagnostics API break and document the new public graph APIs
- build doctor reports from a cloned endpoint snapshot so report analysis and sorting happen outside the graph lock
- compute doctor type and QoS findings from grouped endpoint data in `ros-z-cli`
- serialize QoS compatibility as stable `snake_case` JSON instead of Rust debug text
- keep shared graph-change quiet-window settling for existing CLI commands

## How to Test
- `cargo fmt --check`
- `cargo nextest run -p ros-z-cli`
- `cargo test --doc -p ros-z`

### DIFF
--- a/crates/ros-z-cli/src/cli.rs
+++ b/crates/ros-z-cli/src/cli.rs
@@ -24,7 +24,6 @@ fn parse_positive_duration(value: &str) -> Result<Duration, String> {
     }
     Ok(duration)
 }
-
 /// Graph entity kind accepted by `rosz list`.
 #[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
 pub enum ListTarget {
@@ -122,6 +121,12 @@ pub enum OnlineCommand {
     Watch,
     /// Show the full graph snapshot
     Graph,
+    /// Diagnose pub/sub graph issues and exit non-zero on errors
+    Doctor {
+        /// Maximum time to wait for graph changes to settle
+        #[arg(long, default_value = "2s", value_parser = humantime::parse_duration)]
+        settle_timeout: Duration,
+    },
     /// Dynamically inspect a topic's messages
     Echo {
         topic: String,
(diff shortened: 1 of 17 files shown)


crates/ros-z-cli/src/cli.rs
```diff
@@ -1,6 +1,29 @@
 use clap::{Parser, Subcommand, ValueEnum};
 use clap_complete::Shell;
 
+const DEFAULT_SETTLE_TIMEOUT_SECONDS: f64 = 2.0;
+const MAX_SETTLE_TIMEOUT_SECONDS: f64 = u32::MAX as f64;
+
+fn parse_settle_timeout_seconds(value: &str) -> Result<f64, String> {
```
paudar: Why is this a float and not directly a Duration? (i see that it is converted later)

crates/ros-z/src/graph/query.rs
```diff
@@ -192,6 +204,79 @@ impl GraphView<'_> {
...
+                        topic: topic.to_string(),
+                        publisher: publisher.clone(),
+                        subscription: subscription.clone(),
+                    });
+                }
+            }
+        }
+
+        diagnostics
+    }
```
paudar, line 288: Isn't that duplicated code to src/graph/query.rs l.259?

paudar (commented):
Looks good so far. I have two small comments:

crates/ros-z-cli/src/cli.rs
schmidma: Changed this to parse directly into `Duration` using `humantime::parse_duration`, so the CLI now accepts explicit duration syntax like `2s` or `500ms` and avoids the late float-to-duration conversion.

crates/ros-z/src/graph/query.rs
schmidma: Added a private `GraphView` helper for pub/sub pair traversal and reused it for the diagnostic queries. The CLI still analyzes cloned endpoint snapshots outside the graph lock.

paudar (approved):
LGTM

PR #2569, opened 2026-06-03
warn on recv without visible publishers

schmidma:
## Summary
- add a default 5s receive warning timeout for typed, dynamic, raw, and cache subscribers
- add builder controls to configure or disable the warning, including cache builders for both default and custom stamp modes
- keep recv semantics unchanged: after the timeout, check visible publishers through `GraphView`, warn only if none are visible, then keep waiting
- document the warning behavior and public controls

## Testing
- keep behavioral coverage focused on typed/raw `recv()` continuing after the warning timeout path
- omit structural/private-field builder tests for warning option plumbing

## Test Plan
- [x] `cargo check -p ros-z`
- [x] `cargo nextest run -p ros-z recv_warning_timeout_still_waits_for_eventual_typed_message raw_recv_warning_timeout_still_waits_for_eventual_sample --no-fail-fast`
- [x] `cargo test -p ros-z --doc`
- [x] `cargo nextest run -p ros-z --no-fail-fast`

### DIFF
--- a/crates/ros-z/src/pubsub/raw.rs
+++ b/crates/ros-z/src/pubsub/raw.rs
@@ -5,8 +5,12 @@ use std::time::Duration;
 use zenoh::sample::Sample;
 
 use crate::Result;
+use crate::entity::EndpointEntity;
+use crate::graph::Graph;
 use crate::message::WireDecoder;
-use crate::pubsub::subscriber::{SubscriberBuilder, SubscriberResources};
+use crate::pubsub::subscriber::{
+    SubscriberBuilder, SubscriberResources, recv_sample_with_publisher_warning,
+};
 use crate::qos::QosProfile;
 use crate::queue::BoundedQueue;
 
@@ -37,13 +41,25 @@ impl WireDecoder for RawPayloadCodec {
 /// Received samples are delivered as [`Sample`] values without deserialization.
 pub struct RawSubscriber {
     queue: Arc<BoundedQueue<Sample>>,
+    graph: Arc<Graph>,
+    entity: EndpointEntity,
+    publisher_warning_timeout: Option<Duration>,
     _resources: SubscriberResources,
 }
 
 impl RawSubscriber {
-    pub(super) fn new(queue: Arc<BoundedQueue<Sample>>, resources: SubscriberResources) -> Self {
+    pub(super) fn new(
+        queue: Arc<BoundedQueue<Sample>>,
+        resources: SubscriberResources,
+        graph: Arc<Graph>,
+        entity: EndpointEntity,
+        publisher_warning_timeout: Option<Duration>,
+    ) -> Self {
         Self {
             queue,
+            graph,
+            entity,
+            publisher_warning_timeout,
             _resources: resources,
         }
     }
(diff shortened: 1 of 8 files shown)


crates/ros-z/src/pubsub/raw.rs
```diff
@@ -32,8 +48,20 @@ impl RawSubscriber {
     /// Zenoh and does not deserialize it into a message type. The receive is
     /// cancel-safe: cancelling this future before it completes does not remove a
     /// sample from the queue.
+    ///
+    /// By default, this logs a warning after
+    /// [`DEFAULT_PUBLISHER_WARNING_TIMEOUT`](crate::pubsub::DEFAULT_PUBLISHER_WARNING_TIMEOUT) if
+    /// no sample arrives and no publishers are visible for the topic. The warning does not end the
+    /// receive; this method continues waiting for the next sample.
     pub async fn recv(&mut self) -> Result<Sample> {
```
oleflb, line 78: Why does this function return a result although it never fails?

crates/ros-z/src/pubsub/subscriber.rs
```diff
@@ -334,11 +362,41 @@ where
...
+    queue: &BoundedQueue<Sample>,
+    graph: &Graph,
+    entity: &EndpointEntity,
+    publisher_warning_timeout: Option<Duration>,
+) -> Sample {
+    let Some(timeout) = publisher_warning_timeout else {
+        return queue.recv_async().await;
+    };
+
+    tokio::select! {
```
oleflb: I think join is the better semantic fit here

crates/ros-z/tests/pubsub.rs
```diff
@@ -155,6 +155,51 @@ async fn raw_subscriber_receives_sample_payload() -> zenoh::Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn raw_recv_warning_timeout_still_waits_for_eventual_sample() -> ros_z::Result<()> {
```
oleflb, line 217: This test never tests that the warning is actually emitted. Therefore name is misleading. Maybe you can setup a simple logger that logs into a `String` and you do a `contains` check?

oleflb (changes requested):

crates/ros-z/tests/pubsub.rs
schmidma: Added a log-capture assertion. The test now verifies that the warning is emitted for the topic and still checks that `recv()` continues waiting and delivers a later message after the timeout path.

crates/ros-z/src/pubsub/subscriber.rs
schmidma: I don't think that `join!` is correct here. It waits for both branches to complete. If a sample arrives immediately, `join!(queue.recv_async(), sleep(timeout))` would still wait until the timeout before returning the sample. That is not the intended behavior.

I changed this to `tokio::time::timeout(timeout, queue.recv_async())`, which matches the intended behavior: return immediately when a sample arrives, otherwise warn after the timeout and then continue waiting. ... and is much more concise

crates/ros-z/src/pubsub/raw.rs
schmidma: This signature predates this PR and is kept intentionally here. Even though raw receive currently cannot fail, it stays aligned with the typed and dynamic `recv()` APIs, and changing it would be an unrelated public API cleanup.

oleflb (dismissed):

oleflb (approved):

PR #2570, opened 2026-06-03
warn on endpoint type mismatch

schmidma:
## Summary
- add a generalized graph helper for endpoint type metadata mismatches
- compare all visible endpoints on the same graph name, including publishers, subscribers, services, and clients
- warn when a publisher, subscriber, service server, or service client starts and any visible endpoint on that name advertises a different type name or schema hash
- keep the warnings non-fatal

## How to Test
- Start a ros-z publisher, subscriber, service server, or service client for a graph name with one advertised type.
- Start another visible endpoint on the same graph name with a different advertised type.
- Confirm the endpoint still starts, but logs a warning that includes the local endpoint metadata and mismatching endpoint metadata.

## Verification
- `cargo fmt --check`
- `cargo test -p ros-z --doc`

### DIFF
--- a/crates/ros-z/src/dynamic/schema_service.rs
+++ b/crates/ros-z/src/dynamic/schema_service.rs
@@ -14,6 +14,7 @@ use crate::ServiceTypeInfo;
 use crate::attachment::Attachment;
 use crate::context::GlobalCounter;
 use crate::entity::{EndpointEntity, EndpointKind, NodeEntity, SchemaHash, TypeInfo};
+use crate::graph::Graph;
 use crate::message::{SerdeCdrCodec, Service, WireDecoder, WireEncoder};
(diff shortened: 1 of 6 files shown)


paudar (commented):
Code looks good. 
Works on the Robot and detects mismatching types of Node-Skeletons, that are not implemented yet. 
By default, the log level is set to `Error`. So a `pepsi logs show` does not show the warnings. I think it should be easier to see this conflicts, but are not sure, if i prefer to always  see warnings (change the log level of our binary) or mark the mismatching types as `Error`

schmidma:
I agree, that it should be easier to see these logs via pepsi logs show. But as I understand your request, this is not something this PR should solve. Or what should be done here?

paudar:
You are right, this should be part of a new PR.

paudar (approved):

PR #2572, opened 2026-06-03
refactor(ros-z): move entity helpers to protocol

schmidma:
## Summary
- Move protocol-owned entity helpers onto `ros-z-protocol` entity types.
- Replace `ros-z` graph/node/pubsub/service free-function call sites with entity methods.
- Keep `EndpointGlobalId` in `ros-z` for a separate follow-up cleanup.

## How to Test
- `cargo fmt --check`
- `cargo check -p ros-z-protocol`
- `cargo check -p ros-z`
- `cargo nextest run -p ros-z-protocol key_expr`
- `cargo nextest run -p ros-z-protocol entity::tests`
- `cargo nextest run -p ros-z graph`

### DIFF
--- a/crates/ros-z-protocol/src/entity.rs
+++ b/crates/ros-z-protocol/src/entity.rs
@@ -50,6 +50,21 @@ impl Display for TopicKE {
     }
 }
 
+/// Namespace/name key used to index graph entities by node.
+pub type NodeKey = (String, String);
+
+/// Normalize a node namespace for internal graph indexing.
+///
+/// The root namespace (`"/"`) is stored as an empty string so local and remote
+/// entities use the same key representation.
+pub fn normalize_node_namespace(namespace: &str) -> String {
+    if namespace == "/" {
+        String::new()
+    } else {
+        namespace.to_owned()
+    }
+}
+
 /// ros-z node entity.
 #[derive(Debug, Hash, Clone, PartialEq, Eq)]
 pub struct NodeEntity {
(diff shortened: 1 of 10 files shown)


crates/ros-z-protocol/src/entity.rs
```diff
@@ -281,4 +363,71 @@ mod tests {
...
+    fn entity_kind_returns_endpoint_kind_for_endpoint_entity() {
+        assert_eq!(
+            Entity::Endpoint(endpoint(EndpointKind::Publisher)).kind(),
+            EntityKind::Publisher
+        );
+        assert_eq!(
+            Entity::Endpoint(endpoint(EndpointKind::Subscription)).kind(),
+            EntityKind::Subscription
+        );
+    }
```
paudar, line 404: This test could be extended to test all four endpoint kinds

crates/ros-z/tests/graph.rs
```diff
@@ -202,7 +202,7 @@ mod tests {
         let session = zenoh::open(zenoh::Config::default()).await?;
         let graph = ros_z::graph::Graph::new(&session).await?;
         let node = NodeEntity::new(session.zid(), 1, "removed_node".to_string(), String::new());
```
paudar: Instead of `String::new()` a `"/".to_string()` could be used to check also for the functionality of key dropping the `/`.

paudar (commented):
I have two minor comments regarding the tests; otherwise, the code looks fine. 
It still needs to be tested on the robot.

crates/ros-z-protocol/src/entity.rs
schmidma: Covered all four endpoint kinds in `entity_kind_returns_endpoint_kind_for_endpoint_entity` in bca45a0fe.

crates/ros-z/tests/graph.rs
schmidma: Changed the removed-node graph test to use the root namespace (`/`), so `NodeEntity::key()` normalization is covered in bca45a0fe.

paudar (commented):
Looks good, but has conflicts

schmidma:
conflicts are resolved

paudar (approved):
LGTM

PR #2582, opened 2026-06-04
Set `Restart=no`

oleflb:
## Why? What?

Someone :eyes: configures `hulk` to always restart when stopped. 
I disabled this so we have the same behavior as with the NAOs before: No restart after a crash.

- Fixes #

## ToDo / Known Issues

## How to Test


tools/pepsi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pepsi"
-version = "7.26.0"
+version = "7.27.0"
```
knoellle: Doesn't hurt, but why did you bump the pepsi version?

oleflb: removed

knoellle (approved):
LGTM

PR #2584, opened 2026-06-06
fix(ros-z): use Zenoh-native graph name validation

schmidma:
## Summary
- Relax ros-z graph-name validation to accept Zenoh-native concrete components, including digit-leading and hyphenated names.
- Validate node identities before liveliness formatting so reserved characters like `%` cannot corrupt discovery round-trips.
- Preserve `hulk_ros_z --robot` values except for adding a leading slash, so invalid names are rejected by ros-z instead of rewritten.
- Simplify graph-name validation helpers and keep behavior-focused coverage/docs for the new rules.

### DIFF
--- a/crates/ros-z/src/topic_name.rs
+++ b/crates/ros-z/src/topic_name.rs
@@ -22,24 +22,53 @@ pub enum TopicNameError {
     InvalidNodeName(String),
 }
 
-/// Validate that a topic name component (between slashes) is valid
-/// Components must start with a letter or underscore, followed by alphanumeric or underscores
-fn is_valid_topic_component(component: &str) -> bool {
-    if component.is_empty() {
-        return false;
-    }
-    let bytes = component.as_bytes();
-    if !bytes[0].is_ascii_alphabetic() && bytes[0] != b'_' {
-        return false;
-    }
-    bytes[1..]
-        .iter()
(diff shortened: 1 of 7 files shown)


crates/ros-z/src/topic_name.rs
```diff
@@ -334,23 +386,28 @@ mod tests {
...
-        assert!(is_valid_topic_component("_foo"));
-        assert!(is_valid_topic_component("foo123"));
-        assert!(is_valid_topic_component("foo_bar"));
-        assert!(is_valid_topic_component("FooBar"));
-
-        assert!(!is_valid_topic_component(""));
-        assert!(!is_valid_topic_component("123"));
-        assert!(!is_valid_topic_component("foo-bar"));
-        assert!(!is_valid_topic_component("foo bar"));
+    fn test_valid_graph_components() {
```
knoellle: From this test name I'd expect the strings being tested to be valid, not invalid.

knoellle (commented):
Minor naming issue, looks good otherwise

crates/ros-z/src/topic_name.rs
schmidma: I restructured and cleaned many of the tests, as I had the feeling that there was much duplication. I think the test you commented did not survive...

knoellle (approved):
LGTM

PR #2585, opened 2026-06-06
refactor(ros-z): share dynamic JSON rendering

schmidma:
## Summary
- Move dynamic JSON rendering into reusable `ros-z` APIs with explicit byte and non-finite float policies.
- Keep `ros-z-cli echo --json` output compatible by selecting the legacy full-byte-array/null-float policy.
- Cover shared renderer behavior, including nalgebra dynamic JSON and non-finite floats.


paudar (approved):
LGTM and echo still works on the robot 👍

PR #2588, opened 2026-06-06
Remove game controller TimeWrappers

schmidma:
## Summary
- Publish raw and filtered game-controller state as plain optional values.
- Align game-controller consumers with the optional state schema.
- Keep receive-time wrappers on network messages and player states.

## Test Plan
- [x] `cargo nextest run`
- [x] `cargo test --doc`

### DIFF
--- a/crates/nodes/game_controller_state_filter/src/lib.rs
+++ b/crates/nodes/game_controller_state_filter/src/lib.rs
@@ -12,7 +12,7 @@ use types::{
     field_dimensions::FieldDimensions, filtered_game_controller_state::FilteredGameControllerState,
     filtered_game_state::FilteredGameState, filtered_whistle::FilteredWhistle,
     game_controller_state::GameControllerState, parameters::GameStateFilterParameters,
-    players::Players, time_wrapper::TimeWrapper, world_state::BallState,
+    players::Players, world_state::BallState,
 };
 
 use crate::state::State;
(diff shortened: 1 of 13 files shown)


crates/nodes/game_controller_state_filter/src/lib.rs
```diff
@@ -66,7 +66,7 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
         .build()
         .await?;
     let filtered_game_controller_state_pub = node
-        .publisher::<TimeWrapper<FilteredGameControllerState>>("filtered_game_controller_state")?
+        .publisher::<Option<FilteredGameControllerState>>("filtered_game_controller_state")?
```
alexschmander: Why did you change the `FilteredGameControllerState topic` type to an `Option<>`? This only ever publishes a `Some<>`, doesn't it?

alexschmander (changes requested):

crates/nodes/game_controller_state_filter/src/lib.rs
schmidma: It was before, in the old framework. You are correct, that in the moment, there is no way back. For me this was a semantic decision, that we can have no GameController, after we discovered one. How do you feel about that? Maybe this is future work, and we should reduce to non-optional state?

alexschmander: I think we should reduce to the non-optional state and add handling for missing game controllers as future work.

alexschmander (approved):
LGTM

PR #2590, opened 2026-06-06
Port search suggestor

pejotejo:
## Why? What?
what the title says

depends on #2558 

## How to Test
upload to the robot and look at the published outputs in twix

### DIFF
--- a/crates/nodes/search_suggestor/src/lib.rs
+++ b/crates/nodes/search_suggestor/src/lib.rs
@@ -1,4 +1,7 @@
-use color_eyre::{Result, eyre::WrapErr as _};
+use color_eyre::{
+    Result,
+    eyre::{WrapErr as _, ensure},
+};
 use coordinate_systems::{Field, Ground};
 use linear_algebra::{Isometry2, Point2, Vector2, point, vector};
 use ndarray::{Array2, array};
@@ -61,7 +64,7 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
         .subscriber::<TimeWrapper<IncomingMessage>>("filtered_message")?
         .build()
         .await?;
-    let heatmap_pub = node
+    let additional_heatmap_pub = node
         .publisher::<types::heatmap::Heatmap>("ball_search_heatmap")?
         .build()
         .await?;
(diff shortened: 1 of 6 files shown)


pejotejo:
@alexschmander there is now a fake loca in main, it therefore can be tested now

crates/nodes/search_suggestor/src/lib.rs
```diff
@@ -22,55 +24,180 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
+            let suggested_search_position = point![
+                ((x as f32 + 1.0 / 2.0) / heatmap.cells_per_meter - field_dimensions.length / 2.0),
+                ((y as f32 + 1.0 / 2.0) / heatmap.cells_per_meter - field_dimensions.width / 2.0)
+            ];
+            suggested_search_position_pub
+                .publish(&suggested_search_position)
+                .await?;
+        }
+
+        heatmap_pub.publish(&heatmap.to_message()).await?;
```
alexschmander: Since this is basically just an additional output, you can guard this with a has_subscriber(), then this rather large heatmap is only published, when there actually are subscribers, such as twix.
```suggestion
        if heatmap_pub.has_subscribers() {
            heatmap_pub.publish(&heatmap.to_message()).await?;
        }
```

crates/nodes/search_suggestor/src/lib.rs
```diff
@@ -22,55 +24,180 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
         .await?;
 
-    pending::<()>().await;
+    let field_dimensions = field_dimensions_sub.recv().await?;
+    let initial_parameters_snapshot = parameters.snapshot();
+    let initial_parameters = initial_parameters_snapshot.typed();
+    let (heatmap_length, heatmap_width) = (
+        (field_dimensions.length * initial_parameters.cells_per_meter).round() as usize,
+        (field_dimensions.width * initial_parameters.cells_per_meter).round() as usize,
+    );
```
alexschmander, line 82: cells per meter and heat_map_width are both denominators or factors in denominators causing a panic when configured to 0. 
width
```suggestion
    ensure!(heatmap_length > 0, "<useful error message including the wrong parameter value>");
    ensure!(heatmap_width > 0, "<useful error message including the wrong parameter value>"); 
```

alexschmander (changes requested):

crates/nodes/search_suggestor/src/lib.rs
pejotejo: i use publish_if_subscribed now

crates/nodes/search_suggestor/src/lib.rs
pejotejo: done

alexschmander (approved):
LGTM

PR #2591, opened 2026-06-07
port ball state composer

pejotejo:
## Why? What?
what the title says

depends on #2558 and #2550

## How to Test

Upload to the robot and look in twix if it porduces a ball state with the same ball position as the ball filter outputs or that are send per messages.

### DIFF
--- a/crates/nodes/ball_state_composer/src/lib.rs
+++ b/crates/nodes/ball_state_composer/src/lib.rs
@@ -1,16 +1,19 @@
 use std::{boxed::Box, future::Future, pin::Pin};
-use std::{future::pending, sync::Arc};
+use std::{sync::Arc, time::SystemTime};
 
 use color_eyre::Result;
 
 use coordinate_systems::{Field, Ground};
-use linear_algebra::Isometry2;
+use filtering::hysteresis::greater_than_with_hysteresis;
+use hsl_network_messages::{GamePhase, SubState, Team};
+use linear_algebra::{Isometry2, Point2, Vector2, point};
 use ros_z::{prelude::*, qos::QosDurability};
 use types::{
     ball_position::BallPosition,
-    field_dimensions::FieldDimensions,
+    field_dimensions::{FieldDimensions, Side},
     filtered_game_controller_state::FilteredGameControllerState,
     primary_state::PrimaryState,
+    time_wrapper::TimeWrapper,
     world_state::{BallState, LastBallState},
 };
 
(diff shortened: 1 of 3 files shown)


crates/nodes/ball_state_composer/src/lib.rs
```diff
@@ -21,49 +23,197 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
+        .with_qos(QosProfile {
             durability: QosDurability::TransientLocal,
             ..Default::default()
         })
         .build()
         .await?;
-    let _ball_position_sub = node
-        .subscriber::<Option<BallPosition<Ground>>>("ball_filter/ball_position")?
+    let ball_position_sub = node
+        .subscriber::<Option<BallPosition<Ground>>>("ball_position")?
```
alexschmander: This has the wrong topic name: 
```suggestion
        .subscriber::<Option<BallPosition<Ground>>>("ball_filter/ball_position")?
```

crates/nodes/ball_state_composer/src/lib.rs
```diff
@@ -21,49 +23,197 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
+                );
+
+                rule_ball_state_pub.publish(&rule_ball).await?;
+            }
+        }
+    }
+}
+
+fn compose_rule_ball_state(
+    primary_state: Option<PrimaryState>,
```
alexschmander: This primary state cannot be None anymore. Change its type as such:
```suggestion
    primary_state: PrimaryState,
```

alexschmander (changes requested):

crates/nodes/ball_state_composer/src/lib.rs
pejotejo: done

crates/nodes/ball_state_composer/src/lib.rs
pejotejo: done

pejotejo:
@alexschmander there is now a fake loca in main, it therefore can be tested now

crates/nodes/ball_state_composer/src/lib.rs
```diff
@@ -21,49 +23,197 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
+
+    loop {
+        let Some(field_dimensions) = field_dimensions_cache.get_latest() else {
+            continue;
+        };
+        let field_dimensions = *field_dimensions;
+        let Some(ground_to_field) = ground_to_field_cache.get_latest() else {
+            continue;
+        };
+        let ground_to_field = *ground_to_field;
```
alexschmander: Move both of these into the branches where they are actually used. This free spins, when they return None.

crates/nodes/ball_state_composer/src/lib.rs
```diff
@@ -21,49 +23,197 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
-    let _ground_to_field_sub = node
-        .subscriber::<Isometry2<Ground, Field>>("ground_to_field")?
+    let ground_to_field_cache = node
+        .create_cache::<Isometry2<Ground, Field>>("ground_to_field", 10)?
         .build()
         .await?;
-    let _team_ball_sub = node
-        .subscriber::<BallPosition<Field>>("team_ball")?
+    let team_ball_sub = node
+        .subscriber::<Option<BallPosition<Field>>>("team_ball")?
```
alexschmander: Why did you change this? The corresponding publisher is not yet ported and has a different type

alexschmander (changes requested):

crates/nodes/ball_state_composer/src/lib.rs
pejotejo: no reason, i did the other, so i did this instinctively, removed it again

crates/nodes/ball_state_composer/src/lib.rs
pejotejo: done

alexschmander (approved):
LGTM

PR #2597, opened 2026-06-07
fix: twix focus resize

pejotejo:
## Why? What?
my twix crashed probably to some resize reason and this option failed, now an if let is used to handle it properly


knoellle (dismissed):
LGTM

knoellle (approved):
LGTM

PR #2599, opened 2026-06-08
Twix Remote Control delay fix

pejotejo:
## Why? What?
The remote control is not very responsive and not very usable. This was because all changed parameters and input one got went over the same socket. So the large raw image files stalled the writing. Now there is one separate for writing parameters. 
With this the toggle buttons broke, since without the lag one saw the toggle change multiple times while the button was pressed. The kick mode is now just triggered on press. The safe mode now triggers on flanks.
This whole thing is needed for the video for Kai.

## How to Test
upload to the robot and try out the remote control. It should be more responsive now.

### DIFF
--- a/tools/twix/src/panels/remote_control.rs
+++ b/tools/twix/src/panels/remote_control.rs
@@ -40,8 +40,6 @@ impl<'a> Panel<'a> for RemotePanel {
         let gait_parameter_value = robot.subscribe_json("parameters.rl_walking.gait_frequency");
         let remote_stop_toggle: BufferHandle<bool> =
             robot.subscribe_value("parameters.remote_stop_toggle");
-        let kick_mode_toggle: BufferHandle<bool> =
-            robot.subscribe_value("parameters.behavior.remote_control.kick_mode_toggle");
         let bg_running = Arc::new(AtomicBool::new(true));
 
         let robot_clone = robot.clone();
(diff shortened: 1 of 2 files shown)


tools/twix/src/panels/remote_control.rs
paudar, line 255: Please handle the kick mode toggle in the `drop` function to avoid that true is the last value when closing the panel

tools/twix/src/panels/remote_control.rs
```diff
@@ -228,6 +225,13 @@ fn reset(robot: &Arc<Robot>) {
     update_step(robot, Step::<f32>::default(), 1.0);
 }
```
paudar, line 227: Please also deactivate the kick mode in the reset function.

tools/twix/src/panels/remote_control.rs
```diff
@@ -75,6 +75,11 @@ impl<'a> Panel<'a> for RemotePanel {
 
                 if gilrs.gamepads().next().is_none() {
                     let _ = sender.send((Step::default(), 1.0));
+
+                    set_kick_mode_toggle(&robot_clone, false);
```
paudar: Should this happen here every cycle even if twix was never connected to a  camecontroller or just if south was pressed?

paudar (commented):
Some small remarks regarding the ways to stop the kicking mode:

tools/twix/src/panels/remote_control.rs
pejotejo: done

tools/twix/src/panels/remote_control.rs
pejotejo: done

tools/twix/src/panels/remote_control.rs
pejotejo: done

paudar (approved):
LGTM

PR #2606, opened 2026-06-10
fake loca

pejotejo:
## Why? What?
to keep going, while loca is missing

## How to Test

Upload and look at the ground_to_field


knoellle (approved):
LGTM

PR #2607, opened 2026-06-11
Reduce cross-compile image size

oleflb:
## Why? What?

Reduces cross-compile image size by a factor of 26x.
Only tested that `pepsi build` works, did not try on robot. Maybe @paudar can do this?

- Fixes #

## ToDo / Known Issues

## How to Test

### DIFF
--- a/hulk.toml
+++ b/hulk.toml
@@ -1,2 +1,2 @@
 os_version = "v1.5.3.0-release-1174-2026-03-20"
-sdk_version = "1.1.0"
+sdk_version = "1.2.0"
--- a/services/aliveness/Cargo.lock
+++ b/services/aliveness/Cargo.lock
@@ -323,9 +323,8 @@ dependencies = [
 
 [[package]]
 name = "booster_sdk"
-version = "0.1.0-alpha.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62af599682f9b0df60ef2370f72e5a643baec826514966b534e047285f33050e"
+version = "0.1.1-alpha.1"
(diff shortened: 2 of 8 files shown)


alexschmander:
Why was booster_sdk added in the Cargo.toml of power panic and aliveness?

oleflb:
> Why was booster_sdk added in the Cargo.toml of power panic and aliveness?

Those packages are not in the workspace, but depend on booster_sdk through hula_types.
But they don't receive the workspace patch, causing cargo to emit warning
We could also fix this by including the packages in the workspace, since our reason for excluding them (yocto) is currently irrelevant.

alexschmander:
So those were just adjacent issues we also currently have on main?

alexschmander:
We should definitely bump the SDK version in hulk.toml

oleflb:
Yes, we don't check for these warning by default.
But they are unrelated to the image. This only became obvious since all packages with cross-compile = true have been tested

oleflb:
Yes, but test it first :D

alexschmander:
> Yes, but test it first :D

Will do today :D

alexschmander:
Works! Uploaded both with hulk_ros_z and hulk_booster. Both run without problems, and ort/tensorrt inference works as well. If you bump the sdk version, we can merge this. Very, very nice! Size is down from 60 GB down to 1.64GB and the image builds in less than a minute!

alexschmander (changes requested):
Please bump the sdk version in `hulk.toml`

oleflb:
ooops forgot to push

oleflb:
Huh, kind of weird github uses the commit time for the thread timeline, not the time of push

alexschmander (approved):
LGTM

PR #2611, opened 2026-06-12
Add read-only ros-z backend for Twix

schmidma:
## Summary
- Add ros-z-debug subscriptions and graph revision support needed by Twix.
- Replace Twix communication panels with read-only ros-z topic catalog, subscriptions, and completion.
- Document the new read-only ros-z backend and removed legacy Behavior Simulator panel path.

## Test Plan
- cargo fmt --check
- cargo nextest run -p ros-z --test graph
- cargo nextest run -p ros-z-debug
- cargo nextest run -p twix
- cargo check -p twix
- cargo clippy -p twix --all-targets --locked -- -D warnings
- git diff --check
- cargo test --doc -p ros-z -p ros-z-debug


schmidma:
Replaced by #2629, which uses schmidma:twix-ros-z-read-only as the head branch.

PR #2614, opened 2026-06-13
Being Prepared and Staying Safe

pejotejo:
## Why? What?
Adds the primary state Prepare and the motion command damping. Both existed combined in the old framework, since the robot mode handler handled it there, not it happens in behavior.

## How to Test
Upload to a robot and switch between the robot modes.


pejotejo:
we decided that we dont need this anymore

alexschmander (approved):
LGTM

PR #2620, opened 2026-06-13
Refactor and improve nearest cell in voronoi

JackyBloxx:
## Why? What?

Change the nearest cell function to use a Djikstra and set the seeding cost depending on how far the difference between real position and seed position is.
Ive done that for better seeding Positioning and allowing robots to have no Voronoi area when they are to deep into a obstacle.

## How to Test
Move the calculate voronoi action into playing tree so that it is calculateted constantly. Then let the robot go into an obstical for exampele by standing infront of it.
The Robot should still have an owned area or no if there is another robot that would be faster to that seeding position.

### DIFF
--- a/crates/voronoi/src/lib.rs
+++ b/crates/voronoi/src/lib.rs
@@ -66,6 +66,10 @@ pub enum Ownership {
 }
 
 impl Ownership {
+    fn is_blocked(self) -> bool {
+        matches!(self, Self::Blocked)
+    }
+
     fn is_free(self) -> bool {
         matches!(self, Self::Free)
     }
@@ -194,7 +198,13 @@ impl VoronoiGrid {
             let player_number = robots[robot_index].1;
             let (sin_h, cos_h) = robot_headings[robot_index];
 
-            for (neighbor_index, neighbor) in self.neighbor_indices(current_index) {
+            for (neighbor_index, neighbor) in
+                neighbor_indices(current_index, self.width_tiles, self.height_tiles)
+            {
+                if self.tiles[neighbor_index].is_blocked() {
+                    continue;
+                }
+
                 let rotation_cost = rotation_cost(sin_h, cos_h, neighbor, orientation_bias);
                 let new_cost = current_cost + neighbor.step_cost + rotation_cost;
 
@@ -366,6 +376,9 @@ impl VoronoiGrid {
         mut matches: impl FnMut(Ownership) -> bool,
     ) -> Option<(usize, f32)> {
         let start_index = self.point_to_index(point)?;
+        if matches(self.tiles[start_index]) {
+            return Some((start_index, 0.0));
+        }
 
         let mut distance = vec![f32::INFINITY; self.tiles.len()];
         let mut queue = BinaryHeap::new();
@@ -381,7 +394,9 @@ impl VoronoiGrid {
             if matches(self.tiles[current_index]) {
                 return Some((current_index, current_cost));
             }
-            for (neighbor_index, neighbor) in self.neighbor_indices(current_index) {
+            for (neighbor_index, neighbor) in
+                neighbor_indices(current_index, self.width_tiles, self.height_tiles)
+            {
                 let new_cost = current_cost + neighbor.step_cost;
                 if new_cost < distance[neighbor_index] {
                     distance[neighbor_index] = new_cost;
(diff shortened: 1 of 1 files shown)


schmidma (commented):
~~redacted opencode accident~~ :sweat_smile:

crates/voronoi/src/lib.rs
```diff
@@ -339,45 +328,44 @@ impl VoronoiGrid {
         ))
     }
 
-    fn nearest_non_blocked_cell_index(&self, point: Point2<Field>) -> Option<usize> {
+    fn nearest_free_cell_index(&self, point: Point2<Field>) -> Option<(usize, f32)> {
```
schmidma: The new `start_cost` is the core behavior change, but the API shape makes that hard to see at the call site: `nearest_free_cell_index` sounds like it returns only an index while it actually returns `(index, initial_cost)`.

Could we introduce a named return type, e.g. `SeedCell { index, cost }`, and rename this to something like `nearest_free_seed`? That would make the invariant explicit and avoid tuple-position mistakes as this logic evolves.

crates/voronoi/src/lib.rs
```diff
@@ -339,45 +328,44 @@ impl VoronoiGrid {
...
+        self.nearest_cell_index(point, |ownership| ownership.is_free())
+    }
+
+    fn nearest_cell_index(
```
schmidma: Is this generalization actually necessary and used somewhere? The only usage to me is `is_free()` this feels like premature optimization/unnecessary layers of abstraction

crates/voronoi/src/lib.rs
```diff
@@ -339,45 +328,44 @@ impl VoronoiGrid {
...
+            return Some((start_index, 0.0));
         }
 
-        let width_tiles = self.width_tiles;
-        let height_tiles = self.height_tiles;
-        let start_x = start_index % width_tiles;
-        let start_y = start_index / width_tiles;
-        let max_radius = self.width_tiles.max(self.height_tiles);
+        let mut distance = vec![f32::INFINITY; self.tiles.len()];
+        let mut queue = BinaryHeap::new();
```
schmidma: This allocates a full `distance` vector and a `BinaryHeap` for every robot whose current tile is not free, before the main Voronoi Dijkstra allocates/searches the grid as well.

Since this now runs in the playing behavior path, can we avoid per-seed scratch allocation? Reusing scratch buffers from `seed_sources`, or using an allocation-free bounded/ring search for the nearest free seed, would preserve the new semantics without adding another grid-wide search per robot.

crates/voronoi/src/lib.rs
```diff
@@ -339,45 +328,44 @@ impl VoronoiGrid {
...
+        queue.push(Reverse((NotNan::new(0.0).unwrap(), start_index)));
 
-            for x in min_x..=max_x {
-                for y in [min_y, max_y] {
-                    let index = self.index_from_xy(x, y);
-                    if self.tiles[index].is_free() {
-                        return Some(index);
-                    }
-                }
+        while let Some(Reverse((current_cost, current_index))) = queue.pop() {
```
schmidma: Could we split the algorithm mechanics from the domain decision here? This helper now combines queue update logic, stale-entry handling, neighbor expansion, predicate matching, and seed-cost semantics in one function.

crates/voronoi/src/lib.rs
```diff
@@ -416,3 +404,20 @@ fn rotation_cost(sin_h: f32, cos_h: f32, neighbor: Neighbor, orientation_bias: f
...
+) -> impl Iterator<Item = (usize, Neighbor)> {
+    let x = index % width_tiles;
+    let y = index / width_tiles;
+    NEIGHBORS.into_iter().filter_map(move |neighbor| {
+        let nx = x as isize + neighbor.dx;
+        let ny = y as isize + neighbor.dy;
+        if !(0..width_tiles as isize).contains(&nx) || !(0..height_tiles as isize).contains(&ny) {
+            return None;
+        }
+        Some((ny as usize * width_tiles + nx as usize, neighbor))
```
schmidma: This reimplements the grid-index mapping manually instead of using `index_from_xy`. That makes this helper subtly diverge from the rest of `VoronoiGrid` and keeps low-level x/y/index manipulation spread across the file.

Could this be a `VoronoiGrid::neighbor_indices(&self, index)` method and call `self.index_from_xy(...)`, or otherwise centralize the x/y-to-index conversion?

schmidma (changes requested):

crates/voronoi/src/lib.rs
JackyBloxx: i think that is a good idea
is now introduced a general nearest_matching_cell to also accomadate the neares owner for the nearest ball condition

crates/voronoi/src/lib.rs
JackyBloxx: with a new merged PR we also use this for getting the nearest owner to a cell

crates/voronoi/src/lib.rs
JackyBloxx: moved it so for every Robot uses the same distance vector and BinaryHeap

crates/voronoi/src/lib.rs
JackyBloxx: im not sure how i could do than properly

crates/voronoi/src/lib.rs
JackyBloxx: changed that all functions so that they use index_from_xy and moved neighbor inside VoronoiGrid

crates/voronoi/src/lib.rs
schmidma: as discussed, pushing back to you, let me know how you decide

schmidma (approved):

PR #2622, opened 2026-06-13
Port localization to ros-z

paudar:
## Why? What?

This PR ports the localization node to the ros-z node setup and connects it to ordered future streams for odometry, IMU, fall-down state, and line data.

It also updates the producing nodes to use announcing publishers where required, and changes `ground_to_field` to an optional output so consumers can handle states where localization intentionally has no pose.

Part of #2426 

## ToDo / Known Issues

tune paramerts

make the loca 3D

## How to Test

- Run `cargo fmt --check`.
- Run `cargo check -p line_detection -p localization -p odometer_bridge -p low_state_bridge -p fall_down_state_receiver -p behavior_node`.
- Run the code on the robot and verify that localization publishes `ground_to_field` and `is_localization_converged`, e.g. with a new twix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6432,15 +6432,20 @@ dependencies = [
 name = "localization"
 version = "0.1.0"
 dependencies = [
+ "approx",
  "booster",
  "color-eyre",
  "coordinate_systems",
+ "filtering",
  "geometry",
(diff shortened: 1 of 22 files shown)


paudar:
Closed, since the new loca is so amazing 🥲

PR #2623, opened 2026-06-13
Align with ball in walk_to_centroid

JackyBloxx:
## Why? What?

Aligns with the ball and look at it when the robots goes to its voronoi centroid
- Fixes #

Probably in the future smarter rotation  and head movment to not walk backward toward the possition.

## How to Test
enforce that the robot to calculates the voronoi grid and walk_to_centroid afterward.
then the robot should walk aligned to the ball backwards to the centroid position.

### DIFF
--- a/crates/nodes/behavior_node/src/tree.rs
+++ b/crates/nodes/behavior_node/src/tree.rs
@@ -131,7 +131,6 @@ fn supporter_subtree() -> Node<Blackboard> {
             sequence!(action!(calculate_voronoi_grid), action!(walk_to_centroid)),
             action!(stand)
         ),
-        subtree!(look_at_ball_subtree)
     )
 }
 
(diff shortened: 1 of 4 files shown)


crates/nodes/behavior_node/src/tree.rs
```diff
@@ -130,7 +130,8 @@ fn supporter_subtree() -> Node<Blackboard> {
         selection!(
             sequence!(action!(calculate_voronoi_grid), action!(walk_to_centroid)),
             action!(stand)
-        )
+        ),
+        subtree!(look_at_ball_subtree)
```
pejotejo: it is nice, but dont do it twice (line 129)

crates/world_state/src/behavior/tree.rs
```diff
@@ -126,7 +126,8 @@ fn supporter_subtree() -> Node<Blackboard> {
         selection!(
             sequence!(action!(calculate_voronoi_grid), action!(walk_to_centroid)),
             action!(stand)
-        )
+        ),
+        subtree!(look_at_ball_subtree)
```
pejotejo: same here

crates/nodes/behavior_node/src/tree.rs
JackyBloxx: done

crates/world_state/src/behavior/tree.rs
JackyBloxx: done

pejotejo (approved):
lgtm

PR #2624, opened 2026-06-14
Add min walk speed and dynamic target speed

### LINKED ISSUE #2615: post-game task: overshoot at high walk speed
We currently overshoot, when walking at high walk speeds. We probably want to slow down when getting close to the target.

BenSampaolo:
## Why? What?

- Fixes: #2615

Currently the robot overshoots his position targets when max speed is significantly increased. With this PR the speed is automatically throttled down to a minimum speed after coming into range of the target position. The minimum speed and the distance from where the slow-down begins are both behavior parameters.

## Known issues

This does not help when the turning rate is too slow, in this case the robot could walk a large circle around the position before directly approaching it. 

## How to Test

Upload to the robot and use Twix to manually increase the max speed. When directly approaching the target the robot should slow down before reaching its target position. A situation where the robot tries to reach a position that's anchored in the field coordinate system can be triggered by using the game controller and making the robot walk to kickoff for example.

### DIFF
--- a/crates/nodes/behavior_node/src/node.rs
+++ b/crates/nodes/behavior_node/src/node.rs
@@ -69,10 +69,60 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
     Box::pin(run(ctx))
 }
 
+fn validate_behavior_parameters(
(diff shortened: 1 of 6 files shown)


JackyBloxx (approved):
lgtm

PR #2626, opened 2026-06-14
Bevyhavior Simulator 2 Eclectic Boogaloo

knoellle:
## Why? What?

Ports the bevyhavior simulator to test the new behavior tree.
- Based on bevy with similar scenario API
- Visualizer with scrubbable timeline
  - Timeline is color coded by game state (similar to enum plot)
  - Scenarios can emit markers which are shown in the timeline to easily find relevant sections
  - twix-like behavior tree with node traces
  - twix-like map panel
  - filterable json-based state inspector
- Invariant checks that run every cycle.
  - Failing marks scenario as failed but it continues running.
  - Failures shown in visualizer.

## ToDo / Known Issues

- [x] Render spawned global obstacles
- [x] Play against opponent team
- [x] Render walk command paths
- [x] Simulate head motions/FOV
- [x] Change parameters to behavior_node.json5 insted of deafault.json
- [x] Fix hybrid alignment in motions (robots already turn when far away from destination)
- [x] Clean up old simulator files

- [ ] Add message count
- [ ] Voronoi Overlay

## How to Test

- Run scenario with viewer: `pepsi run --bin <scenario name>`
- Run all scenarios headless `cargo nextest run -p bevyhavior_simulator`


crates/bevyhavior_simulator/src/config.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub walk_translation_speed: f32,
+    pub walk_rotation_speed: f32,
+    pub walk_with_velocity_scale: f32,
+    pub kick_ball_speed_rumpelstilzchen: f32,
+    pub kick_ball_speed_schlong: f32,
+    pub kick_cooldown: Duration,
+    pub ball_friction_per_second: f32,
+    pub ball_visibility_range: f32,
+    /// Total field of view used for filtering ball and robot perception.
+    /// `visible = object_angle >= fov / 2`
```
BenSampaolo, line 20: Why?

crates/types/src/ball_position.rs
```diff
@@ -70,8 +70,8 @@ impl<From, To> Mul<BallPosition<From>> for Isometry2<From, To> {
     Serialize,
 )]
 pub struct SimulatorBallState {
-    pub position: Point2<Field>,
-    pub velocity: Vector2<Field>,
+    pub position: Point2<World>,
+    pub velocity: Vector2<World>,
```
BenSampaolo, line 74: Why not leave the ball in field and use field_to_world to convert?

tools/twix/src/behavior_tree.rs
BenSampaolo, line 1: Is this a simply moving code from tools/twix/src/panels/behavior_tree/mod.rs to here? If yes then why and how compatible is it with the current Twix development? @paudar doesnt like this :)

BenSampaolo (commented):
Some questions...

crates/types/src/ball_position.rs
knoellle: `Field` coordinate system is relative to the team side.
Why should the simulator hold the "physical" ball state in the coordinate system of one of the teams?

crates/bevyhavior_simulator/src/config.rs
knoellle: Why not? It is hard to express this the field name.

tools/twix/src/behavior_tree.rs
knoellle: We want to use this both in twix and in the simulator viewer.
There are basically two options:
- duplicate the code
- put the code in a location where both can use it

I chose the latter, do you have a better idea? Alternatively we could leave it inside the panel and then `use twix::panels::behavior_tree::{ .. }`

BenSampaolo: Apparently its easier the way you proposed :)

BenSampaolo (approved):
LGTM

PR #2627, opened 2026-06-14
fix(ball_filter): actually use odometry

alexschmander:
## Why? What?

Because of how the future map was used previously in the ball filter, the odometry was never used to predict the hypothesis. This PR fixes that.

## How to Test

### DIFF
--- a/crates/nodes/ball_filter/src/lib.rs
+++ b/crates/nodes/ball_filter/src/lib.rs
@@ -241,7 +241,6 @@ fn predict_hypotheses_from_odometry(
     );
 }
 
-#[allow(clippy::too_many_arguments)]
 fn advance_all_hypotheses(
     ball_filter: &mut BallFilter,
     time: Time,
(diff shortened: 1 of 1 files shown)


crates/nodes/ball_filter/src/lib.rs
```diff
@@ -219,6 +239,20 @@ fn advance_all_hypotheses(
         Matrix2::from_diagonal(&filter_parameters.noise.process_noise_resting),
         filter_parameters.log_likelihood_of_zero_velocity_threshold,
     );
+}
+
+#[allow(clippy::too_many_arguments)]
```
BenSampaolo: No longer needed

BenSampaolo (changes requested):
Looks good, minor beauty change

crates/nodes/ball_filter/src/lib.rs
alexschmander: Removed

BenSampaolo (approved):
LGTM!

PR #2630, opened 2026-06-16
Migrate Booster motion stack to ros-z interface

schmidma:
## Why this PR exists

This PR moves Booster robot execution onto the ROS-Z interface path.

The previous ROS-Z Booster setup split robot side effects across several executor nodes: walking, robot mode handling, stand-up, head rotation, kick publishing, and low-command publishing. That made it hard to reason about which node owned the Booster SDK state, which command topic was current, and which timing rules applied to SDK calls.

After this PR, behavior publishes one semantic command stream on `behavior/motion_command`. ROS-Z head nodes derive head targets from that stream. The new `booster_interface` node owns the Booster SDK side effects from one place.

This gives reviewers one runtime path to inspect for Booster execution:

- Behavior selects `MotionCommand`.
- Head nodes publish `head_joints_command`.
- `booster_interface` polls/request Booster modes, sends walking commands, rotates the head, runs stand-up, handles visual-kick SDK state, forwards kick messages to `rt/kick_ball`, forwards LEDs, and keeps Booster mode requests serialized.

### Booster execution

- Adds `booster_interface` as the central ROS-Z runtime node for Booster SDK interaction.
- Removes the old split ROS-Z executor nodes for walking, stand-up, robot mode handling, head rotation, kick, low command publishing, and their stale parameters.
- Uses `Cache` inputs for latest-value `behavior/motion_command` and `head_joints_command`.
- Keeps slow SDK mode polling and mode requests separate from periodic move/head/kick output.
- Tracks one in-flight task per SDK side effect to avoid request floods.
- Handles Damping like every other Booster mode request so SDK mode changes stay serialized.

### Motion command and behavior

- Adds an explicit damping motion intent.
- Aligns behavior and ROS-Z motion topics on `behavior/motion_command`.
- Keeps passive Booster game states in `Stand`; the ROS-Z interface owns actual Booster mode requests.
- Aligns `WalkWithVelocity` motion-type tracking in behavior/world-state code.
- Removes unsupported `HeadMotion::LookAtReferee` instead of routing a command that had no implementation.

### Head motion

- Ports semantic ROS-Z head nodes for `head_motion`, `look_around`, and `look_at`.
- The ROS-Z head path computes `head_joints_command` from the latest motion command, camera/world inputs, and motor-state feedback.
- Reads measured head motor positions directly in the `look_at` fallback path.

### Visual kick

- Keeps Booster SDK visual-kick activation separate from `rt/kick_ball` publishing.
- Tracks visual-kick start/stop requests with an explicit enum instead of a boolean option.
- Documents manual validation for both SDK `visual_kick(true/false)` transitions and `rt/kick_ball` publishing.

## Migration note

This PR intentionally keeps some temporary overlap with the old framework motion code. The ROS-Z Booster/head nodes duplicate small pieces of old motion logic during the migration. Removing the old framework motion nodes is a later PR, so this PR avoids throwaway shared abstractions.

## Review guide

Suggested review order:

1. Start with `crates/nodes/booster_sdk_interface/src/lib.rs`.
   - Check the scheduling loop, mode polling/request lifecycle, mode-request serialization, stale mode-poll handling, stand-up lifecycle, visual-kick lifecycle, and per-effect in-flight tracking.
2. Then read `crates/nodes/booster_sdk_interface/src/control.rs`.
   - Check `MotionCommand` to Booster desired mode, step, and kick conversion.
3. Review the ROS-Z head nodes:
   - `crates/nodes/head_motion/src/lib.rs`
   - `crates/nodes/look_around/src/lib.rs`
   - `crates/nodes/look_at/src/lib.rs`
4. Review behavior/topic wiring:
   - `crates/nodes/behavior_node/src/node.rs`
   - `crates/nodes/behavior_node/src/tree.rs`
   - `crates/world_state/src/behavior/node.rs`
5. Review config deletion/addition last:
   - `etc/parameters/ros_z/base/booster_interface.json5`
   - removed stale ROS-Z executor parameter files

The highest-risk code is the Booster SDK lifecycle code. The main safety properties to check are:

- Damping, Walking, and Prepare mode requests use one serialized SDK request lifecycle.
- Stale SDK mode polls do not restore old confirmed modes across mode-request generations.
- Stale SDK task completions do not clear newer mode or stand-up request state.
- Stand-up command flaps do not start overlapping `get_up()` calls.
- Walking/head/kick side effects only run after confirmed Walking mode.
- Visual kick stops when the command ends or walking becomes disallowed.

## Validation

Local verification after rebasing and rewriting the PR history:

- [x] `cargo nextest run -p booster_sdk_interface -p head_motion -p look_at -p look_around -p types --no-fail-fast`
- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo check --workspace --all-targets`
- [x] `cargo nextest run --workspace --no-fail-fast`
- [x] `cargo run --bin parameter_tester`
- [x] `git diff --check`


crates/nodes/head_motion/src/lib.rs
```diff
@@ -22,29 +27,298 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
...
+        Some(HeadMotion::Center {
+            image_region_target: ImageRegion::Top,
+        }) => HeadJoints {
+            yaw: 0.0,
+            pitch: 0.4,
+        },
+        Some(HeadMotion::Center { .. }) => HeadJoints {
+            yaw: 0.0,
+            pitch: 0.4,
+        },
```
rmburg, line 182: I know, not your code, but should we merge these match arms? or should the two cases contain different joint values? Not necessarily in this PR.

rmburg:
please rebase and resolve merge conflicts

schmidma:
rebase done, and force push :+1:

crates/nodes/robot_mode_handler/src/lib.rs
```diff
@@ -1,117 +0,0 @@
...
-        .qos(QosProfile {
-            durability: QosDurability::TransientLocal,
-            ..Default::default()
-        })
-        .build()
-        .await?;
-    let buttons_sub = node
-        .subscriber::<Buttons<Option<ButtonPressType>>>("buttons")
-        .build()
-        .await?;
```
rmburg, line 39: button handling seems to have been removed but not re-added

etc/parameters/ros_z/base/walking.json5
```diff
@@ -1,69 +0,0 @@
-{
-  parameters: {
-    gait_frequency: 1,
-    stabilizing_interval_completion_threshold: 0.01,
-    stabilizing_interval_compression_factor: 1.0,
-    number_of_actions: 12,
-    number_of_observations: 47,
-    torque_limits: {
-      head: {
```
rmburg, line 9: many of these parameters were removed without replacement, are they no longer needed?

etc/parameters/ros_z/base/command_sender.json5
rmburg, line 1: many of these parameters were removed without replacement, are they no longer needed?

etc/parameters/ros_z/base/robot_mode_handler.json5
```diff
@@ -1,7 +0,0 @@
-{
-  wait_before_prepare: {
-    nanos: 0,
-    secs: 2,
-  },
```
rmburg, line 5: this parameter was removed without replacement, is it no longer needed?

docs/robotics/motion/overview.md
```diff
@@ -17,3 +17,23 @@ The nodes, whose motion is not selected, may exit early.
...
+
+- F1 or stand short press toggles local emergency damping.
+- Changing `booster_interface.remote_stop_toggle` changes emergency damping according to the XOR rule above.
+- Walking commands produce Booster SDK `move_robot` calls only after the SDK reports `Walking` mode.
+- `head_joints_command` produces Booster SDK `rotate_head` calls when walking is allowed.
+- Stand-up commands produce Booster SDK `get_up` requests in `Prepare` or `Damping` mode and are suppressed while emergency damping is active.
+- Visual kick commands start Booster SDK `visual_kick(true)` while walking is allowed.
+- Visual kick commands stop Booster SDK `visual_kick(false)` when the command ends or walking becomes disallowed.
+- Visual kick commands publish `rt/kick_ball` while walking is allowed.
+- Unknown SDK robot mode ids produce a log entry containing the raw id and keep walking disabled.
```
rmburg: note to self: confirm this is correct on real hardware

crates/nodes/behavior_node/src/tree.rs
```diff
@@ -47,9 +47,10 @@ pub fn create_tree() -> Node<Blackboard> {
         sequence!(
             selection!(
                 condition!(is_primary_state, PrimaryState::Initial),
-                condition!(is_primary_state, PrimaryState::Penalized)
+                condition!(is_primary_state, PrimaryState::Penalized),
+                condition!(is_primary_state, PrimaryState::Finished)
             ),
-            sequence!(action!(look_straight_ahead), action!(stand))
+            action!(prepare)
```
rmburg, line 53: phillip sagt das will man nicht so

crates/nodes/booster_sdk_interface/src/lib.rs
```diff
@@ -203,74 +400,1209 @@ async fn handle_led_command(
...
+    if matches!(
+        state.mode_request,
+        ModeRequestState::InFlight {
+            desired_mode: in_flight_mode,
+            ..
+        } if desired_mode == control::DesiredMode::Damping
+            && in_flight_mode != control::DesiredMode::Damping
+    ) {
+        return true;
+    }
```
rmburg: This allows damping requests to be sent while a non-damping request is still in flight. Doesn't this lead to a race condition? If an old non-damping request is still in flight, and we send a damping request before the old request finishes, the old request could put the robot back into whatever mode was requested after we set it to damping.

crates/nodes/booster_sdk_interface/src/lib.rs
```diff
@@ -1,183 +1,380 @@
...
-                    handle_led_command(light_control_client, led_command)
+                let light_control_client = light_control_client.clone();
+                tokio::spawn(handle_led_command(light_control_client, led_command?));
+            }
+            _ = tick.tick() => {
+                let parameters_snapshot = parameters.snapshot();
+                let parameters = parameters_snapshot.typed();
+                let latest_motion_command = motion_command_cache
+                    .get_latest()
+                    .unwrap_or_else(|| default_motion_command.clone());
```
rmburg, line 323: This replays the last motion command indefinitely, even if the behavior node is stalled or crashed. Maybe we should fallback to a safe mode when we haven't received a new motion command after a while?

crates/bevyhavior_simulator/src/robot.rs
rmburg, line 403: Might be redundant since the bevyhavior simulator needs to be adapted to the new framework anyways, but `visual_referee_pose_kind` is now always `Some`, even if we're only looking at a ball. Maybe use `LookAtReferee` like before, or add some kind of target/purpose enum?

rmburg (changes requested):

crates/nodes/head_motion/src/lib.rs
schmidma: Leaving this unchanged in this PR. The duplicate arms mirror the existing legacy head-motion behavior, and changing the actual joint values would be head-motion tuning rather than part of the Booster ROS-Z migration.

crates/nodes/robot_mode_handler/src/lib.rs
schmidma: Changed direction here. I removed the remote/local emergency-stop feature surface from this PR instead of re-adding this part of `robot_mode_handler`.

Button events still flow through `button_event_bridge`, `button_event_handler`, and `primary_state_filter`. The Booster-specific remote/local stop handling will be a later incremental feature.

etc/parameters/ros_z/base/walking.json5
schmidma: The ROS-Z Booster path no longer uses the old `walking` executor node. The parameters that still apply to Booster walking moved to `booster_interface.walking` or `booster_interface.move_robot_message_interval`.

The remaining RL walking parameters are still present for the legacy/framework path in `etc/parameters/default.json`, but they are not consumed by the new ROS-Z Booster interface.

etc/parameters/ros_z/base/robot_mode_handler.json5
schmidma: This PR no longer carries the `robot_mode_handler` replacement parameter. We decided to remove the remote/local emergency-stop handling from this PR and leave it for a later incremental feature.

The old `wait_before_prepare` parameter was declared by the deleted ROS-Z node but not used there, so it does not need a replacement in the current Booster path.

crates/nodes/booster_sdk_interface/src/lib.rs
schmidma: Fixed. Damping no longer supersedes an in-flight non-damping mode request. `Damping`, `Prepare`, and `Walking` now use the same serialized mode-request lifecycle.

I also guarded mode-poll results by mode-request generation so an old poll cannot restore a stale confirmed mode after a mode request starts.

etc/parameters/ros_z/base/command_sender.json5
schmidma: These belonged to the deleted ROS-Z low-command publisher / command-sender path. The Booster ROS-Z path no longer publishes low motor commands through `commands_sender`; it sends SDK-level mode, walking, head, stand-up, kick, and LED effects through `booster_interface`.

The legacy framework command-sender parameters remain in `etc/parameters/default.json`.

crates/nodes/booster_sdk_interface/src/lib.rs
schmidma: Valid concern, but I am leaving it out of this PR to keep the migration scope bounded. A motion-command freshness timeout should be a follow-up safety feature with its own parameter and tests.

crates/nodes/behavior_node/src/tree.rs
schmidma: Fixed. `Initial`, `Penalized`, and `Finished` now publish `Stand` with `look_straight_ahead` again instead of direct `Prepare`.

Added a behavior test for those passive primary states.

crates/bevyhavior_simulator/src/robot.rs
schmidma: Agreed that generic `LookAt` is not enough to distinguish ball gaze from referee gaze. I am leaving this for the simulator/new-framework adaptation instead of adding `LookAtReferee` or a gaze-purpose enum in this migration PR.

rmburg: Agreed, a proper fix is out of scope for this PR. Maybe leave a comment in case anyone ports this code in the future, since this technically introduces a new bug.

schmidma:
Superseded by #2677

PR #2633, opened 2026-06-17
Expose ros-z graph change revisions

schmidma:
## Summary

This PR adds revision tracking to the `ros-z` graph. Each graph mutation now advances a monotonic revision, and graph snapshots carry the revision that produced them. Code that waits for graph-dependent state can now wait for concrete graph changes instead of relying on repeated queries or local polling loops.

The change introduces shared wait primitives for graph consumers and migrates publisher/subscriber waits, transient-local replay, and dynamic schema discovery onto those primitives.

## Motivation

Several `ros-z` paths need to wait until graph state catches up with Zenoh liveliness updates. Before this PR, those paths implemented their own waiting behavior and deadline handling. That made it harder to reason about missed updates, stale snapshots, and timeout behavior.

Revision tracking gives the graph a single observable progress signal. Waiters can now ask: “Has the graph changed since the snapshot I looked at?” Dynamic schema discovery also uses one deadline budget while waiting for publishers and querying schema candidates.

## Reviewer Notes

Start with these areas:

- `crates/ros-z/src/graph.rs`
  - Defines graph revisions and snapshot-facing graph behavior.
  - Adds tests for revision ordering and snapshot revision behavior.

- `crates/ros-z/src/graph/state.rs`
  - Advances and publishes graph revisions after mutations.
  - Sends graph change notifications after releasing the graph mutex.

- `crates/ros-z/src/graph/query.rs`
  - Adds `Graph::wait_until`.
  - Adds `GraphChangeSubscription` so public subscribers observe copied revisions without exposing raw watch receivers.

- `crates/ros-z/src/graph/discovery.rs`
  - Drives graph revisions from Zenoh liveliness history and live `Put`/`Delete` events.

- `crates/ros-z/src/pubsub/`
  - Migrates publisher/subscriber waits and transient-local replay to the shared graph wait logic.

- `crates/ros-z/src/dynamic/discovery.rs`
  - Uses graph revision waits while discovering dynamic schemas.
  - Applies one absolute deadline across publisher waits and schema-service candidate queries.

This PR also removes the bounded initial graph query configuration path. Graph progress now comes from liveliness-driven revisions instead.

## Follow-up

- Transient-local replay with late-discovered publishers is tracked separately in [the issue created during this PR update](https://github.com/HULKs/hulk/issues/2643). PR #2633 exposes graph revisions and makes this easier to observe, but the replay correctness issue exists independently of this PR.

## Scope Notes

- `Graph` remains a live local observation of liveliness state. `Graph::new` intentionally does not define when the distributed graph is ready or complete.
- CLI/operator readiness policy should live above `Graph`, using graph revisions as one available signal. Open a follow-up issue after merge to define the desired `rosz` startup/settling behavior.
- Dynamic schema discovery keeps the normal lifecycle assumption that publishers remain present after discovery; publisher churn during discovery can be handled separately if it becomes observable in practice.

## API Notes

- This PR intentionally removes the configurable initial graph liveliness query path. `ros-z` has no downstream API users outside this mono-repo, so the PR does not keep compatibility shims for that configuration API.
- `Graph` is a live local observation of liveliness state. Graph revisions are local progress tokens, not readiness or completeness markers.

### DIFF
--- a/crates/ros-z-cli/src/model/watch.rs
+++ b/crates/ros-z-cli/src/model/watch.rs
@@ -1,14 +1,38 @@
-use ros_z::graph::GraphSnapshot;
+use ros_z::graph::{GraphRevision, GraphSnapshot};
 use serde::Serialize;
 
 #[derive(Debug, Clone, Serialize)]
 #[serde(tag = "event", rename_all = "snake_case")]
 pub enum WatchEvent {
-    InitialState { snapshot: GraphSnapshot },
-    TopicDiscovered { name: String, type_name: String },
-    TopicRemoved { name: String },
-    NodeDiscovered { namespace: String, name: String },
-    NodeRemoved { namespace: String, name: String },
-    ServiceDiscovered { name: String, type_name: String },
-    ServiceRemoved { name: String },
+    InitialState {
+        snapshot: GraphSnapshot,
+    },
+    TopicDiscovered {
+        revision: GraphRevision,
+        name: String,
+        type_name: String,
+    },
+    TopicRemoved {
+        revision: GraphRevision,
+        name: String,
+    },
+    NodeDiscovered {
+        revision: GraphRevision,
+        namespace: String,
+        name: String,
+    },
+    NodeRemoved {
+        revision: GraphRevision,
+        namespace: String,
+        name: String,
+    },
+    ServiceDiscovered {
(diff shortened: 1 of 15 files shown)


paudar (approved):
LGTM

PR #2634, opened 2026-06-17
Expose ros-z-debug subscription updates

schmidma:
## Summary

This PR lets users subscribe to live updates from `ros-z-debug` subscriptions.

A user can now keep a handle to the latest received data, and also wait for changes on that handle. Updates report new data, status changes with retained diagnostic messages, and missed messages when a receiver falls behind.

Closing a subscription closes the update stream. The last retained data and status stay available through the handle.

## Motivation

Debug tools often need to react when subscription data changes. Before this PR, callers could only drain a small internal event buffer. That mixed two different jobs: storing recent state and notifying live consumers.

This PR makes the split clearer:

- use `latest()` and `window()` to inspect retained data
- use `subscribe_updates()` to wait for future changes

## Reviewer Notes

The main API change is in `crates/ros-z-debug/src/subscription.rs`:

- `drain_events()` is replaced by `subscribe_updates()`
- update receivers only see future updates
- closed subscriptions return `SubscriptionUpdateClosed`

The event types live in `crates/ros-z-debug/src/event.rs`.

The subscription lifecycle also moved into `SubscriptionState::spawn`. This keeps live update stream closure, receive-task cancellation, and weak-state receive-loop ownership in one place. The receive task now exits cooperatively through the existing `CancellationToken` instead of storing and aborting an `AbortHandle` from the builders.

## Test Plan
- `cargo clippy -p ros-z-debug --all-targets --locked -- -D warnings`

### DIFF
--- a/crates/ros-z-debug/src/event.rs
+++ b/crates/ros-z-debug/src/event.rs
@@ -1,86 +1,140 @@
-use std::collections::VecDeque;
+use tokio::sync::broadcast::{
+    Receiver,
+    error::{RecvError, TryRecvError},
+};
 
+use crate::SubscriptionStatusSnapshot;
+
+/// Live notification emitted by a retained debug subscription.
+///
+/// Updates are only delivered to receivers that were already subscribed when the
+/// event happened. Terminal closure is represented by `SubscriptionUpdateClosed`,
+/// not by a `SubscriptionUpdate` variant.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
-pub enum DebugEvent {
-    /// The subscription status snapshot changed.
-    StatusChanged,
+pub enum SubscriptionUpdate {
     /// A new sample was retained as the latest value.
-    ValueUpdated,
-    /// A non-terminal diagnostic message was recorded.
-    Diagnostic(String),
+    DataChanged,
+    /// A non-terminal subscription status snapshot changed.
+    ///
+    /// Terminal close is reported by `SubscriptionUpdateClosed`; use the
+    /// subscription handle's `status()` to inspect retained status.
+    StatusChanged(SubscriptionStatusSnapshot),
+    /// This receiver fell behind and missed updates.
+    Lagged { dropped: u64 },
 }
 
-pub struct EventBuffer {
-    capacity: usize,
-    events: VecDeque<DebugEvent>,
+/// Error returned when a subscription's live update stream has ended.
+///
+/// After this error, no future updates can arrive on that receiver. Use the
+/// subscription handle's retained state methods to inspect final status and data.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[error("subscription update stream closed")]
+pub struct SubscriptionUpdateClosed;
+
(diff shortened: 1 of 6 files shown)


crates/ros-z-debug/src/subscription.rs
```diff
@@ -119,18 +138,35 @@ impl<V> SubscriptionState<V> {
...
+    where
+        F: FnOnce(std::sync::Weak<Self>, CancellationToken) -> Fut,
+        Fut: std::future::Future<Output = ()> + Send + 'static,
+        V: Send + Sync + 'static,
+    {
+        let state = Arc::new(Self::new(status, retention));
+        let weak_state = Arc::downgrade(&state);
+        let cancellation = state.cancellation_token();
+
+        std::mem::drop(tokio::spawn(receive_loop(weak_state, cancellation)));
```
rmburg: Verständnisfrage: why manually drop the `JoinHandle` here?

schmidma: The explicit `drop(...)` was there to make the intentional discard of the spawned task handle obvious. Dropping a Tokio `JoinHandle` detaches the task, which is the behavior we want here; shutdown is handled separately through the `CancellationToken`. I also checked that bare `tokio::spawn(...);` does not warn with our Clippy command, so I removed the explicit `drop(...)` anyway.

crates/ros-z-debug/src/event.rs
```diff
@@ -1,86 +1,140 @@
...
 #[non_exhaustive]
-pub enum DebugEvent {
-    /// The subscription status snapshot changed.
-    StatusChanged,
+pub enum SubscriptionUpdate {
     /// A new sample was retained as the latest value.
-    ValueUpdated,
-    /// A non-terminal diagnostic message was recorded.
-    Diagnostic(String),
+    DataChanged,
```
rmburg, line 17: The `DataChanged` update does not include the value of the latest sample (e.g. an `Arc<SampleRecord<V>>`).
I'm wondering if it would be better to include it... If the receiver is very slow, it might not be able to reconstruct every change event since only the very last value is retained in `state.latest`. Or maybe I'm misunderstanding the how the API is intended to be used

schmidma: I think we should keep `DataChanged` stateless. The update stream is meant to wake consumers up when retained state changes, while `latest()` and `window()` remain the source of truth for data. Including the sample in the notification would turn this into a delivery stream with different guarantees, especially around slow receivers. If a caller needs to reconstruct samples, it should opt into window retention instead of relying on `LatestOnly`.

rmburg (approved):
LGTM

crates/ros-z-debug/src/event.rs
rmburg: I'm not entirely convinced. I think most (all?) receivers that are interested in changes in the data also require the new latest value, and keeping track of a window to retrieve lost updates seems wasteful to me. Including the latest datum in the change would also make subscribers (e.g. enum plot panel) a bit easier to implement, and we wouldn't need logic for reconstructing updates from retained values. Anyway, let's merge this for the sake of velocity and discuss this in-person later.

PR #2636, opened 2026-06-17
Make Twix ros-z subscriptions event-driven

schmidma:
Depends on #2635; GitHub diff includes earlier commits until that base merges.

## Summary
- Drive Twix ros-z buffers from subscription events instead of polling retained history.
- Preserve explicit buffer history semantics while rebuilding subscriptions on retarget/history changes.
- Add retained sample timing and latency support used by event-driven updates.

## Test Plan
- cargo nextest run -p twix
- cargo check -p twix
- cargo clippy -p twix --all-targets --locked -- -D warnings
- git diff --check


schmidma:
Closed because the event-driven wake/lifecycle work is now absorbed into the retained-state core in #2635. The branch tree matched #2635 after the retained-state rewrite, so this draft no longer carries unique reviewable changes.

PR #2640, opened 2026-06-20
Clips Centroid calculation to Field bounds

JackyBloxx:
## Why? What?

Introduces a min and max limit for the voronoi and the centroid.
Only consider grid points inside the field for the centroid calculation.
- Fixes #

## How to Test

Change the max and min values in voronoi.rs and check if the centroid and the field moves accordingly .
the centroid should be bound by the centroid min and max and the field by the voronoi min and max.


BenSampaolo (approved):
Viewed
Viewed
Viewed

PR #2645, opened 2026-06-24
Hydra model output renderer

alexschmander:
## Why? What?

This PR adds a hydra model output renderer cli to visualize the outputs of hydra models.


knoellle:
Decided to commit this to thesis repo instead.

PR #2646, opened 2026-06-24
Hydra bench no engines

knoellle:
## Why? What?

Adds a pepsi command and a tool for benchmarking onnx models on an ssh target.
It measures execution latency of a single model or multiple models in lockstep-parallel.

## How to Test

`pepsi hydra-bench 99 tools/machine-learning/multi-task-yolo/runs/train/*/weights/best.onnx`

This will
- build the benchmark binary and the tensorrt-compile binary
- rsync both binaries and all given neural nets to peter-fox ("robot" nr. 99)
- run tensorrt-compile
- return the compiled engines
- run the hydra-bench binary with each neural network one at a time
- return the benchmark result json

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8188,6 +8188,7 @@ dependencies = [
  "futures-util",
  "glob",
  "indicatif",
+ "itertools 0.14.0",
  "lazy_static",
  "parameters",
  "pathdiff",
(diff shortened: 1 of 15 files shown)


alexschmander (approved):
LGTM

PR #2647, opened 2026-06-25
feat(ros-z-cli): add topic frequency estimation

schmidma:
## Overview

Adds `rosz hz` for topic frequency estimation. The command reports local receive-rate stats plus per-publisher source-rate stats, with compact text output and JSON-lines support.

## Motivation

`rosz echo` is useful for inspecting payloads, but frequency checks only need raw samples and timing metadata. This gives operators a quick `ros2 topic hz`-style check without requiring schema discovery or payload decoding.

## Implementation Guide

- `ros-z`: extends automatic dynamic subscriber discovery with `.raw()`, discovering publisher `TypeInfo` from graph metadata and building a raw subscriber without querying schema services.
- `ros-z-cli`: adds `hz` parsing, rolling interval stats, count/duration/continuous modes, compact text rendering, and JSON report objects.
- Tests cover parser validation, rolling stats, duration scheduling edge cases, and an e2e publisher with schema service disabled.

## Verification

- `cargo fmt --check`
- `cargo test --doc -p ros-z`
- `cargo nextest run -p ros-z-cli`

### DIFF
--- a/crates/ros-z-cli/src/commands/hz.rs
+++ b/crates/ros-z-cli/src/commands/hz.rs
@@ -1,0 +1,323 @@
+use std::{
+    future::Future,
+    io::{self, Write},
+    num::NonZeroUsize,
+    pin::Pin,
+    time::{Duration, Instant},
+};
+
+use color_eyre::eyre::{Result, WrapErr, eyre};
+use ros_z::{attachment::Attachment, pubsub::RawSubscriber};
+use tokio::time::{MissedTickBehavior, Sleep};
+
+use crate::{
+    app::AppContext,
+    cli::HzLimit,
+    model::hz::{HzEstimator, HzReport},
+    render::{OutputMode, json, text},
+};
+
+const TYPE_DISCOVERY_TIMEOUT: Duration = Duration::from_secs(5);
+const REPORT_PERIOD: Duration = Duration::from_secs(1);
+
+#[derive(Default)]
+struct HzReceiveState {
+    warned_invalid_attachment: bool,
+}
+
+fn decode_attachment_for_hz(
+    attachment: Option<&zenoh::bytes::ZBytes>,
+    state: &mut HzReceiveState,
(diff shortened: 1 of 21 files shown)


rmburg:
this has conflicts, please resolve

crates/ros-z-cli/src/commands/hz.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    #[cfg(unix)]
+    fn capture_stderr(mut f: impl FnMut()) -> String {
+        use std::{
+            fs::File,
+            io::{Read, Seek, SeekFrom, Write},
+            os::fd::{AsRawFd, FromRawFd, RawFd},
+        };
+
+        unsafe extern "C" {
```
oleflb: add test dependency to `libc` or even `nix` for safe bindings?

crates/ros-z-cli/src/model/hz.rs
```diff
@@ -0,0 +1,331 @@
...
+        let variance = self
+            .intervals
+            .iter()
+            .map(Duration::as_secs_f64)
+            .map(|seconds| {
+                let delta = seconds - mean;
+                delta * delta
+            })
+            .sum::<f64>()
+            / intervals as f64;
```
oleflb: nit: this is biased, prefer (n-1) estimate

oleflb (changes requested):
Looks good. I don't really care about findings. If you want we can merge now

crates/ros-z-cli/src/commands/hz.rs
schmidma: removed the test

crates/ros-z-cli/src/model/hz.rs
schmidma: yes, done

oleflb (approved):

PR #2648, opened 2026-06-25
Fix Playing Subtree

JackyBloxx:
## Why? What?

add calculate_voronoi before is_closeset to ball so that it is calculatetd before it is used in the new framework
- Fixes #

## How to Test

Check if the Robot follows the Ball when it is closest to it in playing .
You could check that with 2 robots on the field with different playernumbers or in the simulator.


BenSampaolo (approved):
lgtm

PR #2652, opened 2026-06-27
Add ros-z-debug topic observer

schmidma:
Closed in favor of #2653, which uses `schmidma:ros-z-topic-observer` as the PR head instead of an upstream `HULKs` branch.


schmidma:
Closing in favor of #2653, which keeps the same branch contents but uses `schmidma:ros-z-topic-observer` as the PR head instead of an upstream `HULKs` branch.

PR #2653, opened 2026-06-27
Add ros-z-debug topic observer

schmidma:
## Summary
Adds a retargetable topic observation API to `ros-z-debug` for the ROS-Z-native Twix backend.

The PR keeps Twix UI code out of subscription lifecycle details. Callers can observe typed or dynamic topics, retarget them when the selected robot changes, read latest values synchronously, retain time-window samples, and recover when dynamic publishers or schema services appear after startup.

This PR does not migrate Twix panels.

## Architecture
Low-level cached subscriptions are now handle-owned:

- `Arc<Node>::cached_subscription(...)` creates a `CachedSubscriptionBuilder`.
- `build_typed`, `build_dynamic`, or `build_json` returns a `CachedSubscription<T>` or `CachedJsonSubscription`.
- The returned handle owns the receive task lifetime. Dropping the last handle stops the task.
- There is no cache-level factory or manager object.

Topic resolution lives in `TopicReference` and `TargetIdentity`:

- relative topics resolve under the active target namespace
- absolute topics stay fully qualified
- private topics like `~trace` require a target node name

The observation layer sits above cached subscriptions:

- `TopicObserver` owns shared target defaults and graph subscriptions.
- `TopicObservation<T>` and `DynamicTopicObservation` own retargetable cache handles.
- Observations rebuild or retry on topic, target, retention, reconnect, cache-stream, retry-timer, and relevant graph changes.
- During rebuilding, retrying, or blocked private-topic resolution, the previous cache remains readable but stops receiving old-target samples.

Dynamic JSON access renders retained `DynamicPayload` samples on demand. Metadata-preserving JSON records use `SampleRecord<serde_json::Value>`.

## Verification
Recorded on `c983dc04a`:

- `cargo fmt --check`
- `cargo clippy -p ros-z-debug --all-targets --locked -- -D warnings`


crates/ros-z-debug/src/topic.rs
```diff
@@ -96,27 +169,50 @@ impl TopicProjection {
...
+            .node_name()
+            .ok_or_else(|| Error::MissingTargetNodeName {
+                topic: topic.to_string(),
+            })?;
+        let private_topic = private_topic.trim_start_matches('/');
+        let topic = if private_topic.is_empty() {
+            node_name.to_string()
+        } else {
+            format!("{node_name}/{private_topic}")
+        };
```
paudar: This is also done in `qualify_topic_name`, please reduce duplication.

schmidma: Fixed with some cleanup in https://github.com/HULKs/hulk/pull/2653/changes/36f5e7348eb468ba6d4138c41deebf47c64abb38 ... there might be even cleaner solutions, but I'd defer them to future iterations/cleanup

paudar (approved):
It is a big PR, but I think it is good

PR #2654, opened 2026-06-29
feat(twix): add ROS-Z native text and image panel

schmidma:
## Summary
- replace primary Twix with a ROS-Z-native shell and dynamic Text panel
- move the previous websocket Twix implementation to `twix-legacy`
- add explicit `TopicObserver` runtime-handle support used by Twix
- update Twix docs/tooling for the ROS-Z CLI and legacy split

## Verification
- `cargo fmt --check`
- `cargo check -p twix`
- `cargo check -p twix-legacy`
- `cargo nextest run -p twix`
- `git diff --check`

Draft PR targeting `HULKs/main` as requested. This branch includes the ROS-Z graph/topic-observer stack needed by Twix.


docs/tooling/behavior_simulator.md
```diff
@@ -8,17 +8,9 @@ A simplified simulator which can be used for manual or automatic testing of beha
...
 It returns an error if the robotics code encountered a problem or if the scenario file generated an error.
 
-
-Within twix, connect to `localhost` and open a `Behavior Simulator` panel.
-This panel has a timeline slider for scrubbing through the scenario.
-
-!!! info
-
-    To see all robots on the map panel, make sure to enable the behavior simulator overlay. Otherwise only the selected robot is shown.
-
```
rmburg, line 21: We should make sure to re-add this documentation after #2626 is merged (if applicable)

docs/tooling/twix.md
rmburg, line 1: There is a lot of documentation in this file and others that we might want to keep for the future, after we ported the features documented here to the new twix. Suggestion: move old docs to a `twix-legacy.md` or similar to make it easier to find in the future

paudar: The documentation has been quite neglected, as there has simply been too much work in progress. 
I think the documentation should be revised after the RoboCup (and to find the deleted lines, you can check out the tag from last year’s RoboCup)

tools/twix-legacy/src/main.rs
```diff
@@ -0,0 +1,827 @@
...
+    fn title(&mut self, tab: &mut Self::Tab) -> eframe::egui::WidgetText {
+        match &mut tab.panel {
+            Ok(panel) => format!("{panel}").into(),
+            Err((error, _value)) => WidgetText::from(format!("{error}")).color(Color32::LIGHT_RED),
+        }
+    }
+
+    fn id(&mut self, tab: &mut Self::Tab) -> Id {
+        tab.id
+    }
```
paudar, line 822: This function was dropped in the new Twix. If i understand correctly, the id is not set to the egui_dock.

tools/twix/src/main.rs
```diff
@@ -1,285 +1,143 @@
...
+                            ));
+                            if let Err(error) = &tab.panel {
+                                load_error.get_or_insert_with(|| format!("{error:#}"));
+                            }
+                            tab
+                        });
+
+                        if let Some(error) = load_error {
+                            error!("failed to load dock tabs: {error}");
+                            None
```
paudar, line 124: If one saved tab fails to deserialize, the current loading path discards the entire saved dock state and falls back to the default layout. The old Twix kept the dock layout intact by storing the per-tab load error inside that tab and preserving the original JSON when saving again. Could we keep that behavior here as well, so a renamed/corrupt/incompatible panel only turns into an error tab instead of resetting all tabs? A small `TabLoadError { error, saved_value }` state would probably be enough.

paudar: (Maybe we don't need this feature on the RoboCup, but i think that it is useful in general)

paudar (commented):
It works and is a good starting point for a new Twix. 

I would like to be able to build `twix-legacy`, so please add it to `tools/pepsi/src/cargo.rs`.
Then i only have some minor suggenstions.

docs/tooling/behavior_simulator.md
schmidma: #2626 is now in the base branch. The current file contains the new Bevyhavior Simulator docs, so I would leave the old simulator docs out of this PR.

tools/twix/src/main.rs
schmidma: As discussed in person, we’ll defer this to a future PR. The current fallback is acceptable for this RoboCup slice.

docs/tooling/twix.md
schmidma: We’ll defer the larger docs cleanup until after RoboCup. The current PR keeps the docs focused on ROS-Z Twix; old Twix docs can be recovered from last year’s RoboCup tag when we revise the tooling docs.

tools/twix-legacy/src/main.rs
schmidma: This is fixed in the current branch. New Twix stores a per-tab `ui_id` and returns it from `TabViewer::id`; there is also a regression test for duplicate titles producing distinct dock IDs.

schmidma:
We’ll leave Pepsi unchanged in this PR. `twix-legacy` is deprecated, and `cargo run -p twix-legacy` is enough if someone needs it.

paudar (approved):
LGTM

PR #2658, opened 2026-06-30
Add small field location

BenSampaolo:
## Why? What?

Adds the small field measurements for the Incheon venue fields

## How to test

Set the location as location for the behavior simulator then run a scenario (merge-squash the behavior sim 2 branch)

### DIFF
--- a/etc/parameters/ros_z/incheon_small/global.json5
+++ b/etc/parameters/ros_z/incheon_small/global.json5
@@ -1,0 +1,20 @@
+{
+  "field_dimensions": {
+    "ball_radius": 0.065,
+    "length": 8.95,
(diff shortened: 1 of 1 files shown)


etc/parameters/ros_z/incheon_small/global.json5
knoellle, line 1: Wrong directory.
This is correct for old framework, methinks for new framework these parameters should be in `etc/parameters/ros_z/<location>/global.json5`

BenSampaolo: File moved and renamed

knoellle (approved):
LGTM

PR #2661, opened 2026-06-30
feat(twix): add ROS-Z parameter panel

schmidma:
## Summary
- add a ROS-Z-native Twix `Parameter` panel
- discover parameter-capable nodes from remote parameter snapshot services
- add snapshot-backed completion for parameter paths and active layers
- allow editing one parameter path as JSON and writing it to an active layer with an expected revision

## Motivation
This gives the team a working parameter-setting path in the new Twix without building the future tree/form inspector UI yet.

## Implementation
- exposes the shared Twix ROS-Z node and graph through `RobotBackend`
- adds a dockable `Parameter` panel with node/path/layer fields, refresh, set, JSON editor, metadata, and inline status
- uses `RemoteParameterClient` for snapshot/value/set calls with bounded async tasks so egui does not block on remote operations
- derives node completions from ROS-Z graph services and path completions from the latest parameter snapshot JSON
- ignores stale async results when the selected node, path, layer, or expected revision changes while requests are in flight
- disables parameter selectors, refresh, set, and value editing while remote operations are pending
- preserves selected inactive layers as errors instead of silently retargeting writes

## Verification
- `cargo fmt --check`
- `git diff --check hulks/main..HEAD`
- `cargo check -p twix`
- `cargo nextest run -p twix`

### DIFF
--- a/tools/twix/src/panels/parameter.rs
+++ b/tools/twix/src/panels/parameter.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, time::Duration};
+use std::{collections::BTreeSet, future::Future, time::Duration};
 
 use color_eyre::{
     Report,
@@ -10,17 +10,18 @@ use ros_z::{
     graph::Graph,
     parameter::{
         GetNodeParameterValueResponse, GetNodeParametersSnapshotResponse, RemoteParameterClient,
+        SetNodeParameterResponse,
     },
 };
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
 
 use crate::panel::{Panel, PanelCreationContext, PanelUiContext};
 
 const PARAMETER_SNAPSHOT_SUFFIX: &str = "/parameter/get_snapshot";
 const REMOTE_PARAMETER_TIMEOUT: Duration = Duration::from_secs(3);
 
-#[derive(Default)]
 pub struct ParameterPanel {
     node_editor: String,
     node: String,
@@ -33,6 +34,7 @@ pub struct ParameterPanel {
     effective_source_layer: String,
     value_editor: String,
     status: Status,
+    remote: RemoteState,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -40,6 +42,7 @@ struct SnapshotState {
     parameter_key: String,
     revision: u64,
     layers: Vec<String>,
+    path_completions: Vec<String>,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


tools/twix/src/panels/parameter.rs
```diff
@@ -0,0 +1,1998 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct SnapshotRequest {
+    node: String,
+    kind: SnapshotRequestKind,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[allow(dead_code)]
```
rmburg:
```suggestion
```

rmburg:
When clicking set, I get the following error (but it seems to set the parameter successfully). Refreshing does not help.

```
Parameter changed externally: editing revision 4, latest revision is 4. Refresh to discard local edits.
```

tools/twix/src/panels/parameter.rs
```diff
@@ -0,0 +1,1998 @@
...
+                    generation,
+                    node: node_fqn,
+                    revision,
+                });
+                egui_context.request_repaint();
+            }
+            let _ =
+                sender.send(RemoteOperationResult::EventSubscriptionFailed { generation, node });
+            egui_context.request_repaint();
+        });
```
rmburg: this task outlives the panel and selected node, receiving events and requesting repaints indefinitely. Maybe save the `JoinHandle` and cancel the task in `apply_node_change`, or add a `CancellationToken`.

tools/twix/src/panels/parameter.rs
```diff
@@ -0,0 +1,1998 @@
...
+            if self.is_value_dirty() {
+                self.remote.refresh_snapshot_metadata_after_pending = true;
+                self.mark_dirty_editor_stale(revision);
+            } else {
+                self.remote.refresh_snapshot_after_pending = true;
+            }
+            return RemoteFollowUp::None;
+        }
+        if self.is_value_dirty() {
+            self.mark_dirty_editor_stale(revision);
```
rmburg: maybe only mark as stale if the revision is strictly newer than the base revision? this could cause the stale warnings mentioned earlier

tools/twix/src/panels/parameter.rs
```diff
@@ -0,0 +1,1998 @@
...
+    fn apply_snapshot_response(
+        &mut self,
+        response: GetNodeParametersSnapshotResponse,
+    ) -> Result<(), Report> {
+        if !response.success {
+            return Err(eyre!("{}", response.message));
+        }
+
+        let path_completions = parse_snapshot_path_completions(&response.value_json)?;
+        let rendered_snapshot = render_json_for_editor(&response.value_json)?;
```
rmburg: this is always pretty-printed, but only used if `self.layer.is_empty()`. Maybe move the call into the is_empty branch?

rmburg:
Code LGTM, I'll test once more when I have access to a robot.

tools/twix/src/panels/parameter.rs
schmidma: Fixed in 50e3beba9 and refined in 82de8c7c5. The panel now stores the subscription task in `tokio_util::task::AbortOnDropHandle`, so node changes and current subscription termination abort the old receiver task.

tools/twix/src/panels/parameter.rs
schmidma: Fixed in 50e3beba9.

tools/twix/src/panels/parameter.rs
schmidma: Fixed in 50e3beba9. Stale detection now uses `latest_revision > base_revision`, with the fetched value revision preferred and the snapshot revision as fallback.

tools/twix/src/panels/parameter.rs
schmidma: Fixed in 50e3beba9. `render_json_for_editor` now only runs when the empty path view needs the full snapshot.

schmidma:
Fixed in 50e3beba9. The event handler now compares incoming revisions against the editor's base revision and only marks the editor stale when the incoming revision is strictly newer. Added regression coverage for equal value and snapshot revisions.

rmburg (approved):
LGTM

PR #2665, opened 2026-06-30
Parameter locations

knoellle:
## Why? What?

Adjusts `pepsi location` and related commands to properly handle new framework parameter paths.
Also sets the default location to `incheon_small`

- Based on #2663 

## How to Test

- `pepsi location <status/set/get/list>` should work again
- `pepsi pregame` should set location from deploy toml
- `hulk_ros_z` should load the parameters from the default location.


BenSampaolo (approved):
LGTM

PR #2668, opened 2026-07-01
Smarter placement in own cell

### LINKED ISSUE #2474: Dynamic Robot Placement

JackyBloxx:
## Why? What?

Adds smarter supporter positioning based on the Voronoi cell instead of walking directly to the cell centroid.

The new target selection scores candidate positions inside the robot's Voronoi cell using:
- distance to the ball,
- forward field position,
- distance to the Voronoi centroid.

This should place supporters in more useful positions around the ball while still keeping them inside their assigned Voronoi area.

- Fixes #2474 

## ToDo / Known Issues

 Tune the scoring weights for the positioning of the robot.

## How to Test

- Run behavior with several robots and verify that supporters choose targets inside their Voronoi cells.
- Move the ball and verify that supporter targets react to the ball position.
- Test no-ball behavior and verify that supporters fall back to centroid-like positioning.


crates/nodes/behavior_node/src/walk.rs
```diff
@@ -283,3 +288,88 @@ pub fn walk_to_centroid(blackboard: &mut Blackboard) -> Status {
...
+    let centroid_sigma = map
+        .parameters
+        .centroid_anchor_sigma
+        .max(map.parameters.grid_resolution);
+
+    let mut best_target = None;
+
+    for point in candidates {
+        let forward_norm = point.x() / half_length;
+        let forward_term = map.parameters.forward_weight * side_factor * forward_norm;
```
paudar, line 353: I'm confused by the name. 
As far as i see, the score of a cell left of the middle line is increased, if the ball is right of the line. 
Do you want to go closer to the ball or further away?

crates/nodes/behavior_node/src/walk.rs
```diff
@@ -283,3 +288,88 @@ pub fn walk_to_centroid(blackboard: &mut Blackboard) -> Status {
...
+    if count == 0 {
+        return None;
+    }
+
+    let inv_count = 1.0 / count as f32;
+    let centroid: Point2<Field> = point![sum_x * inv_count, sum_y * inv_count];
+
+    let Some(ball_position) = ball_position else {
+        return Some(centroid);
+    };
```
paudar, line 326: Seems logical, just be aware, that this is unlikely to be executed for now, since behavior uses the `searcher_subtree` in that case. (`tree.rs` line 81)

tools/twix-legacy/src/panels/map/layers/voronoi_cell.rs
paudar, line 1: Just curious:  Do you still use `twix-legacy` ?

paudar (commented):
Has merge conflicts

Some small things are unclear to me, but over all looks good

tools/twix-legacy/src/panels/map/layers/voronoi_cell.rs
JackyBloxx: No

crates/nodes/behavior_node/src/walk.rs
JackyBloxx: is just in case if we want to use this action somewhere else so this case is handled

crates/nodes/behavior_node/src/walk.rs
JackyBloxx: i understand why the name is confusing. The goal of this term is to be more agressive forward if the ball is in the opponent half and pulls you back if the ball is on our own side.

so is a term that says how much you want to be forward

paudar: Ok 👍

paudar (approved):
LGTM

PR #2669, opened 2026-07-01
Simulator gameplay fixes

BenSampaolo:
## Why? What?

This PR fixes multiple issues:

1. Adds collision between robots and robots and between robots and balls. This is a naive circle collision implementation which could be improved in the future.
2. The visual kick required the simulated robot to have an exact distance to the ball when kicking. Now for a kick its only required that the ball is not too far away, smaller distances are allowed 
3. Kick in and other variants like goal kick and corner kick are also covered by the auto ref now. 

## Issues

Leaving the field is not handled by the auto ref. This pr also clamps the robot position to the field. Alternatively this could be also an invariant checked by the sim.

## How to Test

Add random ball impulses to the 3v3 scenario and watch the game. The random impulses help resolve stalemates where robots repeatedly pass the ball back and forth when both just try to shoot at the other goal.


crates/bevyhavior_simulator/src/kinematics.rs
```diff
@@ -193,67 +338,89 @@ fn apply_walk_with_velocity_to_pose<Frame>(
...
         ball_position,
         kick_direction,
         kick_power,
+    ) {
+        KickAttempt::Kicked | KickAttempt::CoolingDown => return,
+        KickAttempt::NotInRange => {}
+    }
+
+    let standoff_distance =
+        (config.kick_radius * 0.9).min(ball_position.coords().dot(&kick_direction_vector).max(0.0));
```
knoellle: As discussed, clamp to something > 0.0 there

crates/bevyhavior_simulator/src/kinematics.rs
```diff
@@ -95,6 +147,99 @@ pub fn move_robots(
...
+                let overlap = minimum_distance - distance;
+                robot_positions[i].2 -= normal * (overlap / 2.0);
+                robot_positions[j].2 += normal * (overlap / 2.0);
+            }
+        }
+        clamp_robot_positions_to_field(robot_positions, robot_radius, field_dimensions);
+    }
+}
+
+fn clamp_robot_positions_to_field(
```
knoellle: I don't think the simulator should physically clamp the robot positions to the field area.
If anything, this should be an invariant check, but I think that's something for later. I'd rather get other things done first.

crates/bevyhavior_simulator/src/ball.rs
```diff
@@ -2,21 +2,26 @@ use std::time::SystemTime;
...
+    config::SimulationConfig,
     coordinates::point_world_to_field,
+    game_controller::SimulatorGameState,
+    robot::{SimulatorGroundToWorld, SimulatorRobot, SimulatorRobotId},
 };
 
 #[derive(Resource, Clone, Copy, Debug, Default)]
 pub struct SimulatorBall {
     pub state: Option<SimulatedBall>,
+    pub last_touch_team: Option<Team>,
```
knoellle:
```suggestion
    pub last_touched_by: Option<Team>,
```

crates/bevyhavior_simulator/src/ball.rs
```diff
@@ -56,3 +61,49 @@ pub fn move_ball(
     ball.position += ball.velocity * dt;
     ball.velocity *= (1.0 - config.ball_friction_per_second * dt).clamp(0.0, 1.0);
 }
+
+pub fn update_ball_last_touch_from_robot_contacts(
```
knoellle: Why is this separate from the collision detection?
Just set the last touched by field of the ball every time you process a robot-ball collision.

crates/bevyhavior_simulator/src/auto_referee.rs
```diff
@@ -283,6 +297,56 @@ impl Default for GameStateTransitionRule {
...
+            velocity: Vector2::zeros(),
+            field_side: ball.field_side,
+        });
+        context.ball.last_touch_team = None;
+        context.set_kicking_team(Some(kicking_team));
+        context.set_sub_state(Some(sub_state));
+    }
+}
+
+impl Default for BallOutOfFieldRule {
```
knoellle: Derive this

BenSampaolo: Done

crates/bevyhavior_simulator/src/ball.rs
BenSampaolo: Done

crates/bevyhavior_simulator/src/ball.rs
BenSampaolo: Done

crates/bevyhavior_simulator/src/kinematics.rs
BenSampaolo: That's right, done

crates/bevyhavior_simulator/src/kinematics.rs
BenSampaolo: Now robot radius plus ball radius, done

knoellle (approved):
LGTM

PR #2672, opened 2026-07-01
Twix image overlays

alexschmander:
## Why? What?

Adds twix image overlays back to the new rosz native twix. Inlcudes all functional overlays from the old twix except, feet detection, since we dont have that anymore. 

## How to Test

Open twix, open image panel with new overlays.


paudar (commented):
It seems we might have another Ctrl+C, Ctrl+V error in Main. In `crates/nodes/line_detection/src/lib.rs`, line 70, two publishers are publishing to the same topic name, which could cause problems.

The documentation is out of date, but will be updated after the event.

paudar:
> It seems we might have another Ctrl+C, Ctrl+V error in Main. In `crates/nodes/line_detection/src/lib.rs`, line 70, two publishers are publishing to the same topic name, which could cause problems.

addressed in #2691

tools/twix/src/panels/image/mod.rs
```diff
@@ -18,6 +18,11 @@ use crate::{
     status::format_topic_observation_status,
 };
 
+use self::overlay::{ImageOverlayPainter, ImageOverlays};
+
+mod overlay;
+mod overlays;
```
paudar: This looks a bit confusing. Can overlay.rs be renamed, so that the name includes, that it is part of the image panel?

paudar (commented):
one small comment, otherwise looks good

paudar (approved):
LGTM

PR #2678, opened 2026-07-01
Don't fetch for game branch

knoellle:
## Why? What?

`pepsi gamebranch` automatically fetches all relevant remotes. I don't like that.

## How to Test

run `pepsi gamebranch` without internet.
wouldn't work previously, now does.


alexschmander (approved):
LGTM

PR #2679, opened 2026-07-01
3v3 speedup

knoellle:
## Why? What?

3v3 scenario currently takes about a minute of wall time to simulate because they bounce the ball back and fourth a lot.
- remaining time in half was incorrectly set, causing robots to send messages at max speed
- once message budget runs out, only robots which see the ball participate in driving game forward
- few robots participating = more bouncing ball back and fourth.
- Instead of first goal, simulation ran until one team has at least two goals.

This PR stops scenario after first goal and implements remaining time in half tracking.

Based on #2669 

## How to Test

`cargo run --bin three_vs_three_until_goal_or_half`

### DIFF
--- a/crates/bevyhavior_simulator/src/auto_referee.rs
+++ b/crates/bevyhavior_simulator/src/auto_referee.rs
@@ -425,7 +425,23 @@ pub fn run_auto_referee(
         rule.apply(&mut context);
     }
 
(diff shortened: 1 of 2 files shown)


BenSampaolo (dismissed):
Looks good

BenSampaolo (approved):
Looks still good

PR #2683, opened 2026-07-01
Migrate playernumber

knoellle:
## Why? What?

Updates pepsi/repo crate to write playernumbers in the new configuration location / format.

## How to Test

`./pepsi playernumber 44:4`

PR #2684, opened 2026-07-01
Pregame cleanup

knoellle:
## Why? What?

Removes old communication disabling and recording rate setting from pepsi

- Based on #2678 
- Based on #2683


paudar (approved):
Worked at RoboCup26

PR #2686, opened 2026-07-01
avoid crash when alsa device busy

vivi-enne:
## Why? What?

re-enable microphone crate and avoid crash when alsa device busy, logs the error and not use the microphone

- Fixes #

## ToDo / Known Issues

## How to Test

Test with changing the hardware_device_name or when robot crashes again.


knoellle (approved):
LGMT

PR #2690, opened 2026-07-01
Use only one version of RustDDS

rmburg:
## Why? What?

What the title says. We depended on RustDDS twice.
Might fix #2404

## ToDo / Known Issues

- [ ] test

none

## How to Test

check if DDS things still work on the robot


oleflb:
We no longer use `rustdds`

PR #2691, opened 2026-07-01
Fix line_data topic name

paudar:
## Why? What?

As reviewing #2672, i saw, that there was a wrong topic name for the `line_data`-publisher in the linedetection.
Was renamed

"Immer alles richtig machen"

## How to Test

is a test requred?


pejotejo (approved):
lgtm

PR #2692, opened 2026-07-01
Port team ball

JackyBloxx:
## Why? What?

Ports the team ball filter to have team_balls again
- Fixes #

## How to Test

Check if Robot_A publishes a team_ball when Robot_B sees the Ball.
Best checked if both are in playing, on the same network and have different Playernumbers.


crates/nodes/ball_state_composer/src/lib.rs
```diff
@@ -108,7 +108,11 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
                 });
             }
             received_team_ball = team_ball_sub.recv() => {
-                let team_ball = received_team_ball?;
+                let Some(team_ball) = received_team_ball? else {
+                    ball_state_pub.publish(&None).await?;
+                    last_ball_state = None;
```
alexschmander: When receiving a `None` `team_ball`, this currently resets the entire `last_ball_state`, including a valid locally detected ball.

crates/nodes/team_ball_filter/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    let player_states_sub = node
+        .subscriber::<Players<Option<TimeWrapper<PlayerState>>>>("player_states")
+        .build()
+        .await?;
+    let team_balls_pub = node
+        .publisher::<Players<Option<BallPosition<Field>>>>("team_balls")
+        .build()
+        .await?;
+    let team_ball_pub = node
+        .publisher::<Option<BallPosition<Field>>>("team_ball")
```
alexschmander, line 42: `tools/twix/src/panels/map/layers/ball_position.rs` still uses the old `BallPosition<Field>` type. This needs updating to the `Option`

crates/nodes/team_ball_filter/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn filtered_received_balls(
+        &self,
+        now: ros_z::time::Time,
+        maximum_age: Duration,
+    ) -> Players<Option<BallPosition<Field>>> {
+        self.received_balls
+            .map(|ball| ball.filter(|ball| ball.age_at(now).is_some_and(|age| age < maximum_age)))
+    }
+
+    fn get_best_received_ball(
```
alexschmander: Doesn't this function get the latest ball? I think we could name the function accordingly right?

crates/nodes/team_ball_filter/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        let now = node.clock().now();
+        if filtered_game_controller_state
+            .as_ref()
+            .is_some_and(is_penalty_phase_or_sub_state)
+        {
+            team_ball_pub.publish(&None).await?;
+            continue;
+        }
+
+        let team_ball = team_ball_filter.get_best_received_ball(now, parameters.maximum_age);
```
alexschmander: This age check only applies when actually receiving a filtered game controller state or player states. 
The team ball is therefore not expired, until such a input is received. This is more or less only minor, but will lead to team balls being viewed externally as valid for longer than the parameter specifies. To fix this, you could add a rosz timer on a 100ms or similar and receive in the tokio select on that.

etc/parameters/base/team_ball_filter.json5
alexschmander, line 1: You added this file, but did not delete the old `team_ball_receiver.json5`

Cargo.toml
```diff
@@ -279,7 +279,8 @@ splines = { version = "=4.2.0", features = ["serde"] }
 stereo_visual_odometry = { path = "crates/nodes/stereo_visual_odometry" }
 support_foot_estimator = { path = "crates/nodes/support_foot_estimator" }
 syn = { version = "2.0.98", features = ["extra-traits", "full"] }
-team_ball_receiver = { path = "crates/nodes/team_ball_receiver" }
+systemd = "0.10.0"
```
alexschmander: Why did you add this?

alexschmander (changes requested):
Some things left :)

crates/nodes/ball_state_composer/src/lib.rs
JackyBloxx: done

crates/nodes/team_ball_filter/src/lib.rs
JackyBloxx: done

crates/nodes/team_ball_filter/src/lib.rs
JackyBloxx: i renamed it

crates/nodes/team_ball_filter/src/lib.rs
JackyBloxx: true i now added a timer

etc/parameters/base/team_ball_filter.json5
JackyBloxx: done

Cargo.toml
JackyBloxx: i dont know i deleted it

crates/nodes/ball_state_composer/src/lib.rs
alexschmander: This fixes the case where a None team ball clears a valid local ball, but the other direction still has the same problem right? A None from the local ball filter still clears a valid team ball.
Also, ignoring the team ball's None means its expiry no longer clears a previously selected team ball.
Could we keep the local and team ball separately and recompute the ball state when either changes? That way, None only clears its own source, and we publish None when neither has a valid ball.

alexschmander (approved):

PR #2694, opened 2026-07-01
Add no message behavior

JackyBloxx:
## Why? What?

removes stale inputs in player_state_filter when we dont have any messages anymore.
and adds a condition when we have no messages anymore to get striker

## How to Test

test if we are out off messages if the robot the goes into that part of the tree and if the player_state is None for every player.


paudar:
@JackyBloxx, is this still relevant?

JackyBloxx:
not relevant

PR #2695, opened 2026-07-01
feat: reduce HSL state messages

pejotejo:
## Why? What?
only send messages in playing and when we or the ball are moving, also make messagecount dynamically and set default messageintervall to 0.5s with 3 robots

## How to Test

look at the messages

### DIFF
--- a/etc/parameters/ros_z/base/behavior_node.json5
+++ b/etc/parameters/ros_z/base/behavior_node.json5
@@ -156,6 +156,16 @@
       secs: 1,
     },
     remaining_amount_of_messages_to_stop_sending: 20,
+    message_budget: 1200,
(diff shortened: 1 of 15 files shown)


etc/parameters/ros_z/base/behavior_node.json5
JackyBloxx, line 1: Why defineyou parmeters in behavior_node.json5 and in role_assignment.json5
and also you set our message_budget to 1200 and also why define a message budget when you could get that info from the game controller

etc/parameters/ros_z/base/role_assignment.json5
JackyBloxx, line 1: her also

PR #2696, opened 2026-07-01
more messages

JackyBloxx:
## Why? What?

make message interval faster  6000 per halftime 2000 per robot,
2000 / 600s (Playtime) = 3.333 Hz or every 300000000ns

we have spare messages because we only sen in playing

## How to Test

test if we send messages in this interval


BenSampaolo (approved):
LGTM

PR #2701, opened 2026-07-02
Add ROS-Z observation policy

schmidma:
## Summary

Adds observer-side policy configuration for ROS-Z debugging.

## Motivation

Local subscriber queue capacity was previously coupled to QoS history depth, which caused observation workloads using default `KeepLast(10)` QoS to emit queue overflow warnings under load. Observation code now has explicit knobs for local queue capacity and overflow reporting without changing advertised QoS by default.

## Implementation

- Adds `ros_z::pubsub::QueueOverflowReporting`.
- Adds local subscriber `queue_capacity(...)` and `queue_overflow_reporting(...)` builder knobs across typed, raw, and dynamic subscriber builders.
- Adds `ros_z_debug::ObservationPolicy` for retention, optional subscriber QoS, optional local queue capacity, overflow reporting, and update buffer capacity.
- Threads `ObservationPolicy` through cached subscriptions and topic observations while preserving existing defaults.

## Verification

- `cargo fmt --check`
- `cargo nextest run -p ros-z --test pubsub`
- `cargo check -p twix`
- `git diff --check`

### DIFF
--- a/crates/ros-z-debug/src/observation.rs
+++ b/crates/ros-z-debug/src/observation.rs
@@ -1866,7 +1866,7 @@ fn send_observation_update<T>(
 
 #[cfg(test)]
 mod tests {
-    use std::{sync::Arc, time::Duration};
+    use std::{num::NonZeroUsize, sync::Arc, time::Duration};
 
     use parking_lot::Mutex;
     use ros_z::{prelude::*, time::Time};
@@ -2241,6 +2241,7 @@ mod tests {
         let cache_state = Arc::new(CachedSubscriptionState::<String>::new(
             fresh_cache,
             RetentionPolicy::LatestOnly,
+            NonZeroUsize::new(256).unwrap(),
         ));
         let (observation_update_sender, observation_update_receiver) =
             broadcast::channel(super::UPDATE_BUFFER_CAPACITY);
@@ -2307,6 +2308,7 @@ mod tests {
         let cache_state = Arc::new(CachedSubscriptionState::<String>::new(
             stale_cache.clone(),
             RetentionPolicy::LatestOnly,
+            NonZeroUsize::new(256).unwrap(),
         ));
         let cache = cache_state.handle();
         let (observation_update_sender, observation_update_receiver) =
@@ -2353,6 +2355,7 @@ mod tests {
         let cache_state = Arc::new(CachedSubscriptionState::<String>::new(
             CachedSubscriptionStatusSnapshot::new(CachedSubscriptionStatus::WaitingForFirstSample),
             RetentionPolicy::LatestOnly,
+            NonZeroUsize::new(256).unwrap(),
         ));
         cache_state.store_latest(string_sample_record("ready before install"));
         let cache = cache_state.handle();
@@ -2400,6 +2403,7 @@ mod tests {
         let cache_state = Arc::new(CachedSubscriptionState::<String>::new(
             CachedSubscriptionStatusSnapshot::new(CachedSubscriptionStatus::WaitingForFirstSample),
             RetentionPolicy::LatestOnly,
+            NonZeroUsize::new(256).unwrap(),
         ));
         let cache = cache_state.handle();
         let built_cache = cache.clone();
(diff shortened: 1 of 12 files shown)


crates/ros-z-debug/src/observation.rs
```diff
@@ -451,7 +463,6 @@ impl TopicObservationBuilder<DynamicPayload> {
         }
     }
 }
-
 /// Builder for a dynamic topic observation.
```
rmburg, line 467: nit: I like the newline between top-level items

crates/ros-z/src/pubsub/raw.rs
```diff
@@ -79,6 +79,26 @@ where
...
     }
 
+    /// Set the local raw subscriber receive queue capacity.
+    ///
+    /// This does not change advertised endpoint QoS. If unset, capacity is
+    /// derived from the effective QoS history depth.
+    pub fn queue_capacity(self, queue_capacity: std::num::NonZeroUsize) -> Self {
+        Self {
+            inner: self.inner.queue_capacity(queue_capacity),
+        }
```
rmburg, line 89: nit: this uses a different builder pattern style than the other functions in this PR. The others pass `mut self`, mutate `self`, and return `self` instead of constructing a new `Self`.

crates/ros-z/src/pubsub/subscriber.rs
```diff
@@ -21,18 +22,46 @@ use crate::queue::BoundedQueue;
...
+    ///
+    /// Dropped samples are still accounted internally.
+    Silent,
+    /// Emit a debug log message for each local queue overflow.
+    Debug,
+    /// Emit a warning log message for each local queue overflow. This is the default reporting mode.
+    #[default]
+    Warn,
+}
+
```
rmburg, line 57: This enum could be replaced by the more general [`tracing::Level`](https://docs.rs/tracing/latest/tracing/struct.Level.html). This is already used to specify logging levels and also includes `Level::OFF` to disable messages. Not sure which is better.

crates/ros-z/src/prelude.rs
```diff
@@ -44,3 +46,15 @@ pub use crate::{Message, SerdeCdrCodec, Service};
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn prelude_exports_queue_overflow_reporting() {
+        let reporting = QueueOverflowReporting::Silent;
+
+        assert_eq!(reporting, QueueOverflowReporting::Silent);
+    }
+}
```
rmburg: is this test necessary? all other items in this prelude don't have such a test.

rmburg (commented):
Looks good, just some small nits. Every comment with "nit:" is not blocking.

crates/ros-z-debug/src/observation.rs
schmidma: Fixed in `2a705c309`.

crates/ros-z/src/pubsub/raw.rs
schmidma: Leaving this as-is because `RawSubscriberBuilder` already uses this forwarding style for all methods in the impl, including `qos`, `locality`, and `transient_local_replay_timeout`. Changing only the new methods would make this file internally inconsistent.

crates/ros-z/src/pubsub/subscriber.rs
schmidma: Keeping the domain enum for now. `tracing::Level` does not include `OFF`; disabling is represented by `tracing::level_filters::LevelFilter::OFF`. Using `Level` or `LevelFilter` would also expose `TRACE`, `INFO`, and `ERROR`, but this API only supports the three intended overflow reporting modes: silent, debug, and warn.

crates/ros-z/src/prelude.rs
schmidma: Removed in `2a705c309`.

rmburg (approved):
abfahrt

PR #2712, opened 2026-07-02
striker when alone

JackyBloxx:
## Why? What?

Add the condition when we are alone on the field that we are striker.

## How to Test

test if the robot do Striker behavior when alone on field.

Is testable with behavior simulator : you need to change the playing tree that when we are alone do striker otherwiese do nothing


paudar (approved):
LGTM

PR #2717, opened 2026-07-02
Localize game branches

knoellle:
## Why? What?

Currently game branches use the current date in UTC.
This isn't very intuitive when in timezones far from UTC since we usually think in Spieltagen.
This PR changes it to use the local timezone.

## How to Test

Run
 ```
cp deploy.toml.example deploy.toml
./pepsi gamebranch
```
when local date is different from date at UTC+0

Then look at branch name to confirm matches local date.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7661,7 +7661,7 @@ dependencies = [
 
 [[package]]
 name = "pepsi"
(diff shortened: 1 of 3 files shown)


paudar (approved):
LGTM

PR #2721, opened 2026-07-02
Remove obsolete vision runtime stack

schmidma:
## Summary
- remove the old generated vision runtime stack and ROS-Z producer nodes
- keep raw camera image topics for modern detection
- move detection inference and search-suggestor heatmap work to `spawn_blocking`

## Motivation
The old vision producers are no longer used by the ROS-Z runtime, but their topics, parameters, and Twix overlays still kept stale paths alive. Modern detection still needs raw camera topics.

## Implementation
- remove old field-border, image-segmentation, line-detection, segment-filter, stereo-pair runtime code
- remove stale parameter files and new Twix overlays for deleted topics
- preserve `tools/twix-legacy` as migration backup
- wrap ONNX session creation/inference and search heatmap updates in Tokio blocking tasks

## Verification
- `cargo check -p hulk_ros_z`
- `cargo check -p detection --features ort-dynamic`
- `cargo check -p search_suggestor`
- `cargo check -p twix`
- `cargo check -p twix-legacy`
- `cargo nextest run -p detection check_image`
- `cargo nextest run -p twix save_preserves_topic`
- `cargo nextest run -p hulk_ros_z`
- `cargo fmt --check`


crates/nodes/detection/src/lib.rs
```diff
@@ -158,17 +215,23 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
...
+    };
+
+    Ok(DetectionOutput {
+        session,
+        detected_objects,
+        detected_poses,
+        inference_duration,
+        post_processing_duration,
+        non_maximum_suppression_duration,
+    })
```
oleflb, line 220: This looks weird. Is this intended?

crates/nodes/image_receiver/src/lib.rs
```diff
@@ -41,20 +38,10 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
         })
         .build()
         .await?;
-    let ycbcr422_image_pub = node
-        .publisher::<TimeWrapper<YCbCr422Image>>("inputs/ycbcr422_image")
-        .build()
-        .await?;
-    let stereo_image_pair_pub = node
```
oleflb, line 48: we still need this

crates/nodes/search_suggestor/src/lib.rs
```diff
@@ -104,102 +121,52 @@ async fn run(ctx: Arc<Context>) -> Result<()> {
...
-        let parameters = parameters_snapshot.typed();
+        let parameters = parameters_snapshot.typed().clone();
 
         let ground_to_field = ground_to_field_cache
             .get_latest()
             .map(|ground_to_field| *ground_to_field);
-        let primary_state = primary_state_cache.get_latest();
-        let primary_state = primary_state.as_deref();
-        let mut ball_was_seen = false;
+        let primary_state = primary_state_cache.get_latest().map(|state| *state);
```
oleflb: copied/cloned?

oleflb (changes requested):

crates/nodes/detection/src/lib.rs
schmidma: Fixed. The helper now constructs `DetectionOutput` directly instead of returning a tuple and immediately reconstructing the struct. The explicit `drop(model_outputs)` / `drop(session_outputs)` calls keep the session borrow scoped before moving `session` back into the output.

crates/nodes/image_receiver/src/lib.rs
schmidma: Restored `inputs/stereo_image_pair` publishing and the `StereoImagePair` type. `inputs/ycbcr422_image` stays removed, since only the stereo pair is still needed for the future localization path.

crates/nodes/search_suggestor/src/lib.rs
schmidma: Changed this to `.as_deref().copied()` for both cached values, so the copy from the cached `Arc<T>` is explicit.

schmidma:
Decided to wait for old stuff deletion

Issue #2389, opened 2026-03-13
post-game task: Reintegrate ball search


BenSampaolo:
What the title says

PR #2723, opened 2026-07-02
Ball search behavior

BenSampaolo:
## Why? What?

Issue #2389

This PR adds a smarter searching behavior which uses a new heat map. While the old heat map was a probability distribution of where the ball could be the new one is a priority map which area to check when searching for the ball.
This PR helps to better search the field and make the robot actually move around when looking for the ball. 

## ToDo / Known Issues

This PR needs the team messages to communicate and search more efficiently and therefore will work better when #2692 is merged.

## How to Test

Deploy on the robot and enter playing mode. Make sure the robot doesn't see a ball and has loca. The robot should circle around the field and look for the ball.


BenSampaolo:
~~Cells with obstacles should be able to be chosen by the search suggester~~

BenSampaolo:
~~Parameters are wired badly (Use own parameters instead of behavior parameters in search suggester node)~~

BenSampaolo:
~~Faster head motions
Probably more tuning, especially faster cell regeneration with higher hysteresis and bigger FOV~~

etc/parameters/ros_z/base/search_suggestor.json5
```diff
@@ -1,15 +1,25 @@
...
-  minimum_validity: 0.01,
-  own_ball_weight: 1.0,
-  team_ball_weight: 1.0,
-  rule_ball_weight: 1.0,
-  rule_ball_weight_increment: 0.005,
-  decay_distance_factor: 0.03,
-  heatmap_decay_range: {
-    start: 0.0,
-    end: 3.0
+  search_suggestor: {
```
rmburg, line 2: un-nest this (remove the wrapping `search_suggestor: {...}`)

crates/nodes/search_suggestor/src/lib.rs
```diff
@@ -33,7 +31,7 @@ pub fn run_boxed(ctx: Arc<Context>) -> Pin<Box<dyn Future<Output = Result<()>> +
 async fn run(ctx: Arc<Context>) -> Result<()> {
     let node = ctx.create_node("search_suggestor").build().await?;
 
-    let parameters = node.bind_parameter_as::<SearchSuggestorParameters>("search_suggestor")?;
+    let node_parameters = node.bind_parameter_as::<Parameters>("search_suggestor")?;
```
rmburg, line 34: undo this change to un-nest the parameters in `search_suggestor.json5`

crates/nodes/behavior_node/src/node.rs
```diff
@@ -409,7 +424,11 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
             outgoing_message_pub.publish(&message).await?;
         }
 
-        if let Some(message) = blackboard.try_sending_state_message() {
+        let head_yaw = head_joints_command_cache
+            .get_latest()
+            .map(|head_joints| head_joints.yaw)
+            .unwrap_or_default();
+        if let Some(message) = blackboard.try_sending_state_message(head_yaw) {
```
rmburg, line 431: can you put `head_yaw` into the blackboard like the other subscriptions of the behavior node instead of passing it into this function?

PR #2725, opened 2026-07-02
Out with the old; In with the new!

knoellle:
## Why? What?

Removes the `old` flag from gammaray and upload which allowed uploading and running the now removed old framework hulk binary.

## How to Test

Everything for the new framework should continue working as expected but gammaray and upload no longer accept `--old`.

### DIFF
--- a/tools/k1-setup/launch-hulk
+++ b/tools/k1-setup/launch-hulk
@@ -14,7 +14,7 @@ chown -R booster:booster /home/booster/hulk/logs
   --user $(id -u booster) \
   hulk \
   /home/booster/hulk/bin/hulk_ros_z \
-  --robot {} \
+  --robot ROBOT_NUMBER \
   --location default-location \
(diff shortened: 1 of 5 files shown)


tools/k1-setup/launch-hulk
```diff
@@ -13,6 +13,11 @@ chown -R booster:booster /home/booster/hulk/logs
   --env RUST_BACKTRACE=1 \
   --user $(id -u booster) \
   hulk \
-  /home/booster/hulk/bin/hulk_booster --log-path "${INTERNAL_LOG_PATH}" \
+  /home/booster/hulk/bin/hulk_ros_z \
+  --robot {} \
```
rmburg:
```suggestion
  --robot ROBOT_NUMBER \
```

rmburg (changes requested):

rmburg (dismissed):
continvoucly morge

rmburg (approved):

PR #2729, opened 2026-07-02
change kick parameters and only use rumpelstilzchen

pejotejo:
## Why? What?
what the title says

## How to Test

kick a bit, maybe with loca


paudar (approved):
Looks good as parameter for the next game.

PR #2730, opened 2026-07-02
Watch pings

knoellle:
## Why? What?

Allows `pepsi ping --watch / -w` which repeatedly pings each robot independently.
This is better than `watch pepsi ping` because each robot has its own independent timeout instead of barrier-syncing on pepsi exiting.

## How to Test

`./pepsi stups -w 42 43`


paudar (approved):
LGTM

PR #2738, opened 2026-07-03
Change Message sending

JackyBloxx:
## Why? What?

Only sending messages in Playing and when we have a ground_to_field

## How to Test

look in twix if we publish any hulk_messages in any other state then Playing


pejotejo (approved):
legtm

PR #2750, opened 2026-07-03
Remove old playerstates

### LINKED ISSUE #2737: let voronoi forgett information
Let voronoi forget knowledge of other players also if they don't get penalized (e.g. after time)

JackyBloxx:
## Why? What?

remove old player_state after
- Fixes #2737 

## How to Test

take two robots and both in playing in the same wifi.
when one robot turns off or dosnt send any message anymore we should forget the state from them


pejotejo (approved):
logotomie

PR #2751, opened 2026-07-03
Always clock the jetsons

### LINKED ISSUE #2710: Investigate how to handel Jetson Clock

knoellle:
## Why? What?

Adds a systemd unit which repeatedly runs `jetson_clocks`.
Also includes a `jetson_clocks --show` ExecPre to allow us to determine where/when the clock speeds break again.

- Fixes #2710

## How to Test

`pepsi gammastrahl`


pejotejo (approved):
works i guess

PR #2752, opened 2026-07-03
Only consider cells in centroid bounds

JackyBloxx:
## Why? What?

Only consider VOronoi cells inside the field so no support should walk out of the field
- Fixes #


pejotejo (approved):
logotomie

PR #2754, opened 2026-07-03
Leuchtturm is leuchtturming again

Franziska-Sophie:
## Why? What?

When the robot is in Playing and does not see the ball it is supposed to do the Leuchtturm, i.e., turn around on its spot, to find the ball. This behavior did not work and this PR fixes this.

- Fixes #

## ToDo / Known Issues

-

- increase the turn speed (but it seems to be clamped or the walk policy can’t turn faster)

## How to Test

Upload on the robot, go into primary state Playing, make sure the robot sees no ball. It should start turning now.


pejotejo (approved):
lgtm

PR #2764, opened 2026-07-03
Recover low state source timestamps

schmidma:
## Summary
- recover `/low_state` source times by matching low-state motor values to `/joint_states`
- publish low-state-derived ROS-Z topics with the matched vendor timestamp
- allow `joint_states` through the K1 Zenoh DDS bridge config

## Motivation
`/low_state` has no embedded timestamp, so the bridge previously tagged it with receive/republish time. Matching to `/joint_states.header.stamp` removes receive-path jitter for IMU and motor-state consumers.

## Implementation
- add `Publisher::publish_with_source_time`
- add `ros2::sensor_msgs::joint_state::JointState`
- buffer recent low-state and joint-state samples in `low_state_bridge`
- match exact f32 bit keys for position, velocity, and effort/torque
- drop stale or non-monotonic matches instead of publishing guessed timestamps

## Verification
- `cargo nextest run -p ros-z -p low_state_bridge`
- `cargo build -p ros2 -p low_state_bridge -p hulk_ros_z`
- `cargo fmt --check`

### DIFF
--- a/docs/setup/upload.md
+++ b/docs/setup/upload.md
@@ -13,6 +13,11 @@ pepsi upload <NAO>
 Pepsi takes care of downloading and installing the SDK, calling `cargo` to trigger compilation, and uploading the software to the robot.
 When successful, you are presented with a green tick in your shell, and the robot is showing rotating rainbow eyes.
 
+## Runtime Topics
+
+The `low_state_bridge` process publishes low-state-derived inputs only after it can match `rt/low_state` samples with timestamped `rt/joint_states` samples.
+If these inputs stop updating after deployment, verify that both topics are produced and that `joint_states` is allowed through the DDS bridge configuration.
+
 !!! tip
 
     Continue reading [here](../workflow/overview.md) about our workflow and how to contribute to the project.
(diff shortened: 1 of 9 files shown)


docs/setup/upload.md
```diff
@@ -13,6 +13,11 @@ pepsi upload <NAO>
 Pepsi takes care of downloading and installing the SDK, calling `cargo` to trigger compilation, and uploading the software to the robot.
 When successful, you are presented with a green tick in your shell, and the robot is showing rotating rainbow eyes.
 
+## Runtime Topics
+
+The `low_state_bridge` process publishes low-state-derived inputs only after it can match `rt/low_state` samples with timestamped `rt/joint_states` samples.
+If these inputs stop updating after deployment, verify that both topics are produced and that `joint_states` is allowed through the DDS bridge configuration.
```
alexschmander: Why did you add this doc section? It seem out of place.

alexschmander (changes requested):

docs/setup/upload.md
schmidma: Removed the section. It was meant as an operational note for the new topic dependency, but this upload page is not the right place for it.

alexschmander (approved):
LGTM

PR #2768, opened 2026-07-03
change order penalized stop primary state

### LINKED ISSUE #2760: penalized > stop
At the moment when we are penalized in stop we don't go into primary state penalized, but rather stay in stop until it is over. We should go directly to penalized.

pejotejo:
## Why? What?
fixes #2760


paudar (approved):
LGTM

PR #2774, opened 2026-07-04
Ignore balls outside the field

### LINKED ISSUE #2756: remove balls too far outside the field
We need to remove balls outside the field.  We should still trust balls 1-2m outside the field because ouf kicks and loca jitter

pejotejo:
## Why? What?
What the title says, but only if the robot is connected to a gamecontroller
fixes #2756

### DIFF
--- a/crates/nodes/behavior_node/src/node.rs
+++ b/crates/nodes/behavior_node/src/node.rs
@@ -376,8 +376,7 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
         {
             let length = blackboard.field_dimensions.length;
             let width = blackboard.field_dimensions.width;
-            let border = blackboard.field_dimensions.border_strip_width
-                - blackboard.parameters.border_strip_ignore_ball_zone;
+            let border = blackboard.parameters.believe_ball_border_width;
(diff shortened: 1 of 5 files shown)


crates/nodes/behavior_node/src/node.rs
```diff
@@ -353,14 +353,28 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
+        if let Some(ball) = blackboard.world_state.ball
+            && blackboard
+                .world_state
+                .filtered_game_controller_state
+                .is_some()
+        {
+            let length = blackboard.field_dimensions.length;
+            let width = blackboard.field_dimensions.width;
+            let border = blackboard.field_dimensions.border_strip_width
+                - blackboard.parameters.border_strip_ignore_ball_zone;
```
paudar: For me, this calculation is to complicated. Please make it easier to avoid mistakes in stressful situations.
(line a parameter, that is added to the field lines)

pejotejo: done

paudar (approved):
LGTM

PR #2776, opened 2026-07-04
Slower walk to Kick_off

paudar:
## Why? What?

Since we overshoot heavily in walk in, i reduced the wak_speed to reach the pose in the ready phase. 

Also found outdated parameter fields and removed them.

## How to Test

Testgame


pejotejo (approved):
lgtm

PR #2780, opened 2026-07-04
Download journal with logs

knoellle:
## Why? What?

Downloads the whole systemd journal during `pspsps postgame` and `pspsps logs download`

## How to Test

- `./pepsi postgame`
- or `./pepsi logs download`
- then inspect log folder, should contain `system.log` with journal content.


pejotejo (approved):
logotomie

PR #2787, opened 2026-07-04
Fix localization pose projection

oleflb:
## Why? What?

Pose projection was broken since an inverse was missing. This PR fixes that.

Context: Pose projection projects the current interpolated solver result so it coincides with the robot_to_ground. This is currently required so camera extrinsic chains in non-loca3d nodes remain correct.

- Fixes #

## ToDo / Known Issues

## How to Test


rmburg (approved):
abfahrt

PR #2806, opened 2026-07-04
Add individual Goalkeeper Hysteresis

### LINKED ISSUE #2797: change walk hysteresis for the goal keeper
In active defence our goal keeper should be positioned more percise

paudar:
## Why? What?

The Goalkeeper does not update it's position often enough. Therefore it get's a own hysteresis.

- Fixes #2797

## How to Test

Testgame with at least two robots


BenSampaolo (approved):
LGTM

PR #2808, opened 2026-07-04
keeper uses own goal

schmidma:
oops... sign flip


BenSampaolo (approved):
Looks suspicious but I'll let it pass this time ;)

PR #2809, opened 2026-07-04
Remove kickoff obstacles

### LINKED ISSUE #2796: remove kickoff obstacles
Those are probably a relic from the nao era we dont need anymore.

JackyBloxx:
## Why? What?

Remove kickoff obstacles to get faster into our field and dont bound against the obstacle
- Fixes #2796

## How to Test

let the robot walk to the opponents half infront of the center circle and then let the robot walk back in ready


pejotejo (approved):
lgtm

PR #2815, opened 2026-07-06
remove some NAO-specific code/assets

### LINKED ISSUE #2142: rename NAO -> ?
Many things are called nao that could be named robot or K1 or something else. This should be addressed to avoid confusions

rmburg:
## Why? What?

what the title says

- Fixes #2142 partially (there are still some NAO-specific things in mujoco projects, I'm keeping them for now)

## ToDo / Known Issues

remove even more

remove even more

## How to Test

that's the neat part: you don't


knoellle (approved):
🧹

PR #2820, opened 2026-07-18
Bump egui + bevy versions

oleflb:
## Why? What?

Bumps bevy to 0.19.0 and egui to 0.35.0. This also requires to increase MSRV to `1.95.0`
Motivation is the new simulator, which will need features from bevy 0.19 

- Fixes #

## ToDo / Known Issues

## How to Test

- Check apps still behave as expected


tools/twix/src/configuration/keybind_plugin.rs
```diff
@@ -7,14 +7,14 @@ use super::keys::{KeybindAction, Keybinds};
 type ActionList = Arc<Vec<KeybindAction>>;
 
 pub fn register(ctx: &Context) {
-    ctx.on_begin_pass("keybinds", Arc::new(begin_frame))
+    ctx.on_begin_pass("keybinds", Arc::new(|ui| begin_frame(ui.ctx())))
 }
 
 fn begin_frame(ctx: &Context) {
```
knoellle: This function isn't used anywhere else. Why not change its signature instead of wrapping it in a closure?

knoellle (dismissed):
Minor comment, LGTM otherwise

tools/twix/src/configuration/keybind_plugin.rs
oleflb: done

alexschmander (approved):
LGTM

PR #2827, opened 2026-07-24
Remove trailing semicolons in macro in expressions

knoellle:
## Why? What?

The bail macro cannot be used in an expression context in latest rust nightly, causing compilation to fail. This PR changes all our `bail!` invocations to be in a statement context, making everything compile again.

[Rust tracking issue #79813](https://github.com/rust-lang/rust/issues/79813)
Based on #2863

## How to Test

```sh
rustup run nightly ./pepsi clippy
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2461,15 +2461,16 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "cfg_aliases"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
+version = "0.1.2"
+dependencies = [
+ "cfg_aliases 0.2.2",
+]
 
 [[package]]
 name = "cfg_aliases"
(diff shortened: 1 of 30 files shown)


paudar (commented):
I run the pepsi command as stated in the how to test section and got the same warnings as output as on main. Is that expected behavior?
(All of the warnings give a hint to explicitly specify a number as `f32`)

knoellle:
The warnings are separate and unrelated to the hard compilation errors fixed in this PR.
If you want I can fix those too though.

paudar:
My underlying question is: how can I reproduce this build failure? 
I thing a fix of the warnings should not be part of this PR.

knoellle:
Rebased onto #2863 since the ort version we use on main doesn't build with latest nightly anymore.

> how can I reproduce this build failure?

```sh
rustup install nightly
rustup update
rustup run nightly ./pepsi clippy
```

AFAIK you only ran the third command last time but had an older nightly version where the issues I fixed didn't occur yet.

oleflb:
I saw this issue is already fixed upstream in 2a86ae81a42882a18a4e85891e361ab9d827ffa6. Can you please check if a version upgrade also fixes the issue?

knoellle:
Removed `cfg_aliases` shim

PR #2835, opened 2026-08-23
Bump egui + bevy (again)

oleflb:
## Why? What?

Bumps `egui` to `0.36.1`. Since https://github.com/emilk/egui/pull/8321 got merged, we can get rid of our fork. Unfortunately the `wgpu` version no longer aligns with `egui-wgpu` (30) and `bevy` `0.19`, therefore we now use a pinned dev version of `bevy`.
This also requires raising MSRV to at least 1.97. We hadn't yet set the rust-version in our Cargo tomls, so I included that in this PR.

I bumped directly to `1.98.0`, however this also requires a small change to the nix flake, since `1.98` is not yet officially published in the `nixpkgs` `Unstable` channel.

- Fixes #

## ToDo / Known Issues

## How to Test

- Run CI


tools/ci/github-runners/Containerfile
```diff
@@ -25,8 +25,7 @@ RUN apt-get update && apt-get install --no-install-recommends --yes \
 
 RUN useradd -u 1000 hulk && \
     usermod --add-subuids 10000-75535 hulk && \
-    usermod --add-subgids 10000-75535 hulk && \
-    podman system migrate
```
alexschmander, line 29: Why did you remove the `podman system migrate`? Didn't we have this because of some weirdness of the runners? Or was this fixed by using the newer, simpler cross compilation sdk image?

tools/sdk_container/Containerfile
alexschmander, line 1: You also need to bump the sdk version in `team.toml. Currently, the new sdk image is not automatically built, and the old 1.95 one is used, which fails upon the MSRV check.

alexschmander (changes requested):

alexschmander (commented):
Also, I now get this warning during `./pepsi upload --prepare`: 
```bash
warning: the following packages contain code that will be rejected by a future version of Rust: binrw v0.15.1
```

tools/ci/github-runners/Containerfile
oleflb: https://github.com/HULKs/hulk/actions/runs/32641200432/job/97198736779

This runs at container creation, imo it does not really make sense to run any migrations here.

tools/sdk_container/Containerfile
oleflb: done

oleflb:
> Also, I now get this warning during `./pepsi upload --prepare`:
> 
> ```shell
> warning: the following packages contain code that will be rejected by a future version of Rust: binrw v0.15.1
> ```

Bumped `binrw` version, they fixed the problem.

alexschmander (approved):
LGTM

PR #2836, opened 2026-08-25
De-Default

pejotejo:
## Why? What?
I removed a lot of Default Traits, where I don't think the default case is clear and the user should think about what to happen themselves. 

I edited the different Structures and Enums commit by commit, so you can see which changes i did why easier going through the commits.

## ToDo / Known Issues
There are probably some more, I forgot.

## How to Test
Test if the code still compiles and the robot still works (i couldn't yes, because, i'm not in the lab this week)


crates/bevyhavior_simulator/src/behavior_runtime.rs
```diff
@@ -30,17 +30,17 @@ use crate::invariant_checks::BEHAVIOR_TICK_ERROR_CHECK_NAME;
...
 pub struct SimulatorRobotBehavior {
     pub tree: BehaviorNodeTree<BehaviorBlackboard>,
-    pub blackboard: BehaviorBlackboard,
+    pub blackboard: Option<BehaviorBlackboard>,
     pub static_layout: NodeTrace,
 }
 
 impl SimulatorRobotBehavior {
-    pub fn new(parameters: BehaviorParameters) -> Self {
+    pub fn with_standard_tree() -> Self {
```
knoellle, line 38: I know you're trying to remove defaults here, but isn't this a good candidate for using the Default trait?

crates/types/src/world_state.rs
```diff
@@ -34,26 +34,6 @@ pub struct WorldState {
     pub suggested_search_position: Option<Point2<Field>>,
 }
 
-#[allow(clippy::derivable_impls)]
-impl Default for WorldState {
```
knoellle, line 38: I think WorldState should keep a default impl.
The main issue with it is the `PlayerNumber` in `PlayerState`, right? I don't really see why the player number should be part of the player *state*. It's a parameter, not state. If we move it out of Player/World-State, they can retain their default impl.

crates/types/src/world_state.rs
```diff
@@ -93,14 +61,14 @@ impl BallState {
...
 }
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize, Message)]
+#[derive(Clone, Debug, Serialize, Deserialize, Message)]
 pub struct RobotState {
     pub ground_to_field: Option<Isometry2<Ground, Field>>,
-    pub player_number: PlayerNumber,
-    pub primary_state: PrimaryState,
+    pub player_number: Option<PlayerNumber>,
+    pub primary_state: Option<PrimaryState>,
```
knoellle, line 68:
```suggestion
    pub primary_state: PrimaryState,
```

crates/bevyhavior_simulator/src/behavior_runtime.rs
```diff
@@ -30,17 +30,17 @@ use crate::invariant_checks::BEHAVIOR_TICK_ERROR_CHECK_NAME;
 #[derive(Component)]
 pub struct SimulatorRobotBehavior {
     pub tree: BehaviorNodeTree<BehaviorBlackboard>,
-    pub blackboard: BehaviorBlackboard,
+    pub blackboard: Option<BehaviorBlackboard>,
```
knoellle, line 33: This seems like a hack. Is there no way to create this when creating the component?

crates/nodes/behavior_node/src/send_message.rs
```diff
@@ -98,7 +98,7 @@ impl Blackboard {
                     });
 
             let message = HulkMessage::State(StateMessage {
-                player_number: self.world_state.robot.player_number,
+                player_number: self.world_state.robot.player_number?,
```
knoellle, line 101: I really dislike the node running in some semi-broken state when parameters aren't available.

crates/nodes/behavior_node/src/node.rs
```diff
@@ -253,65 +253,21 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
...
+        };
 
-        let player_number = player_number_cache
-            .get_latest()
-            .map(|n| *n)
-            .unwrap_or_default();
-        blackboard.parameters = parameters.snapshot().typed().clone();
+        let Some(primary_state) = primary_state_cache.get_latest() else {
+            continue;
+        };
```
knoellle, line 270: The previous loop had no `continue`s. So previously we began publishing `Dampening` motion command immediately after startup. With this change, that is delayed until parameters and primary state are available. Is that intended/acceptable behavior?

crates/nodes/behavior_node/src/node.rs
```diff
@@ -400,7 +396,9 @@ pub async fn run(ctx: Arc<Context>) -> Result<()> {
             MotionCommand::Prepare => Some(MotionType::Prepare),
         };
 
-        if previous_motion_command != motion_command || motion_type != blackboard.last_motion_type {
+        if previous_motion_command.as_ref() != Some(&motion_command)
+            || motion_type != blackboard.last_motion_type
```
knoellle, line 400: Is this `||` necessary? Is it possible for the motion type to differ from its predecessor but the motion command to stay the same? Also, one is `old != new`, the other `new != old`.

crates/nodes/behavior_node/src/voronoi.rs
```diff
@@ -56,7 +56,9 @@ fn collect_sites(
     blackboard: &Blackboard,
     robot_pose: Pose2<Field>,
 ) -> Vec<(Pose2<Field>, PlayerNumber)> {
-    let robot_player_number = blackboard.world_state.robot.player_number;
+    let Some(robot_player_number) = blackboard.world_state.robot.player_number else {
```
knoellle, line 59: Again, I think its really weird to run behavior in some degraded state if there is no player number.

crates/types/src/filtered_game_state.rs
```diff
@@ -1,16 +1,12 @@
...
 pub enum FilteredGameState {
-    #[default]
     Initial,
     Ready,
     Set,
-    Playing {
-        ball_is_free: bool,
-        kick_off: bool,
-    },
+    Playing { ball_is_free: bool, kick_off: bool },
```
knoellle, line 9: Any idea why this formatting changed?

knoellle (changes requested):
Can't test on a robot but have opinions on some of the code changes.

PR #2851, opened 2026-08-30
Goal Box Obstacle during Opponent Goal Kick

### LINKED ISSUE #2802: players are not allowed in opponent goalboxes during goalkick
During our game against Berlin our striker did not walk out of the goalbox during the goalkick. This might be because we cant walk out of squared obstacles right now, one might check the path planner for that.  It might also be some goal keeper behavior.

soda239:
## Why? What?

A rule obstacle was added to stay outside the opponents goal box area during their goal kick.

- Fixes #2802

## How to Test

With the robot on the field with the GameContoller. Maybe possible in simulator.

### DIFF
--- a/crates/nodes/rule_obstacle_composer/src/lib.rs
+++ b/crates/nodes/rule_obstacle_composer/src/lib.rs
@@ -7,6 +7,8 @@ use hsl_network_messages::{SubState, Team};
 use linear_algebra::{Point, point, vector};
 use ros_z::{prelude::*, qos::QosDurability};
 use serde::{Deserialize, Serialize};
+use types::field_dimensions::Half::Opponent;
+use types::field_dimensions::Side::{Left, Right};
 use types::{
     field_dimensions::FieldDimensions, filtered_game_controller_state::FilteredGameControllerState,
     filtered_game_state::FilteredGameState, rule_obstacles::RuleObstacle, world_state::BallState,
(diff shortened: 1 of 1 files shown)


crates/nodes/rule_obstacle_composer/src/lib.rs
```diff
@@ -103,6 +102,35 @@ fn compose_rule_obstacles(
...
+            },
+            possible_ball,
+        ) => {
+            if let Some(ball) = possible_ball {
+                let goal_kick_obstacle = RuleObstacle::Circle(Circle::new(
+                    ball.ball_in_field,
+                    parameters.free_kick_obstacle_radius,
+                ));
+                rule_obstacles.push(goal_kick_obstacle);
+            }
```
oleflb, line 122: Is this required? I think we already respect not touching the ball whenever ball is not free.

paudar: Currently, a block position is being determined in our behaviour that lies outside the rule obstacle. The rule obstacle is relevant for the path planner to ensure that, upon reaching the block position, it does not pass too close to the ball.

crates/nodes/rule_obstacle_composer/src/lib.rs
```diff
@@ -103,6 +102,35 @@ fn compose_rule_obstacles(
...
+
+            let goal_box_obstacle = RuleObstacle::Rectangle(Rectangle {
+                min: point!(
+                    field_dimensions.length / 2.0 - field_dimensions.goal_box_area_length,
+                    -field_dimensions.goal_box_area_width / 2.0
+                ),
+                max: point!(
+                    field_dimensions.length / 2.0,
+                    field_dimensions.goal_box_area_width / 2.0
+                ),
```
paudar: Take a look at the functions `goal_box_corner` and `goal_box_goal_line_intersection` from `field_dimensions`. There the same calculation is done :)

paudar (commented):
One small comment:

crates/nodes/rule_obstacle_composer/src/lib.rs
soda239: Done

paudar (approved):
LGTM and worked on the robot

PR #2807, opened 2026-07-04
port whistle detection 

vivi-enne:
## Why? What?

Ports whistle detection and tunes whistle band pass filter.

- Fixes #2283 

## ToDo / Known Issues

*If this is a WIP describe which problems are to be fixed.*

## Ideas for Next Iterations (Not This PR)

*If there are some improvements that could be done in a next iteration, describe them here.*

## How to Test

Borrow whistle and put robot on field and blow whistle.

PR #2855, opened 2026-09-05
add audio spectrum to twix again

vivi-enne:
## Why? What?

The audio spectrum panel is added to twix again. Is nice to have for tuning the whistle. Only usable after merging #2807 otherwise the corresponding topics are not provided. 

- Fixes #

## ToDo / Known Issues

## How to Test
Connect to twix and open Audio panel. See spectrum and waterfall for all three channels.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,9 +197,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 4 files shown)


tools/twix/src/panels/audio/mod.rs
```diff
@@ -0,0 +1,660 @@
...
+                            0.5
+                        } else if ratio > 2.0 {
+                            0.8
+                        } else {
+                            0.95
+                        };
+                        self.current_max_magnitude = (current * decay_factor).max(target_max);
+                    }
+
+                    self.waterfall_history.push_front(magnitudes);
```
alexschmander, line 203: This adds the latest sample to the history on every repaint, even if no new audio data arrived. Moving the mouse or receiving updates in another panel therefore advances the waterfall and evicts actual samples. Could you track the last processed sample and only append when it changes?

Cargo.lock
alexschmander, line 1: Why are all these dependencies updated? There are no `Cargo.toml` changes, and this also updates workspace dependencies like Tokio and Zenoh. Please revert the unrelated lockfile changes.

alexschmander (changes requested):
A couple of comments :)

PR #2857, opened 2026-09-07
port behavior panel

pejotejo:
## Why? What?

what the title says

## How to Test

`./twix --router...` and look at the behavior tree of a connected robot


JackyBloxx:
I tested i but the Behavior tree never appeard

PR #2858, opened 2026-09-07
Aligne the Voronoigrid to the map center

JackyBloxx:
## Why? What?

Centers the Voronoi grid to the map center

## How to Test

Check in twix if the Voronoi grid is aligned to the center of the map.
To do this with one Robot on the field move in the behavior tree calculate_voronoi_grid to the first place in the playing tree


BenSampaolo (approved):
LGTM! Next time better PR message from the start please :)

Issue #2853, opened 2026-09-04
Controller in playing


BenSampaolo:
We are currently able to use the controller while in play which could be a rule violation. @pejotejo claims that we probably want to have our own controller runtime to have better/easier control when this is active and more features (Matching the boosterized state?)

PR #2859, opened 2026-09-07
direct remote control on robot

pejotejo:
## Why? What?
Adds a controller handler, that takes the direct controller input, also updates gammaray for that. Also disables the booster controller service with boosterize and reenables it with boosterize. Then it rebinds the behavior remote control tree to the right controls. The remote controll is always active if a controller is connected, so all controllers need to be turned off during the game, else it will not do anything.

loca for the kick for real torwandschießen without the field

## How to Test
Gammaray and upload the robot connect a controller and look in the docs for controls, afterwards boosterize the robot and look if it can wave and kickbox again.

### DIFF
--- a/crates/types/src/controller_input.rs
+++ b/crates/types/src/controller_input.rs
@@ -16,6 +16,13 @@ impl ControllerInput {
             .map_or(0.0, |axis| axis.value)
     }
 
+    pub fn button_value(&self, name: &str) -> f32 {
+        self.buttons
+            .iter()
+            .find(|button| button.name == name)
+            .map_or(0.0, |button| button.value)
+    }
+
     pub fn is_pressed(&self, name: &str) -> bool {
         self.buttons
             .iter()
--- a/crates/types/src/controller_input.rs
+++ b/crates/types/src/controller_input.rs
@@ -1,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize, ros_z::Message)]
+pub struct ControllerInput {
+    pub connected: bool,
+    pub device_name: String,
+    pub axes: Vec<ControllerAxis>,
+    pub buttons: Vec<ControllerButton>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ros_z::Message)]
+pub struct ControllerAxis {
+    pub name: String,
+    pub value: f32,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ros_z::Message)]
+pub struct ControllerButton {
+    pub name: String,
+    pub pressed: bool,
+    pub value: f32,
+}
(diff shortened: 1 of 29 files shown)


alexschmander (changes requested):
Needs to be rebased...

tools/sdk_container/Containerfile
alexschmander, line 1: After changing the sdk Containerfile, please bump the sdk version in the team.toml.

docs/tooling/remote_control.md
alexschmander, line 1: You could add here as a safety warning, that any connected controller overrides the robots behavior started motion, potentially leading to problematic situations when a controller is left connected by accident.

alexschmander (changes requested):

docs/tooling/remote_control.md
knoellle: I disagree. Overriding normal behavior with remote control should require pressing a button(-combo) on the controller first. Same thing for disabling it again.
The robot must not suddenly switch back to normal game behavior just because the controller connection drops or it turns off due to low battery/idle time!

My suggestion: ignore connection state, have a toggle button for enabling remote control. Enabled state shouldn't be affected by controller (dis-)connection either.

tools/sdk_container/Containerfile
pejotejo: done

docs/tooling/remote_control.md
pejotejo: i changed it that 
- the start button also needs to be toggled in,
- if the controller disconnects without toggeling, the robot does stand
- if the primary state changes, the remote controll toggles off

alexschmander (changes requested):
Needs to be rebased again...

docs/tooling/remote_control.md
```diff
@@ -0,0 +1,25 @@
+# Remote Control
+
+`./pepsi gammaray <robot>` disables the manufacturer controller services for HULK remote control.
+`./pepsi boosterize <robot>` restores the `RemoteController` section in `/opt/booster/Daemon/bin/child.ini` and enables and starts `joystick_ros2`.
+
```
alexschmander, line 5: This doesn't mention, that the robot needs to be in Initial for the remote control to work.

crates/types/src/controller_input.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn is_pressed(&self, name: &str) -> bool {
+        self.buttons
+            .iter()
+            .any(|button| button.name == name && button.pressed)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ros_z::Message)]
+pub struct ControllerAxis {
+    pub name: String,
```
alexschmander, line 35: Both here and in `ControllerButton`, individual items are identified via string matching... Please make these enums, so that this is safer.

alexschmander (changes requested):

alexschmander:
Does this close #2853 ?

alexschmander (approved):

PR #2860, opened 2026-09-08
Fix single Robot Search/Kick Behavior

JackyBloxx:
## Why? What?

In the bahavior tree if the robot was alone on the field it got stuck in a loop from:  Im alone go to stricker behavior without searching for the ball first, therefore we got stuck in a position.

Also in kick.rs we checkt for the false ball when we had not a ball percept the kick action failed beside having a ball from the ball filter.

## How to Test

Check if only one robot is on the field if it searches for the ball(action: leuchturm) if you take the ball away.


pejotejo (approved):
lgtm

PR #2861, opened 2026-09-08
Fix K1 leg Forward kinematic and asset dimensions

BenSampaolo:
## Why? What?

Outdated assets had wrong sizes and two of the joints were interpreted wrong for the forwards kinematic

## How to Test

Look at the official Booster documentation. Relatively straightforward


pejotejo (approved):
lgtm

PR #2863, opened 2026-09-09
Bump ort

alexschmander:
## Why? What?

Upgrade `ort` to rc.13 while keeping ONNX Runtime 1.22 and CUDA 12.8. Migrate inference APIs and recreate the runtime container during `gammaray` to activate the required preload.

## ToDo / Known Issues

- Verify inference and clean shutdown on the robot.

- Upgrade ort to rc14 and with it ONNX Runtime, once the Jetson OS supports CUDA 13 (Jetpack 7 needed)

## How to Test

Run `./pepsi gammaray 42`, then `./pepsi upload 42`. Check HULK startup, GPU inference, and shutdown for errors.


crates/nodes/stereo_visual_odometry/src/feature_extractor.rs
```diff
@@ -68,17 +67,20 @@ pub struct Matches<'a, From, To> {
...
             .with_engine_cache_path(parent.display())
             .build();
-        let cuda = CUDAExecutionProvider::default().build();
+        let cuda = CUDA::default().build();
         let session = Session::builder()?
-            .with_execution_providers([tensorrt, cuda])?
-            .with_optimization_level(GraphOptimizationLevel::Level3)?
-            .with_intra_threads(2)?
+            .with_execution_providers([tensorrt, cuda])
+            .map_err(ort::Error::<()>::from)?
```
oleflb, line 79: why is this required? from docs it looks like Error is implemented without this conversion as well

tools/k1-setup/inference-runtime/README.md
```diff
@@ -0,0 +1,29 @@
+# Inference runtime compatibility
```
oleflb: Is this really needed? I think the relevant info should be at dependency selection or in the docs

tools/k1-setup/hulk-runtime.container
```diff
@@ -15,6 +15,7 @@ WorkingDir=/home/booster/hulk
 
 Environment=LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu/tegra
 Environment=ORT_DYLIB_PATH=/usr/local/lib/libonnxruntime.so
+Environment=LD_PRELOAD=/usr/local/lib/libonnxruntime.so
```
oleflb, line 18: why preload?

tools/pepsi/src/gammaray.rs
```diff
@@ -322,8 +316,20 @@ async fn gammaray_robot(
 
     robot
         .ssh_to_robot()?
-        .arg("sudo systemctl enable hulk && sudo systemctl restart hulk")
-        .ssh_with_log("enabling and restarting hulk", &progress_bar)
+        .arg("sudo systemctl stop hulk")
```
oleflb: shouldn't `hulk` automatically be stopped when `hulk-runtime` stops? If not, we should fix this at `systemd` level, not here

oleflb (changes requested):

tools/k1-setup/hulk-runtime.container
alexschmander: https://github.com/pykeio/ort/issues/614

There seems to be an issue during teardown of the onnxruntime session. This preloading fixed that issue.
The above mentioned issues was already fixed upstream, but the next release of ort is a bump of onnxruntime, requiring a CUDA bump form 12 to 13 and thereby JetPack 6 to 7 bump...

crates/nodes/stereo_visual_odometry/src/feature_extractor.rs
alexschmander: `eyre` requires `Send` and `Sync`, which this new `SessionBuilder` error does not implement...

tools/pepsi/src/gammaray.rs
alexschmander: Fair point, `hulk.service` already `Requires=hulk-runtime.service`... Removed

tools/k1-setup/inference-runtime/README.md
alexschmander: Moved to `docs/setup/development_environment.md`. This is still not the best place for it, but a better place would probably require a complete docs overhaul for the deployment setup.

oleflb (approved):

PR #2864, opened 2026-09-10
Fix/cargo container cache invalidations

alexschmander:
## Why? What?

We currently experience cargo cache invalidations, when building for native target, e.g. ./pepsi, ./twix, and in the cross-compilation container, e.g. ./pepsi upload --prepare, in an alternating fashion. This is because of cache paths inside and outside of the cross-compilation container not being identical and thus being invalidated. 
This PR introduces separate target paths for native and cross-compilation targets in pepsi.   

## ToDo / Known Issues

- I am not sure if this also fixes the cache invalidation we saw sometimes caused by rust analyzer touching files. 

## How to Test

- cargo clean
- warm-up: `./pepsi` 
- warm-up: `./pepsi upload --prepare`
- Should now remain hot: `./pepsi` 
- Should now remain hot: `./pepsi upload --prepare`


crates/repository/src/cargo.rs
```diff
@@ -189,12 +202,14 @@ fn build_command_string(
     root: String,
     tagged_image_name: String,
     pwd: String,
+    target_directory: &Path,
```
knoellle, line 205: Type inconsistent with the other arguments. The other paths are converted to string at callsite. Then you can also remove the additional argument to `format!`.

crates/repository/src/cargo.rs
```diff
@@ -23,6 +23,16 @@ pub enum Environment {
     Docker { sdk_image: SDKImage },
 }
 
+impl Environment {
+    /// Default artifact directory, relative to the repository root.
+    pub fn default_target_directory(&self) -> &Path {
```
knoellle, line 28: "Default" suggests it can be overridden somehow. I don't think we need an override here though.

Furthermore, the native branch is never actually used, so I'd like to remove it. I guess you made this function to deduplicate code between the docker and podman branches, but all the other path handling there is duplicated anyway. I'm fine with either but please make it consistent.

tools/pepsi/src/tensorrt_compile.rs
```diff
@@ -68,14 +66,14 @@ pub async fn tensorrt_compile(arguments: Arguments, repository: &Repository) ->
     let progress_compile = multiprogres.task("Compile", false);
     let progress_download = multiprogres.task("Download", false);
 
-    let binary_path = get_binary(arguments.build.profile(), BINARY_NAME);
+    let cargo_arguments = cargo::Arguments {
+        manifest: Some(manifest(repository)),
+        environment: arguments.environment,
+        cargo: arguments.build,
+    };
+    let binary_path = cargo_arguments.binary_path(repository, BINARY_NAME).await?;
```
knoellle, line 74: This changes the path but `tools/tensorrt-compile/README.md` still has some references to the old path. Fls pix

tools/pepsi/src/cargo.rs
```diff
@@ -66,6 +66,73 @@ pub trait CargoCommand {
     fn profile(&self) -> &str;
 }
 
+impl<CargoArguments: Args> Arguments<CargoArguments> {
+    async fn resolve(
```
knoellle, line 70: Seems a bit weird to factor out a function because you want to reuse the code in a second location but the new function returns two values, one of which is discarded in the new use.

knoellle (changes requested):
Nice! Was able to reproduce improved hot-rebuild times, though I have some comments on the implementation.

crates/repository/src/cargo.rs
alexschmander: Done

crates/repository/src/cargo.rs
alexschmander: Done

tools/pepsi/src/tensorrt_compile.rs
alexschmander: Done

tools/pepsi/src/cargo.rs
alexschmander: Done

knoellle (approved):
Code LGTM but can't test deployment. Reenable merge when ready when tested (already)

PR #2865, opened 2026-09-14
Fix ground provider displacement

BenSampaolo:
## Why? What?

Fixes a frame mismatch in the ground transform. Rotates the displacement between soles from Robot into the horizontal frame before projecting onto XY. Minimal difference (Sub cm scale)

## How to Test

Look at the code


alexschmander (changes requested):
CI is failing. Otherwise looks good

alexschmander (approved):
LGTM

PR #2867, opened 2026-09-14
fix: use game controller default messaging and behavior parameter for network prameters in Behavior tree simulator

JackyBloxx:
## Why? What?

Before this fix the default message count from the gamecontroller would be overwriten only for the hulks team to u16::MAX. Also the message interval for the opponent team was defaultet to 0 therefore using all its messages in the first seconds. For that i changed the default to the behavior parameter from the json5

## How to Test

compare to the old behavior tree simulator how the message from the opponent team goes down in comparision to this fix and also compare the message limit from the old to this change.

both teams should start with 12000 Messages like the game_controller.rs says and also  the messages should go down for both teams in an interval like in the parameters


pejotejo (approved):
lgtm

PR #2868, opened 2026-09-15
twix: add save/restore functionality for subtrees

oleflb:
## Why? What?

This PR allows saving single panels/layouts as presets, which can easily be recreated.
A layout is just a normal layout tree and can therefore be created at any level in the tree and freely moved around.

- Fixes #

## ToDo / Known Issues

## How to Test

- Create layouts, save them, move them, delete them


alexschmander (approved):
LGTM

